PR #87, opened 2021-12-02
[actors consensus fix] gating behind version checks to really fix F24

### LINKED ISSUE #49: F24: Forest does not reject BLS identity address during account actor creation

laudiacay:
Closes #49


raulk:
FVM only supports nv14 and above, so this patch is not really needed.

PR #163, opened 2021-12-18
Added Arch checking in makefile (specifically to catch M1 Macs)

aronchick:


raulk:
Go code has moved out of this repo into the https://github.com/filecoin-project/filecoin-ffi/ repo which implements the Go<>Rust glue layer. So this PR no longer applies.

PR #272, opened 2022-01-17
simplify gas charging logic

Stebalien:
(there's no reason to sum it twice)


marcellina-from-alexandria (approved):
lgtm. forest is doing just a subtract and a compare at this point, although i think this is clearer what's going on... if you're trying to optimize, you might want to switch to match them.

laudiacay (approved):
no idea who that other person is but i concur

PR #273, opened 2022-01-18
Harden actor deletion (and some other errors)

### LINKED ISSUE #185: Nothing should be allowed after actor deletion
Currently:

1. The actor self delete method will return to the actor. At this point, the actor can't safely do much of anything.
2. An actor can be deleted by a re-entrant call. That means _any_ send could end up deleting the actor.

We have three options:

1. Always return to the actor after deletion, re-entrant or otherwise. In this case, we'd need to forbid any state-related syscalls (and send?) after deletion but the caller would still be able to abort, return some value, etc. They just wouldn't be able to work with state and/or send messages.
2. Return to the kernel after self deletion but revert after re-entrant deletion. This is slightly safer (reduces spooky action at a distance) but less flexible. I say "slightly safer" because actors need to handle re-entrency safely anyways, and this is _definitely_ less flexible.
3. Trap inside the actor self delete method, returning directly to the caller. This is the safest way forward, but makes it more difficult to return a result (unless `delete_self` takes a return value).

I started out with 3, but I'm leaning towards 1 now. This is _just_ another form of re-entrancy and the will need to guard against that anyways.

Stebalien:
These cases should be unreachable in our current actors, but would not have been unreachable in user programmable actors.

Fixes #185 by _not_ returning fatal errors if the current actor doesn't exist in state.

After deletion:

1. Actors have "zero" balance.
2. Getting & setting the root fails with "illegal actor".
3. Sending still works, but sending with a value will fail because the balance is zero.
4. Sending _to_ this actor will fail.

Effectively, I'm treating "deletion" as "unlinking" the actor in the state-tree. It still _exists_ until it returns, it just can't be looked-up or linked back into the state.

NOTE: the exact error codes will change with https://github.com/filecoin-project/fvm-specs/pull/53.

### DIFF
--- a/fvm/src/executor/default.rs
+++ b/fvm/src/executor/default.rs
@@ -249,12 +249,11 @@ where
         }
 
         // Deduct message inclusion gas cost and increment sequence.
-        self.state_tree_mut()
-            .mutate_actor(&Address::new_id(sender_id), |act| {
-                act.deduct_funds(&gas_cost)?;
-                act.sequence += 1;
-                Ok(())
-            })?;
+        self.state_tree_mut().mutate_actor_id(sender_id, |act| {
+            act.deduct_funds(&gas_cost)?;
+            act.sequence += 1;
+            Ok(())
+        })?;
 
         Ok(Ok((sender_id, gas_cost, inclusion_cost)))
     }
--- a/fvm/src/kernel/default.rs
+++ b/fvm/src/kernel/default.rs
@@ -1,7 +1,7 @@
 use std::collections::BTreeMap;
 use std::convert::{TryFrom, TryInto};
 
-use anyhow::anyhow;
+use anyhow::{anyhow, Context as _};
 use byteorder::{BigEndian, WriteBytesExt};
 use cid::Cid;
 use filecoin_proofs_api::seal::{
@@ -26,6 +26,7 @@ use rayon::iter::{IndexedParallelIterator, IntoParallelRefIterator, ParallelIter
 use super::blocks::{Block, BlockRegistry};
 use super::error::Result;
 use super::*;
+use crate::account_actor::is_account_actor;
 use crate::builtin::{is_builtin_actor, is_singleton_actor, EMPTY_ARR_CID};
 use crate::call_manager::{CallManager, InvocationResult};
 use crate::externs::{Consensus, Rand};
@@ -110,14 +111,21 @@ where
         let act = state_tree
             .get_actor(addr)?
             .ok_or(anyhow!("state tree doesn't contain actor"))
-            .or_error(ExitCode::ErrIllegalArgument)?;
+            .or_illegal_argument()?;
+
+        if !is_account_actor(&act.code) {
+            return Err(
+                syscall_error!(SysErrIllegalArgument; "target actor is not an account").into(),
+            );
+        }
 
         let state: crate::account_actor::State = state_tree
             .store()
             .get_cbor(&act.state)
-            .or_fatal()?
+            .context("failed to decode actor state as an account")
+            .or_fatal()? // because we've checked and this should be an account.
             .ok_or(anyhow!("account actor state not found"))
-            .or_fatal()?;
+            .or_fatal()?; // because the state should exist.
 
         Ok(state.address)
     }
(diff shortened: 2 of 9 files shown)


fvm/src/executor/default.rs
```diff
@@ -249,12 +249,11 @@ where
         }
 
         // Deduct message inclusion gas cost and increment sequence.
-        self.state_tree_mut()
-            .mutate_actor(&Address::new_id(sender_id), |act| {
-                act.deduct_funds(&gas_cost)?;
-                act.sequence += 1;
-                Ok(())
-            })?;
+        self.state_tree_mut().mutate_actor_id(sender_id, |act| {
```
Stebalien, line 252: Small optimization.

fvm/src/kernel/default.rs
```diff
@@ -110,14 +111,21 @@ where
         let act = state_tree
             .get_actor(addr)?
             .ok_or(anyhow!("state tree doesn't contain actor"))
-            .or_error(ExitCode::ErrIllegalArgument)?;
+            .or_illegal_argument()?;
+
+        if !is_account_actor(&act.code) {
```
Stebalien, line 116: Not directly related to deletion bug. But we would have previously returned a fatal error here if we attempted to "resolve" a non-account ID address.

fvm/src/kernel/default.rs
```diff
@@ -876,6 +900,9 @@ fn verify_seal(vi: &SealVerifyInfo) -> Result<bool> {
         bytes_32(&vi.interactive_randomness.0),
         &vi.proof,
     )
-    .or_fatal()
-    .context("failed to verify seal proof") // TODO: Verify that this is actually a fatal error.
+    .or_illegal_argument()
```
Stebalien, line 903: Another drive-by fix.

fvm/src/machine/default.rs
```diff
@@ -244,18 +245,17 @@ where
             .into());
         }
 
-        // TODO: make sure these are actually fatal.
         let mut from_actor = self
```
Stebalien, line 248: We can hit this if the actor has been deleted.

fvm/src/machine/default.rs
```diff
@@ -244,18 +245,17 @@ where
...
+            .context("cannot transfer from non-existent sender")
+            .ok_error(ExitCode::SysErrSenderInvalid)?;
 
         let mut to_actor = self
             .state_tree
             .get_actor_id(to)?
-            .ok_or_else(|| anyhow!("receiver actor does not exist in state during transfer"))
-            .or_fatal()?;
+            .context("cannot transfer to non-existent receiver")
+            .ok_error(ExitCode::SysErrInvalidReceiver)?;
```
Stebalien, line 258: We probably shouldn't hit this, but it can't hurt.

fvm/src/kernel/mod.rs
```diff
@@ -121,7 +121,7 @@ pub trait BlockOps {
 /// Depends on BlockOps to read and write blocks in the state tree.
 pub trait SelfOps: BlockOps {
     /// Get the state root.
-    fn root(&self) -> Cid;
+    fn root(&self) -> Result<Cid>;
```
Stebalien, line 124: This needs to return an error if the actor no longer exists in the state-tree.

fvm/src/kernel/default.rs
```diff
@@ -110,14 +111,21 @@ where
         let act = state_tree
             .get_actor(addr)?
             .ok_or(anyhow!("state tree doesn't contain actor"))
-            .or_error(ExitCode::ErrIllegalArgument)?;
+            .or_illegal_argument()?;
+
+        if !is_account_actor(&act.code) {
+            return Err(
+                syscall_error!(SysErrIllegalArgument; "target actor is not an account").into(),
```
arajasek, line 118: still having a hard time wrapping my head around fvm error handling, but what happens to the error code here? does the calling actor get to decide what to do with the error?

there may be a risk of breaking consensus if that's not the case.

shared/src/error/mod.rs
```diff
@@ -19,13 +19,18 @@ pub enum ExitCode {
     /// Indicates failure to find an actor in the state tree.
     SysErrSenderInvalid = 1,
 
-    /// Indicates failure to find the code for an actor.
+    /// Indicates that the message sender was not in a valid state to send this message. This means
```
arajasek, line 22: Nit that "the message shouldn't have been included on-chain" isn't strictly true -- a sender's balance can change within tipset execution causing not enough funds to cover message gas.

Also note that the nonce case never actually happens, since blocks containing such messages are invalid (and so not on chain).

Stebalien: Fixed.

PR #274, opened 2022-01-18
feat: remove errors from infallible sdk functions

### LINKED ISSUE #190: Remove results from SDK syscalls that shouldn't fail
Basically, giving the errors to the actor gives the actor a chance to react. But the _default_ behavior should be to just panic (abort) because all these errors mean that there is a bug in the actor.

- `actor::resolve_address` should only fail if the address is malformed, which is statically impossible in rust.
- `actor::get_actor_code_cid` ditto.
- `actor::new_actor_address` can only fail if the input buffer is to small, which should be statically impossible.
- `crypto::hash_blake2b` shouldn't fail unless we pass in a bad buffer.
- `gas::charge` only fails if the name parameter is out of range (a bug).
- Almost everything in `message`. We return errors for consistency (and "just in case we need it later"), but these methods simply can't fail and there's nothing a reasonable actor could do about it if they did (we don't even know why they'd fail right now).
- Everything in `network`.
- `sself::get_root`, `sself::current_balance`

On the maybe side: `ipld::put`, `ipld::get`, `sself::set_root`. If something goes wrong here, the actor _may_ want to recover, but that's not going to be the common case.

Stebalien:
Either:

1. These functions statically cannot fail (https://github.com/filecoin-project/fvm-specs/issues/580 to follow up on this).
2. These functions can only fail when passed invalid data that we've ensured is statically impossible with the type system.

Changes:

- FVM
    - `curr_epoch` syscall returns an i64 to be consistent everywhere.
- SDK
    - Newly infallible: `actor::resolve_address`, `actor::get_actor_code_cid`, `crypto::hash_blake2b`, `gas::charge`, `network::*`, `sself::current_balance`.
    - Removed `ipld::transaction` (dead code).
    - Changed `Address` and `Cid` to by-reference in some cases for better consistency.
- Actors
    - Removed "fatal" errors.
    - Adjusted to new infallible syscalls.

fixes #190

### DIFF
--- a/actors/runtime/src/test_utils.rs
+++ b/actors/runtime/src/test_utils.rs
@@ -502,24 +502,24 @@ impl Runtime<MemoryBlockstore> for MockRuntime {
         )
     }
 
-    fn current_balance(&self) -> Result<TokenAmount, ActorError> {
+    fn current_balance(&self) -> TokenAmount {
         self.require_in_call();
-        Ok(self.balance.borrow().clone())
+        self.balance.borrow().clone()
     }
 
-    fn resolve_address(&self, address: &Address) -> Result<Option<Address>, ActorError> {
+    fn resolve_address(&self, address: &Address) -> Option<Address> {
         self.require_in_call();
         if address.protocol() == Protocol::ID {
-            return Ok(Some(*address));
+            return Some(*address);
         }
 
-        Ok(self.id_addresses.get(address).cloned())
+        self.id_addresses.get(address).cloned()
     }
 
-    fn get_actor_code_cid(&self, addr: &Address) -> Result<Option<Cid>, ActorError> {
+    fn get_actor_code_cid(&self, addr: &Address) -> Option<Cid> {
         self.require_in_call();
 
-        Ok(self.actor_code_cids.get(addr).cloned())
+        self.actor_code_cids.get(addr).cloned()
     }
 
     fn get_randomness_from_tickets(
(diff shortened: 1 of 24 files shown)


Stebalien:
I'm fixing this _now_ as part of my effort to remove fatal errors from the actors.

Stebalien:
> I'm fixing this now as part of my effort to remove fatal errors from the actors.

Eh, that commit is small enough that I've just tacked it onto this PR.

actors/runtime/src/test_utils.rs
```diff
@@ -670,13 +670,13 @@ impl Runtime<MemoryBlockstore> for MockRuntime {
...
-    fn total_fil_circ_supply(&self) -> Result<TokenAmount, ActorError> {
+    fn total_fil_circ_supply(&self) -> TokenAmount {
         unimplemented!();
     }
 
-    fn charge_gas(&mut self, _: &'static str, _: i64) -> Result<(), ActorError> {
+    fn charge_gas(&mut self, _: &'static str, _: i64) {
         // TODO implement functionality if needed for testing
-        Ok(())
+        ()
```
laudiacay, line 679: clippy is complaining about this

laudiacay (changes requested):
nice, cleans things up a lot :D have a few nits and they're all clippy warnings but i think this is ready to merge otherwise

laudiacay (approved):

PR #276, opened 2022-01-19
Return TokenAmount instead of BigInt

arajasek:
Not necessary, but allows us to use the splitting into int64 utils in the FFI directly.

Sidebar: having the 2 types with the same name is annoying, no?


Stebalien:
> having the 2 types with the same name is annoying, no?

Yeah, I'd _like_ to use the tuple in most places, honestly. But that'll take some work.

arajasek:
@Stebalien suggests using the u64 based one everywhere except in actors itself, which works for me...will take a stab at the refactor tomorrow

Stebalien:
We'd likely need to implement a `checked_add` and `checked_sub` function as well.

arajasek:
Closing in favour of #281

PR #281, opened 2022-01-20
Make the FVM use tuple-based TokenAmount only

arajasek:
This needs some cleanup, but ready for a review. Some notes:

- I moved to always using values since it's just 2 u64s now, but I'm not sure whether that was the right thing to do
- The new type's methods need to be scrutinized for correctness. As far as possible I converted to BigInt before performing any operations.

### DIFF
--- a/shared/src/sys/mod.rs
+++ b/shared/src/sys/mod.rs
@@ -1,7 +1,15 @@
 //! This module contains types exchanged at the syscall layer between actors
 //! (usually through the SDK) and the FVM.
 
+use std::cmp::Ordering;
+use std::fmt;
+use std::hash::{Hash, Hasher};
+use std::ops::{Add, Mul, Sub};
+
+use num_traits::Zero;
+
 pub mod out;
+pub mod tokenamount_ser;
 
 pub type BlockId = u32;
 pub type Codec = u64;
@@ -13,12 +21,65 @@ pub struct TokenAmount {
     pub hi: u64,
 }
 
+impl TokenAmount {
+    pub fn checked_sub(&self, v: &TokenAmount) -> Option<TokenAmount> {
+        if self < v {
+            return None;
+        }
+
+        crate::econ::TokenAmount::from(self)
+            .checked_sub(&crate::econ::TokenAmount::from(v))
+            .unwrap()
+            .try_into()
+            .ok()
+    }
+
+    pub fn checked_add(&self, v: &TokenAmount) -> Option<TokenAmount> {
+        crate::econ::TokenAmount::from(self)
+            .checked_add(&crate::econ::TokenAmount::from(v))
+            .unwrap()
+            .try_into()
+            .ok()
+    }
+}
+
 impl From<TokenAmount> for crate::econ::TokenAmount {
     fn from(v: TokenAmount) -> Self {
         crate::econ::TokenAmount::from(v.hi) << 64 | crate::econ::TokenAmount::from(v.lo)
     }
 }
 
+impl From<&TokenAmount> for crate::econ::TokenAmount {
+    fn from(v: &TokenAmount) -> Self {
+        crate::econ::TokenAmount::from(v.hi) << 64 | crate::econ::TokenAmount::from(v.lo)
+    }
+}
+
+impl From<u32> for TokenAmount {
+    fn from(v: u32) -> Self {
+        TokenAmount {
+            hi: 0,
+            lo: v as u64,
+        }
+    }
+}
+
+impl From<u64> for TokenAmount {
+    fn from(v: u64) -> Self {
+        TokenAmount { hi: 0, lo: v }
+    }
+}
+
+impl From<u128> for TokenAmount {
+    fn from(v: u128) -> Self {
+        TokenAmount {
+            hi: (v >> u64::BITS) as u64,
+            lo: v as u64,
+        }
+    }
+}
+
+// review: these next 2 methods should fail on negative input, unless the map to u128 does that for you
 impl TryFrom<crate::econ::TokenAmount> for TokenAmount {
     type Error = <crate::econ::TokenAmount as TryInto<u128>>::Error;
     fn try_from(v: crate::econ::TokenAmount) -> Result<Self, Self::Error> {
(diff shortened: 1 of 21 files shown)


arajasek (commented):
Some notes:

- Not sure exactly what to do when we fail to convert bigints into tuples. Should we just have a `From` for that that panics?
- Relatedly, how can we super duper ensure that we never have negative numbers masquerading as very large positive numbers?
- looks like rustfmt has overdone its formatting in places, why this happen

Stebalien:
> Not sure exactly what to do when we fail to convert bigints into tuples. Should we just have a From for that that panics?

One usually uses `TryFrom`.

> Relatedly, how can we super duper ensure that we never have negative numbers masquerading as very large positive numbers?

Check all operations to ensure we don't wrap, check on conversion.

> looks like rustfmt has overdone its formatting in places, why this happen

Just... let it do its thing.

shared/src/sys/mod.rs
```diff
@@ -38,3 +99,129 @@ impl<'a> TryFrom<&'a crate::econ::TokenAmount> for TokenAmount {
...
+
+impl Mul<u64> for &TokenAmount {
+    type Output = TokenAmount;
+
+    fn mul(self, rhs: u64) -> Self::Output {
+        *self * rhs
+    }
+}
+
+impl Mul<i64> for &TokenAmount {
```
Stebalien, line 205: So, two options:

1. Don't implement mul, add, etc. and expect the user to call `checked_*` everywhere.
2. Implement them, but make the output something like `Result<TokenAmount, OverflowError>`. Users will need to write things like `(a * b)?`.
3. Implement them, but make the output some special `MaybeTokenAmount` with a `check(self) -> Result<TokenAmount, OverflowError>` function. Users would write, e.g., `(a * b + c).check()?`.

See https://crates.io/crates/checked for some inspiration.

shared/src/sys/mod.rs
```diff
@@ -38,3 +99,129 @@ impl<'a> TryFrom<&'a crate::econ::TokenAmount> for TokenAmount {
...
+}
+
+impl Mul<u64> for TokenAmount {
+    type Output = TokenAmount;
+
+    fn mul(self, rhs: u64) -> Self::Output {
+        let eta: crate::econ::TokenAmount = self.into();
+        crate::econ::TokenAmount::from(eta * rhs)
+            .try_into()
+            .unwrap()
```
Stebalien, line 181: We don't actually _need_ BigInt here. We can:

1. Convert to a u128 (this should compile to a cast if we do it right).
2. Perform a checked operation with u128.
3. Convert back.

arajasek: Yessir!

shared/src/sys/mod.rs
arajasek: Okay, I went with 2 here. I think it works fine, except that I'm very liberally `unwrap`ing in places -- I'm not sure what the impact of that is for exit codes (though these are mostly fatal cases anyway).

shared/src/sys/mod.rs
```diff
@@ -1,40 +1,243 @@
...
+    type Error = TokenAmountError;
+    fn try_from(v: crate::econ::TokenAmount) -> Result<Self, Self::Error> {
+        match v.sign() {
+            Sign::Minus => Err(Underflow),
+            Sign::NoSign => Ok(TokenAmount::zero()),
+            Sign::Plus => match u128::try_from(v) {
+                Ok(v128) => Ok(v128.into()),
+                Err(_) => Err(Overflow),
+            },
+        }
```
Stebalien, line 155: `TryFrom::try_from(&v)`

shared/src/sys/mod.rs
```diff
@@ -1,40 +1,243 @@
...
+    fn from(v: u128) -> Self {
+        TokenAmount {
             hi: (v >> u64::BITS) as u64,
             lo: v as u64,
-        })
+        }
+    }
+}
+
+// review: these next 2 methods should fail on negative input, unless the map to u128 does that for you
```
Stebalien, line 144: https://docs.rs/num-bigint/0.4.3/src/num_bigint/bigint/convert.rs.html#89-95

shared/src/sys/mod.rs
```diff
@@ -1,40 +1,243 @@
...
+    }
+
+    pub fn checked_add(&self, rhs: TokenAmount) -> Option<TokenAmount> {
+        match u128::from(self).checked_add(u128::from(rhs)) {
+            None => None,
+            Some(v) => Some(v.into()),
+        }
+    }
+
+    // should they have inline directives?
```
Stebalien, line 227: In this case, we might as well.

shared/src/sys/tokenamount_ser.rs
```diff
@@ -0,0 +1,62 @@
...
+#[serde(transparent)]
+pub struct TokenAmountDe(#[serde(with = "self")] pub TokenAmount);
+
+/// Serializes TokenAmount as bytes following Filecoin spec.
+pub fn serialize<S>(token: &TokenAmount, serializer: S) -> Result<S::Ok, S::Error>
+where
+    S: serde::Serializer,
+{
+    // review note: safe to deref? should we have a second impl of from with pointer rcvr?
+    let int = crate::econ::TokenAmount::from(*token);
```
Stebalien, line 27: It's always safe unless you use `unsafe`.

shared/src/sys/tokenamount_ser.rs
```diff
@@ -0,0 +1,62 @@
...
+    // let int: crate::econ::TokenAmount = token.into();
+    let (sign, mut bz) = int.to_bytes_be();
```
Stebalien, line 29: I'd actually implement this manually:

1. If zero, return an empty buffer.
2. If non-zero, write a single byte (positive), then `int.to_be_bytes()` (after removing the leading zero bytes).

shared/src/sys/tokenamount_ser.rs
```diff
@@ -0,0 +1,62 @@
...
+    let sign: Sign = match sign_byte {
+        1 => Sign::Minus,
+        0 => Sign::Plus,
+        _ => {
+            return Err(serde::de::Error::custom(
+                "First byte must be valid sign (0, 1)",
+            ));
+        }
+    };
+    Ok(BigInt::from_bytes_be(sign, &bz[1..]).try_into().unwrap())
```
Stebalien, line 61: This will panic on an invalid token amount. Don't unwrap.

Stebalien (changes requested):
Partial review:

1. Let's get rid of all the unwraps. In general, they're best used in tests, and cases where you actually _want_ to crash when something has gone wrong (which is basically never in Filecoin).
2. Is there any way to accept a block with a message that tries to send more than the max FIL?

Stebalien:
Closing due to conflicts. We ended up not doing this so we could ship integration faster, but will likely revisit it in the future.

PR #291, opened 2022-01-21
Remove all uses of hash maps in actors

### LINKED ISSUE #232: Figure out hash-map security
Most modern languages use a random seed in hash-maps to prevent the DoS attack where an attacker could turn `O(1)` operations into `O(n)` operations. However, because we're operating in a deterministic environment, we can't do that.

Note 1: This only impacts system actors. Basically, we need to make sure that a system actor never inserts untrusted data as keys into a HashMap. It's fine if, e.g., a miner actor uses a hash map when called directly by a user.

Note 2: We _should not_ do anything like use chain randomness here. That would just increase the attack surface to non-system actors (a malicious miner could grind on a message till it uses hash maps in such a way that it runs out of gas).

Proposed solution: use btrees and sorted slices as much as possible. These are always log(n).

Stebalien:
HashMaps/Sets have worst-case O(N) runtime for most operations. Runtime perf is _expected_ to be `O(1)`, but that makes randomness assumptions. Unfortunately, in a deterministic system (e.g., actors compiled to  WASM), we can't make these assumption (in the presence of malicious agents, at least).

Basically, I'm concerned an attacker may be able to pick "keys" (addresses, sector infos, etc.) such that some `O(n)` algorithm turns into an `O(n^2)` algorithm.

In practice:

1. This doesn't _really_ matter for _most_ actors. It only matters for "shared" actors where some malicious party might influence the state in such a way that some other actor (e.g., cron, in our case), might get stuck.
2. This could be significantly mitigated by seeding HashMaps with chain randomness. It would still be possible to pull off attacks, but the attacks would likely be expensive and would maybe result in a single small block delay. However, seeding HashMaps with randomness may impact gas estimation.

Overall, it's likely better to just "fix" this. In nearly all cases, the changes here shouldn't noticably impact the runtime because the theoretical runtime was _already_ `O(n log(n))`. In some cases, these changes may actually improve it.

Fixes #232.

### DIFF
--- a/actors/market/src/lib.rs
+++ b/actors/market/src/lib.rs
@@ -1,7 +1,7 @@
 // Copyright 2019-2022 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
 
-use std::collections::{HashMap, HashSet};
+use std::collections::{BTreeMap, BTreeSet};
 
 use actors_runtime::runtime::{ActorCode, Runtime};
 use actors_runtime::{
@@ -9,7 +9,6 @@ use actors_runtime::{
     CALLER_TYPES_SIGNABLE, CRON_ACTOR_ADDR, MINER_ACTOR_CODE_ID, REWARD_ACTOR_ADDR,
     STORAGE_POWER_ACTOR_ADDR, SYSTEM_ACTOR_ADDR, VERIFIED_REGISTRY_ACTOR_ADDR,
 };
-use ahash::AHashMap;
 use bitfield::BitField;
 use fvm_shared::address::Address;
 use fvm_shared::bigint::BigInt;
@@ -22,7 +21,7 @@ use fvm_shared::error::ExitCode;
 use fvm_shared::piece::PieceInfo;
 use fvm_shared::reward::ThisEpochRewardReturn;
 use fvm_shared::sector::StoragePower;
-use fvm_shared::{MethodNum, METHOD_CONSTRUCTOR, METHOD_SEND};
+use fvm_shared::{ActorID, MethodNum, METHOD_CONSTRUCTOR, METHOD_SEND};
 use log::info;
 use num_derive::FromPrimitive;
 use num_traits::{FromPrimitive, Signed, Zero};
@@ -297,10 +296,10 @@ impl Actor {
         let (network_raw_power, _) = request_current_network_power(rt)?;
 
         // Drop invalid deals
-        let mut proposal_cid_lookup = HashSet::new();
+        let mut proposal_cid_lookup = BTreeSet::new();
         let mut valid_proposal_cids = Vec::new();
         let mut valid_deals = Vec::with_capacity(params.deals.len());
-        let mut total_client_lockup: HashMap<Address, TokenAmount> = HashMap::new();
+        let mut total_client_lockup: BTreeMap<ActorID, TokenAmount> = BTreeMap::new();
         let mut total_provider_lockup = TokenAmount::zero();
 
         let mut valid_input_bf = BitField::default();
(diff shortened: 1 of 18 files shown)


actors/market/src/lib.rs
```diff
@@ -340,7 +339,10 @@ impl Actor {
             };
 
             // drop deals with insufficient lock up to cover costs
-            let lockup = total_client_lockup.entry(client).or_default();
+            let client_id = client
```
laudiacay, line 342: these are guaranteed unique?

actors/miner/src/lib.rs
```diff
@@ -1650,17 +1646,21 @@ impl Actor {
                 .map_err(|e| e.wrap("failed to load deadlines"))?;
 
             // Group declarations by deadline, and remember iteration order.
-            let mut decls_by_deadline = HashMap::<u64, Vec<ExpirationExtension>>::new();
+            //
+            // Vec::new() is a const expr, so this operation here is free (well, it'll add 576 bytes
+            // to the compiled binary and that'll get copied to the stack, but that should be pretty
+            // cheap.
+            const EMPTY_VEC: Vec<ExpirationExtension> = Vec::new();
+            let mut decls_by_deadline = [EMPTY_VEC; WPOST_PERIOD_DEADLINES as usize];
```
laudiacay, line 1654: why this way?

laudiacay (approved):

laudiacay:
pending "this is fine" as answers to questions above, LGTM

actors/market/src/lib.rs
Stebalien: Yeah, IDs are unique.

actors/miner/src/lib.rs
Stebalien: Well... I wanted to avoid the overhead of using a btreemap when we had a small number of elements anyways.

raulk (commented):
Like discussed sync, this PR would really benefit from a quantitative analysis of the concrete tradeoffs (A/B benchmarking) being identified:

- at which scale does the hashing operation in hashmaps start weighing on perf and gas
- at which scale do performance characteristics cross over
- the actual difficulty of degrading performance through the proposed vector (how practical is it, or is it only theoretical?)
- how much gas would be necessary to degrade performance of a hashmap in the way proposed
- for the sizes of maps we expect to see, would in average b-trees perform worse than hashmaps?

Stebalien:
Well, sticking to hash-maps is a security quagmire so I don't think there's much point in debating it. I can't think of a world where we _don't_ makt this change. I'd:

1. Merge this.
2. Profile as we sync with mainnet.
3. Optimize as necessary.

laudiacay:
drand and a cronjob to periodically re key every single hashmap in the
runtime

Stebalien:
> drand and a cronjob to periodically re key every single hashmap in the

Well, we don't store them in state so that isn't really an issue. The issue here is that, even _with_ randomness, the block producer could, e.g., make cron slow.

I'll admit that attacks like this are likely really hard to pull off, but:

1. I'd be very surprised if using btrees significantly reduced performance.
2. I just don't want to have to worry about it.

Stebalien:
The benchmarks are pretty noise, but I was able to get some basic numbers with mocked crypto and an in-memory store:

- ~10% improvement to aggregate commit (reasonably reliable improvement, possibly due to changes in bitfields?).
- No change in a large PublishStorageDeals
- No change for DeclareFaultsRecovered
- Maybe 1%-2% improvement for SubmitWindowedPost
- Maybe 2% improvement in _toy_ (specs-actors based) cron test-case.

The main things missing are a real cron test and gas-based benchmarking. But I think we have enough information to merge.

Stebalien:
My unscientific testing on mainnet didn't show a perf regression.

raulk (approved):

actors/market/src/lib.rs
```diff
@@ -1,15 +1,14 @@
 // Copyright 2019-2022 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
 
-use std::collections::{HashMap, HashSet};
+use std::collections::{BTreeMap, BTreeSet};
```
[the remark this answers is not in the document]
Stebalien: #316

Stebalien:
I've been running this all night and haven't had any sync issues.

PR #293, opened 2022-01-25
verify consensus fault extern will need to have the miner address as param

arajasek:
the extern provider (lotus) will have no way of knowing the miner otherwise

### DIFF
--- a/sdk/src/crypto.rs
+++ b/sdk/src/crypto.rs
@@ -22,13 +22,12 @@ pub fn verify_signature(
     let signature = signature
         .marshal_cbor()
         .expect("failed to marshal signature");
-    let signer = signer.marshal_cbor().expect("failed to marshal address");
     unsafe {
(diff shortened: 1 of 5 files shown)


sdk/src/crypto.rs
```diff
@@ -22,13 +22,12 @@ pub fn verify_signature(
...
         .marshal_cbor()
         .expect("failed to marshal signature");
-    let signer = signer.marshal_cbor().expect("failed to marshal address");
     unsafe {
         sys::crypto::verify_signature(
             signature.as_ptr(),
             signature.len() as u32,
-            signer.as_ptr(),
-            signer.len() as u32,
+            signer.to_bytes().as_ptr(),
```
arajasek, line 29: unrelated change

arajasek:
Oh nvm, i think this isn't needed, ignore me

PR #294, opened 2022-01-25
WIP: Add cgo bindings for externs

arajasek:
~Depends on #281~ 

This fixes up most of the externs stuff within the FVM. The next step is probably to extract the CGO side of things into the FFI repo (branch asr/fvm), and fully integrate with Lotus there (latest branch feat/fvm).

This isn't compiling locally for me because I seem to have run into more cargo bugs, will investigate more.


Stebalien:
I'm switching this back to master as switching the token type will require quite a bit more work to handle all the potential error cases (unwrapping isn't going to work).

Stebalien:
Replacing with #305.

PR #295, opened 2022-01-25
various fixes to the VM found on the new conformance tests

Stebalien:
1. Check the exit code in the actors runtime, in addition to checking the syscall result.
2. Verify signatures in conformance tests.
3. Correctly marshal the signer address in the sdk (don't use CBOR).
4. Avoid panicking when passing non-utf8 log messages (drive-by fix).
5. Fix parameter order in some crypto syscalls.

Unfortunately, I'm not updating the conformance tests in this PR due to some outstanding issues.

### DIFF
--- a/actors/runtime/src/runtime/fvm.rs
+++ b/actors/runtime/src/runtime/fvm.rs
@@ -238,7 +238,13 @@ where
         if self.in_transaction {
             return Err(actor_error!(SysErrIllegalActor; "runtime.send() is not allowed"));
         }
-        Ok(fvm::send::send(&to, method, params, value)?.return_data)
+        // TODO: distinguish between "syscall" errors and actor exit codes.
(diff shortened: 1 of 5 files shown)


Stebalien:
To make it further, we need to either:

1. Include the proof verification results inside the test vectors.
2. Download the proof params in CI.

raulk (approved):

raulk:
> 1. Include the proof verification results inside the test vectors.

For these vectors, that isn't the problem because they run just fine in Lotus. But yes, we need that solution to test scenarios where proof validation fails. However, vectors that do that failed extraction and are not included here.

> 2. Download the proof params in CI.

Yep.

Stebalien:
Ah... the problem was gas. We charge that _inside_ the kernel.

PR #296, opened 2022-01-25
ci: allow custom package installation for cargo

galargh:
The new target in the makefile is not needed because testing `fvm_token_actor` package is covered by the check named `test`. To make it more explicit and faster, I extracted `fvm_token_actor` tests to another job.

I also made it possible to install additional packages per `cargo` target in the workflow. Then, I used it to install `ocl-icd-opencl-dev` in the new `test-fvm-token-actor` job.

@raulk heads up, looks like there is some flakiness in the tests. One of the CI runs failed as follows: https://github.com/filecoin-project/ref-fvm/runs/4935482784?check_suite_focus=true 

###### Testing
- [x] CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -36,7 +36,10 @@ jobs:
             args: --package fvm --no-default-features
           - name: test
             command: test
-            args: --all --exclude fvm --exclude conformance_tests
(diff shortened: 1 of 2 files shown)


raulk (approved):
Thank you!

PR #299, opened 2022-01-26
update conformance vectors

Stebalien:
There were no more bugs in the FVM itself but this PR fixes a few bugs in the conformance test system:

1. Charge gas even for operations that aren't forwarded.
2. Correctly decode the circulating supply in test vectors (doesn't fit in an f64).

Additionally, it improves debuggability a bit:

1. Traces gas charges.
2. Traces value-only sends.
3. Make state-diffing work with partial states (i.e., just diff the relevant actors).
4. Very basic actor state diffing (diffs the root object as cbor).

### DIFF
--- a/testing/conformance/src/vector.rs
+++ b/testing/conformance/src/vector.rs
@@ -45,9 +45,9 @@ pub struct MetaData {
 pub struct PreConditions {
     pub state_tree: StateTreeVector,
     #[serde(default)]
-    pub basefee: Option<f64>,
+    pub basefee: Option<u128>,
     #[serde(default)]
-    pub circ_supply: Option<f64>,
+    pub circ_supply: Option<u128>,
     #[serde(default)]
     pub variants: Vec<Variant>,
 }
@@ -112,13 +112,6 @@ pub struct RandomnessRule {
     pub epoch: ChainEpoch,
     #[serde(with = "base64_bytes")]
     pub entropy: Vec<u8>,
-}
-
-#[derive(Debug, Deserialize)]
-#[serde(tag = "class")]
-pub enum TestVector {
-    #[serde(rename = "message")]
-    Message(MessageVector),
 }
 
 #[derive(Debug, Deserialize)]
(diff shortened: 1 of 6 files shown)


testing/conformance/src/vector.rs
```diff
@@ -45,9 +45,9 @@ pub struct MetaData {
 pub struct PreConditions {
     pub state_tree: StateTreeVector,
     #[serde(default)]
-    pub basefee: Option<f64>,
+    pub basefee: Option<u128>,
     #[serde(default)]
-    pub circ_supply: Option<f64>,
```
raulk, line 50: @noot FYI -- some of the new vectors might fail with Forest's driver due to this.

testing/conformance/tests/runner.rs
```diff
@@ -108,58 +108,85 @@ fn check_msg_result(expected_rec: &Receipt, ret: &ApplyRet, label: impl Display)
...
-                log::error!("actor {} has state {:?}, expected {:?}", k, va, ve)
-            }
-            None => log::error!("missing actor {}", k),
-            _ => {}
-        }
+    Ok(())
+}
+
+/// Compares the resulting state root with the expected state root. Currently,
+/// this doesn't do much, but it could run a statediff.
```
raulk, line 147: nit: should probably update the comment since we're doing a lot more now.

testing/conformance/tests/runner.rs
```diff
@@ -276,10 +303,33 @@ async fn run_vector(
...
 
-    match vector {
-        TestVector::Message(v) => {
+    // Test vectors have the form:
+    //
+    //     { "class": ..., rest... }
+    //
+    // Unfortunately:
+    // 1. That means we need to use serde's "flatten" and/or "tag" feature to decode them.
+    // 2. Serde's JSON library doesn't support arbitrary precision numbers when doing this.
```
raulk, line 313: That's unfortunate.

raulk: Is there an issue for this upstream?

raulk (approved):

testing/conformance/tests/runner.rs
Stebalien: I believe https://github.com/serde-rs/serde/issues/1183 is best, but there are a bunch of related issues. But I'll link to that one.

PR #302, opened 2022-01-27
fvm: remove error tracing from the kernel

Stebalien:
1. Remove error recording logic from the kernel entirely.
2. Remove error recording methods from the call manager.

This simplifies both and may allow us to store additional information now that the call manager handles everything internally.

TODO:

- Propagate abort messages.
- Consider recording additional information.


Stebalien:
@raulk I'm going to take a cleanup pass but I wanted feedback before I took this any further. Basically, it has always annoyed me how:

1. The kernel returns some error.
2. We turn back around and tell the kernel to store the error.
3. The kernel just tells the call manager to store the error.

Now, we can bypass the kernel entirely and the kernel can focus on filecoin relevant function instead of error tracing.

raulk (commented):
This is LGTM so far, @Stebalien.
- Detaching the error form the kernel makes a ton of sense, as it belongs to the outer context (call manager).
- `InvocationData` as a term is a good choice too, as it aligns well with `InvocationContainer`.
- I like encapsulating the behaviour in a Backtrace object.

In a nutshell, big thumbs up.

raulk:
> Consider recording additional information.

What additional info are you thinking of?

Stebalien:
> What additional info are you thinking of?

- Message/syscall parameters.
- Maybe a backtrace if an actor aborts (when debugging is enabled). I'm not sure what we can do here but it looks like wasmtime might capture something in the trap.

But not in _this_ PR.

Stebalien:
Ok, so I ended up doing a few things:

1. I did add the method number & params so we now get a nice stack trace.
2. I've switched to stack frames (actor exits) and an ultimate cause (syscall error).
3. I've introduced a new `Abort` type for passing aborts through wasmtime. This forced me to restructure the send logic a bit, but doing so caught a bug (we were returning a syscall error instead exiting with a system exit code if an actor returned an invalid block handle as the return value).

fvm/src/call_manager/default.rs
```diff
@@ -285,57 +277,113 @@ where
...
-                    super::NO_DATA_BLOCK_ID
-                };
+            let mut store = Store::new(&engine, InvocationData::new(kernel));
+
+            // Store parameters, if any.
+            let param_id = if params.len() > 0 {
+                match store.data_mut().kernel.block_create(DAG_CBOR, params) {
+                    Ok(id) => id,
+                    // This could fail if we pass some global memory limit.
+                    Err(err) => return (Err(err), store.into_data().kernel.take()),
```
Stebalien, line 291: Having to do this manually is _kind_ of annoying, but I like having a clear cutoff where consider the syscall to have succeeded.

fvm/src/call_manager/default.rs
```diff
@@ -285,57 +277,113 @@ where
...
+            let result: std::result::Result<RawBytes, Abort> = (|| {
                 // Instantiate the module.
-                let instance = linker.instantiate(&mut store, &module).or_fatal()?;
+                let instance = linker
+                    .instantiate(&mut store, &module)
+                    // The module should already have been checked by the system.
+                    .map_err(Abort::Fatal)?;
+
+                // Lookup the invoke method.
+                let invoke: wasmtime::TypedFunc<(u32,), u32> = instance
```
Stebalien, line 306: I just moved the type up here and made it explicit because I get nervous about these things.

Stebalien:
(note: we should squash before merging)

fvm/src/call_manager/default.rs
raulk: Paranoia pays off ;-)

fvm/src/executor/mod.rs
```diff
@@ -43,18 +44,23 @@ impl ApplyRet {
         message: impl Into<String>,
         miner_penalty: BigInt,
     ) -> ApplyRet {
+        // This is a little funky, but it's the simplest way to record a sane error message.
+        let mut bt = Backtrace::default();
+        bt.set_cause(Cause {
+            module: "system",
+            function: "execute",
+            error: ErrorNumber::IllegalOperation,
```
raulk, line 52: This is a bit weird, as we're picking this error from the syscall namespace?

fvm/src/syscalls/vm.rs
```diff
@@ -1,50 +1,42 @@
...
-use wasmtime::Trap;
 
-use super::error::{trap_from_code, trap_from_error};
+use super::error::Abort;
 use super::Context;
-use crate::kernel::{ClassifyResult, ExecutionError};
+use crate::kernel::{ClassifyResult, Context as _};
 use crate::Kernel;
 
+/// An unihabited type. We use this in `abort` to make sure there's no way to return without
```
raulk, line 9: nit

```suggestion
/// An uninhabited type. We use this in `abort` to make sure there's no way to return without
```

raulk (approved):

fvm/src/executor/mod.rs
Stebalien: Yeah, I agree. I think the answer is to introduce a new type:

```rust
enum MessageFailure {
     PreValidation(...),
     Failure(Backtrace),
}
```
```

Stebalien: How does it look now?

raulk (approved):

Stebalien:
And... I forgot to squash. Oh well.

PR #305, opened 2022-01-31
Move CGO/FFI logic to filecoin-ffi

Stebalien:
See https://github.com/filecoin-project/filecoin-ffi/pull/230.


raulk (commented):
Doing this LGTM; just need to adjust build targets and maybe CircleCI config.

PR #307, opened 2022-02-04
actors v7: implement FIP-19, SnapDeals

elmattic:
Implements FIP-19.

I still need to add unit tests once merged (on `forest` side).

Maybe the tests with the FVM-enabled `Lotus` could be used in the meantime.


Stebalien:
This has been rolled into #349.

PR #308, opened 2022-02-07
feat: caching wasmtime engine

Stebalien:
At the moment, the cache is pretty dumb and doesn't let us "precompile". But it's a place to start and it solves the core performance problem.

### DIFF
--- a/fvm/src/machine/default.rs
+++ b/fvm/src/machine/default.rs
@@ -9,9 +9,9 @@ use fvm_shared::version::NetworkVersion;
 use fvm_shared::ActorID;
 use log::debug;
 use num_traits::{Signed, Zero};
-use wasmtime::{Engine, Module};
+use wasmtime::Module;
 
-use super::{Machine, MachineContext};
+use super::{Engine, Machine, MachineContext};
 use crate::blockstore::BufferedBlockstore;
 use crate::externs::Externs;
 use crate::gas::price_list_by_epoch;
@@ -51,6 +51,7 @@ where
     #[allow(clippy::too_many_arguments)]
     pub fn new(
         config: Config,
+        engine: Engine,
         epoch: ChainEpoch,
         base_fee: TokenAmount,
         base_circ_supply: TokenAmount,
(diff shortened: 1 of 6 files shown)


fvm/src/machine/default.rs
```diff
@@ -166,6 +164,10 @@ where
 
     #[cfg(feature = "builtin_actors")]
     fn load_module(&self, code: &Cid) -> Result<Module> {
+        if let Some(code) = self.engine.get(code) {
+            return Ok(code);
+        }
+
         use anyhow::Context;
         // TODO: cache compiled code, and modules?
```
raulk, line 172: Can probably remove this TODO.

fvm/src/machine/default.rs
```diff
@@ -166,6 +164,10 @@ where
 
     #[cfg(feature = "builtin_actors")]
     fn load_module(&self, code: &Cid) -> Result<Module> {
+        if let Some(code) = self.engine.get(code) {
```
raulk, line 167: Comment would be great here.

testing/conformance/tests/runner.rs
```diff
@@ -190,11 +190,13 @@ fn compare_state_roots(bs: &MemoryBlockstore, root: &Cid, vector: &MessageVector
 async fn conformance_test_runner() -> anyhow::Result<()> {
     pretty_env_logger::init();
 
+    let engine = Engine::default();
```
raulk, line 193: For M1, do we intend to have Lotus create a single engine for the full process lifetime?

raulk (approved):
LGTM as a starting point. For M1 it's probably ok for the client to instantiate a singleton Engine for the full lifetime of the process. And it's also OK for that engine to hold the compiled modules in memory, because it's only the built-in actors.

I think we should prioritise precompilation. It's a better experience if built-in actors are compiled as part of node initialization instead of on the first message that touches them.

Stebalien:
> I think we should prioritise precompilation. It's a better experience if built-in actors are compiled as part of node initialization instead of on the first message that touches them.

Agreed.

testing/conformance/tests/runner.rs
Stebalien: Yes, the FFI branch currently does that (stores it in a lazy static).

PR #320, opened 2022-02-11
feat: remove patches

Stebalien:
We don't really need them for workspaces. The only patch we need is for CIDs.


Stebalien:
I'm just going to go ahead and merge this as it's just a dep change.

PR #323, opened 2022-02-12
actors: fix wasm build flags

### LINKED ISSUE #322: rustflags set for wasm-builder is actually not working
rustflags set like [https://github.com/filecoin-project/ref-fvm/blob/master/actors/miner/build.rs#L6-L10](https://github.com/filecoin-project/ref-fvm/blob/master/actors/miner/build.rs#L6-L10) is actually not working (at least for  `1.60.0-nightly` or `nightly-2021-12-14` which l've already tested)

the latest [substrate-wasm-builder 5.0.0-dev in it's git repo](https://github.com/paritytech/substrate/tree/master/utils/wasm-builder) works fine.

and also, some flags are incorrect(or out of date?) like `-Coverflow-checks=true` and `-Clto=true`, all the `true`s should be switched to `yes`

Stebalien:
- Use substrate-wasm-builder (upstream) not wasm-builder (a fork).
- Use the latest git to make sure the flags are actually applied
- Fix the flags:
    - overflow-checks=yes
    - lto=thin (I tried fat, but the build times were significantly and the binary sizes were no smaller)

fixes #322

### DIFF
--- a/actors/account/build.rs
+++ b/actors/account/build.rs
@@ -2,11 +2,10 @@ fn main() {
     use wasm_builder::WasmBuilder;
     WasmBuilder::new()
         .with_current_project()
-        .import_memory()
         .append_to_rust_flags("-Ctarget-feature=+crt-static")
         .append_to_rust_flags("-Cpanic=abort")
-        .append_to_rust_flags("-Coverflow-checks=true")
-        .append_to_rust_flags("-Clto=true")
+        .append_to_rust_flags("-Coverflow-checks=yes")
+        .append_to_rust_flags("-Clto=thin")
         .append_to_rust_flags("-Copt-level=z")
         .build()
 }
--- a/actors/account/Cargo.toml
+++ b/actors/account/Cargo.toml
@@ -25,6 +25,6 @@ runtime-wasm = ["actors-runtime/runtime-wasm"]
 
 
 [build-dependencies]
-wasm-builder = "3.0.1"
+wasm-builder = { package = "substrate-wasm-builder", git = "https://github.com/paritytech/substrate" }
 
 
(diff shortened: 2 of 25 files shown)


dignifiedquire:
is the ci failing expected?

Stebalien:
Hm. No. It appears that the dev version, at least, is producing some form of compressed wasm.

actors/account/build.rs
```diff
@@ -5,8 +5,8 @@ fn main() {
         .import_memory()
```
dtynn, line 5: i think all the `import_memory` should be removed, otherwise the generated wasm would ask for importing `env::memory` instead of exporting memory module, like:
`(import "env" "memory" (memory (;0;) 16))` vs `(export "memory" (memory 0))`

dtynn:
> Hm. No. It appears that the dev version, at least, is producing some form of compressed wasm.

bianry is [compressed](https://github.com/paritytech/substrate/blob/master/utils/wasm-builder/src/wasm_project.rs#L633-L650) for non `dev` profile, and wasm-builder uses `release` by default.

add 
```
[dependencies]
sp-maybe-compressed-blob = { git = "https://github.com/paritytech/substrate.git", rev = "0a92e3e" }
```

and use decompress func like this
```
        let engine = Engine::default();

        let bin = actor::wasm::WASM_BINARY.expect("get compact wasm binary");
 
        let raw = sp_maybe_compressed_blob::decompress(
            bin,
            sp_maybe_compressed_blob::CODE_BLOB_BOMB_LIMIT,
        )
        .expect("decompress bin");

        let module = Module::from_binary(&engine, raw.as_ref()).expect("load compact module");
```
could solve the problem.

Stebalien:
@dtynn thanks for your help on this!

raulk (approved):

PR #326, opened 2022-02-14
chore: switch to stable rust

### LINKED ISSUE #325: set proper MSRV and default toolchain
both are currently `nightly` which is a moving target.

default toolchain should be `stable` and MSRV should be a version string that is released (not nightly).

ref https://github.com/ChainSafe/forest/pull/1228

Stebalien:
fixes #325

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,7 @@ This repository contains the reference implementation of the Filecoin VM ([specs
 
 ## Build requirements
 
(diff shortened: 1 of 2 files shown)


Stebalien:
It looks like wasm-builder requires nightly. And as far as I can tell, that requirement is entirely arbitrary.

arajasek (approved):

PR #327, opened 2022-02-14
FVM fixes for lotus integration

Stebalien:
Lotus integration isn't quite bug-free, but there's no reason to keep these fixes in a separate branch while we continue working on that.

### DIFF
--- a/fvm/src/kernel/default.rs
+++ b/fvm/src/kernel/default.rs
@@ -33,6 +33,7 @@ use crate::externs::{Consensus, Rand};
 use crate::gas::GasCharge;
 use crate::market_actor::State as MarketActorState;
 use crate::power_actor::State as PowerActorState;
+use crate::reward_actor::State as RewardActorState;
 use crate::state_tree::ActorState;
 use crate::syscall_error;
 
@@ -151,6 +152,11 @@ where
             )
             .ok_or_else(|| anyhow!("failed to subtract"))
             .or_fatal()
+    }
+
+    fn get_fil_mined(&self) -> Result<TokenAmount> {
+        let (reward_state, _) = RewardActorState::load(self.call_manager.state_tree())?;
+        Ok(reward_state.total_storage_power_reward())
     }
 
     fn power_locked(&self) -> Result<TokenAmount> {
@@ -379,7 +385,10 @@ where
     fn total_fil_circ_supply(&self) -> Result<TokenAmount> {
         self.call_manager
             .context()
-            .base_circ_supply
+            .fil_vested
+            .checked_add(&self.get_fil_mined()?)
+            .ok_or(anyhow!("overflow when adding mined fil to vested fil"))
+            .or_fatal()?
             .checked_add(&self.get_reserve_disbursed()?)
             .ok_or(anyhow!(
                 "overflow when adding reserve to base circulating supply"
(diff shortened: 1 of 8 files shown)


fvm/src/kernel/default.rs
```diff
@@ -138,19 +139,20 @@ where
...
-                    .call_manager
-                    .state_tree()
-                    .get_actor_id(RESERVE_ACTOR_ID)?
-                    .ok_or_else(|| anyhow!("reserve actor state couldn't be loaded"))
-                    .or_fatal()?
-                    .balance,
-            )
-            .ok_or_else(|| anyhow!("failed to subtract"))
-            .or_fatal()
+        let initial_reserve_balance = BigInt::from(300_000_000) * FILECOIN_PRECISION;
```
raulk, line 142: Not introduced here, but should probably make this number a named const.

fvm/src/kernel/default.rs
```diff
@@ -377,23 +379,13 @@ where
...
-            .or_fatal()?
-            .checked_sub(&self.market_locked()?)
-            .ok_or(anyhow!("underflow when subtracting market locked funds"))
-            .or_fatal()
+        Ok((&self.call_manager.context().fil_vested
+            + &self.get_fil_mined()?
+            + &self.get_reserve_disbursed()?
+            - &self.get_burnt_funds()?
+            - &self.power_locked()?
+            - &self.market_locked()?)
```
raulk, line 387: Can we add `map_err` here to add a context to these errors, like previously? Although probably not worth it because these errors will actually never happen.

arajasek (approved):

fvm/src/machine/default.rs
```diff
@@ -64,10 +64,15 @@ where
             "initializing a new machine, epoch={}, base_fee={}, nv={:?}, root={}",
             epoch, &base_fee, network_version, state_root
         );
+
+        if network_version != NetworkVersion::V14 {
+            return Err(anyhow!("unsupported network version: {}", network_version));
+        }
```
raulk, line 70: Better change to a `<`, since we'll soon support V15? (and the FVM is really designed to support V14 and above).

fvm/src/reward_actor.rs
```diff
@@ -0,0 +1,94 @@
+use anyhow::Context;
```
raulk, line 1: Ouch.

raulk (approved):
LGTM. Most are nits, except for the network version guard, which I think should really be a less than in order to future-proof.

fvm/src/kernel/default.rs
Stebalien: Previously, we were adding context to overflow errors (which couldn't happen), not the other errors. But I'll go ahead and change the actual functions to add context to their errors.

fvm/src/machine/default.rs
Stebalien: I'd rather _explicitly_ decide with network versions we support. For example, to support nv15 we'll need support for multiple actor versions.

fvm/src/reward_actor.rs
Stebalien: Yeah... But we'll be able to drop this once we drop support for nv14.

fvm/src/machine/default.rs
arajasek: I concur with @Stebalien here

Issue #319, opened 2022-02-11
Gas mismatches on internal IPLD gets


Stebalien:
The FVM currently only charges IPLD lookup gas when the _user_ asks for a block. However, lotus will (usually) charge this gas when some syscall internally needs to lookup some state:

- If VerifySignature needs to lookup a key (e.g., if it's passed an ID address), it'll charge to read the signer's state.
- NewActorAddress doesn't charge gas to lookup the senders "key" address. Honestly, this just looks like an oversight.
- Loading actors and resolving addresses for consensus faults charges for every IPLD access.

That last part is, to say the least, annoying, because we rely on lotus to handle consensus fault detection.

Any solution here is going to be a bit ugly.

- For syncing with mainnet:
    - We can make VerifySignature and NewActorAddress "work".
    - We can get lotus to... do internal gas accounting and tell us how much additional gas to charge.
- For M1, we can revisit this and try to find a nicer solution.

raulk:
_I had drafted this comment yesterday, but never posted it. I see you arrived to similar conclusions to mine, @Stebalien._

The `verify_consensus_fault` situation is unfortunate. The issue with returning gas to be charged ex post facto, is that it can lead to deferred execution halting, which is highly undesirable.

Some ideas:
1. Expose the gas charger object from the FVM to the node; the node can then cross the FFI boundary to charge.
2. Simpler: on every Extern call, move the remaining gas value to the node as an input parameter, and expect the node to return the new remaining gas on return. If the FFI returns in error and the remaining gas is zero, it means that it ran out of gas.
3. Redesign the `verify_consensus_fault` logic so that the core logic can live in the FVM, with a more atomized Extern FFI interface for specific operations. But I'm afraid that won't help because the lookback will probably still be one big FFI operation resulting in many gas charges depending on IPLD accesses.

I think (2) is probably best here. The downside is that we'd be splitting the gas charging logic across two boundaries, and now the node needs to know details about the gas pricing too.

Stebalien:
Actually, now that I think about it, I _don't_ think gas accounting is an issue. _Where_ an actor runs out of gas doesn't matter as long as it either consistently runs out of gas or doesn't.

But yeah, option 2 (tell the system how much gas is remaining) is likely the better fix.

Stebalien:
> NewActorAddress doesn't charge gas to lookup the senders "key" address. Honestly, this just looks like an oversight but it doesn't really matter.

[...]

PR #330, opened 2022-02-16
allow the verify consensus fault extern to return gasused

arajasek:
Addresses #319

### DIFF
--- a/shared/src/consensus/mod.rs
+++ b/shared/src/consensus/mod.rs
@@ -13,16 +13,6 @@ pub struct ConsensusFault {
     pub fault_type: ConsensusFaultType,
 }
 
-/// Result of checking two headers for a consensus fault, with the gas used
-/// Only for v14 and earlier
-#[derive(Clone, Debug)]
-pub struct ConsensusFaultWithGas {
-    /// The fault.
-    pub fault: Option<ConsensusFault>,
-    /// Gas used in checking the fault
-    pub gas_used: i64,
-}
-
 /// Consensus fault types in VM.
 #[derive(FromPrimitive, Clone, Copy, Debug)]
 #[repr(u8)]
--- a/shared/src/consensus/mod.rs
+++ b/shared/src/consensus/mod.rs
@@ -13,6 +13,16 @@ pub struct ConsensusFault {
     pub fault_type: ConsensusFaultType,
 }
 
+/// Result of checking two headers for a consensus fault, with the gas used
+/// Only for v14 and earlier
+#[derive(Clone, Debug)]
+pub struct ConsensusFaultWithGas {
+    /// The fault.
+    pub fault: Option<ConsensusFault>,
+    /// Gas used in checking the fault
+    pub gas_used: i64,
+}
+
 /// Consensus fault types in VM.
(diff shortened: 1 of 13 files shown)


actors/runtime/src/runtime/fvm.rs
```diff
@@ -358,12 +359,22 @@ where
...
-        &self,
+        &mut self,
         h1: &[u8],
         h2: &[u8],
         extra: &[u8],
     ) -> Result<Option<ConsensusFault>, Error> {
-        fvm::crypto::verify_consensus_fault(h1, h2, extra).map_err(|_| Error::msg("no fault"))
+        let ret = fvm::crypto::verify_consensus_fault(h1, h2, extra)
+            .map_err(|_| Error::msg("failed to verify fault"))?;
+        self.charge_gas("verify_consensus_fault_accesses", ret.gas_used);
```
Stebalien, line 369: We should be charging this gas _inside_ the FVM Kernel, not inside the actor runtime.

shared/src/consensus/mod.rs
```diff
@@ -13,10 +14,25 @@ pub struct ConsensusFault {
...
+    pub fault_type: ConsensusFaultType,
+    /// Gas used in checking the fault
+    pub gas_used: i64,
+}
+
 /// Consensus fault types in VM.
 #[derive(FromPrimitive, Clone, Copy, Debug)]
 #[repr(u8)]
 pub enum ConsensusFaultType {
+    NoFault = 0,
```
Stebalien, line 35: I'm pretty strongly against a "no fault" variant as it means all the other fields in the `ConsensusFault` object will be garbage. We only do this kind of thing in go because go doesn't have enums/options.

Stebalien (changes requested):

Stebalien:
(basically, this change shouldn't have to touch the actors or actor runtime)

actors/runtime/src/runtime/fvm.rs
arajasek: Makes sense

shared/src/consensus/mod.rs
arajasek: synced on this: decision is to move the "no fault" (Go) -> Option (Rust) step into the FFI

shared/src/consensus/mod.rs
```diff
@@ -13,6 +13,16 @@ pub struct ConsensusFault {
...
 
+/// Result of checking two headers for a consensus fault, with the gas used
+/// Only for v14 and earlier
+#[derive(Clone, Debug)]
+pub struct ConsensusFaultWithGas {
+    /// The fault.
+    pub fault: Option<ConsensusFault>,
+    /// Gas used in checking the fault
+    pub gas_used: i64,
+}
```
Stebalien, line 24: 1. There's no real need for a named type here. We can just use `(Option<ConsensusFault>, i64)`.
2. This doesn't really belong in "shared" (shared is for types shared between actors and the FVM, not the FVM and lotus).

Stebalien (approved):
One nit, but LGTM.

Stebalien (approved):
lgtm

PR #334, opened 2022-02-16
fix: charge gas for state access when resolving a key address

Stebalien:
But don't charge when looking up the "origin" key, to be consistent with lotus.

partially addresses #319

### DIFF
--- a/fvm/src/kernel/default.rs
+++ b/fvm/src/kernel/default.rs
@@ -103,13 +103,14 @@ impl<C> DefaultKernel<C>
 where
     C: CallManager,
 {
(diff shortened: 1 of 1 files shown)


Stebalien:
NOTE: I believe a better fix is to:

1. Always charge gas here.
2. Pre-resolve the origin key per https://github.com/filecoin-project/ref-fvm/issues/319#issuecomment-1042300001.

But this is the faster fix and we can address that later once we decide.

Stebalien:
Testing: I've added a test vector to trigger this case.

arajasek (approved):
Gross but fine

PR #338, opened 2022-02-18
feat: remove a block copy on read

Stebalien:
By using a `Box` instead of an `Rc`. We had an `Rc` to facilitate sharing (immutable) blocks between actors, but we never used it for that _anyways_.

Also make the block fields private.


Stebalien:
For context:

- `Vec<u8>` -> `Box<u8>` only allocates if the vector has unused space (which would be unusual in this case).
- `Vec<u8>` -> `Rc<[u8]>` always allocates.

PR #339, opened 2022-02-18
feat: cache block existence

Stebalien:
When we read blocks from the underlying blockstore, cache that they exist. That way we can skip them in flush in case we write them back unchanged (which is, unfortunately, something we tend to do quite frequently).

Lotus does this by checking if we "have" blocks in the datastore before recursively copying some subdag. I'm trying to cheat here a bit to avoid the overhead of calling into cgo.


Stebalien:
Preliminary testing indicates that this is useless, but I'll let it run for a bit.

Stebalien:
Possibly slower...

PR #343, opened 2022-02-25
clean child in link node only if child is deleted

connormullett:
Ran into an issue when attempting to sync calibnet on Forest. Seems to prevent panics and allows the node to sync up to the network

### DIFF
--- a/ipld/hamt/src/node.rs
+++ b/ipld/hamt/src/node.rs
@@ -390,10 +390,10 @@ where
                 let deleted = child_node.rm_value(hashed_key, bit_width, depth + 1, key, store)?;
                 if deleted.is_some() {
                     *child = Pointer::Dirty(std::mem::take(child_node));
+                    // Clean to retrieve canonical form
+                    child.clean()?;
                 }
 
-                // Clean to retrieve canonical form
-                child.clean()?;
                 Ok(deleted)
             }
             Pointer::Dirty(n) => {


Stebalien:
> Ran into an issue when attempting to sync calibnet on Forest. Seems to prevent panics and allows the node to sync up to the network

Did you record the epoch at which this failed? We can extract a test vector with https://github.com/filecoin-project/lotus/tree/master/cmd/tvx.

connormullett:
> Did you record the epoch at which this failed? We can extract a test vector with https://github.com/filecoin-project/lotus/tree/master/cmd/tvx.

No, we tried from many different heights (potentially issues with network upgrades on our end) and ran into same results each time at different heights. We believe this to be a small logical error that went unnoticed for a while and only resurfaced once we started adding support for calibnet.

The forest PR that has this fix is here: https://github.com/ChainSafe/forest/pull/1370

This PR is mainly to bring awareness to what we were experiencing, show the fix, and potentially fix anything upstream.

Stebalien:
How did you arrive at this fix?

1. Looking at the matching go code, the old rust code was correct. It cleans the child on delete, whether or not the value was modified.
2. If this was the issue, it's concerning as `cleanChild` should be a no-op unless we've deleted something. Otherwise, we should already be "canonical".

Stebalien:
I think https://github.com/filecoin-project/ref-fvm/blob/e53d1416fab5b7f2787f884171ace4cb2b6edb9f/ipld/hamt/src/pointer.rs#L125 may be the issue? That doesn't look like an error in go:

https://github.com/filecoin-project/go-hamt-ipld/blob/801910a6d7f17c752c877075c56fea548ffa0460/hamt.go#L596

My guess is that the _parent_ is expected to handle that case and remove the empty child. But a test vector would be really helpful.

connormullett:
> How did you arrive at this fix?
> 
> 1. Looking at the matching go code, the old rust code was correct. It cleans the child on delete, whether or not the value was modified.
> 2. If this was the issue, it's concerning as `cleanChild` should be a no-op unless we've deleted something. Otherwise, we should already be "canonical".

1. We came to this fix after debugging in the VM as we had everything seemingly "correct" on our part. Our debug logs were showing 

`internal error: entered unreachable code: clean is only called on dirty pointer'`

which is the wildcard case in `clean()`. Moving where the child is cleaned led us to this fix

2. From the current implementation, it doesn't look like clean is a no op. If the child isn't a dirty pointer then it will panic with the `unreachable!` macro. Which is why our fix should be fine.

Stebalien:
Oh, I see. I have no idea how this hasn't been hit on mainnet in the past... Thanks for the fix and the explanation!

Stebalien (approved):

PR #353, opened 2022-03-01
fix: prove_replica_update gas-mismatch

Stebalien:
Usually, we should avoid hanging onto state between transactions because there might be a recursive call. Unfortunately, while _rust_ can statically prevent this, go doesn't.

For now, we hack around this by making it possible to copy the blockstore in go (where "copy" just references it). In nv16, we'll want to change this to avoid re-using state across calls.

### DIFF
--- a/actors/miner/src/lib.rs
+++ b/actors/miner/src/lib.rs
@@ -887,7 +887,8 @@ impl Actor {
         params: ProveReplicaUpdatesParams,
     ) -> Result<BitField, ActorError>
     where
-        BS: Blockstore,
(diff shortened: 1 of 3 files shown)


raulk (commented):
We should to transplant this PR onto https://github.com/filecoin-project/builtin-actors. But happy transplanting the entire base branch too.

raulk:
@arajasek mind confirming that this is absorbed in #360, and closing if that's the case?

arajasek:
@raulk Absorbed into https://github.com/filecoin-project/builtin-actors/pull/50 actually, but yes :)

PR #354, opened 2022-03-02
expose encode as a public method of Address

LesnyRumcajs:
In Forest we need this `String` representation for serialization purposes.

Technically we could use `.to_string()` method as `Address` is `Display`, but to my understanding a display string, in general, is allowed to have some fluff around it to make it human-readable. In other words, there's no guarantee someone won't add some extra content n `fmt` which would break the serialization.

### DIFF
--- a/shared/src/address/mod.rs
+++ b/shared/src/address/mod.rs
@@ -183,11 +183,36 @@ impl Address {
             _ => Err(Error::NonIDAddress),
         }
     }
+
+    /// encode converts the address into a string
+    pub fn encode(&self) -> String {
+        match self.protocol() {
+            Protocol::Secp256k1 | Protocol::Actor | Protocol::BLS => {
+                let ingest = self.to_bytes();
+                let mut bz = self.payload_bytes();
+
+                // payload bytes followed by calculated checksum
+                bz.extend(checksum(&ingest));
+                format!(
+                    "{}{}{}",
+                    self.network.to_prefix(),
+                    self.protocol(),
+                    ADDRESS_ENCODER.encode(bz.as_mut()),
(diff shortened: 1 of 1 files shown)


Stebalien:
> but to my understanding a display string, in general, is allowed to have some fluff around it to make it human-readable

I believe the convention is for "display" on string-like things to encode said string-like things correctly. For example:

- The `url` crate doesn't provide a separate encode. The docs say to use `Into<String>`. As a matter of fact, they deprecated `into_string`
- `Path` _doesn't_ implement `Display` because it can't faithfully represent a path as a string.
- `IpAddr` doesn't have a special `encode` method either.

raulk:
I think a dedicated `encode` method is probably non-idiomatic. Instead, how do we feel about implementing `ToString` specifically for this type? That way we won't fall back to `Display` through the blanket implementation.

Stebalien:
> That way we won't fall back to Display through the blanket implementation.

Rust won't let you do that (you'd need specialization).

LesnyRumcajs:
@Stebalien @raulk Thanks for the feedback. Perhaps I had the notion of `ToString` from some other technology. If you believe `Display` is fine to use for serialization purposes I can definitely close this PR and use what we have.

As for the `ToString` trait, the docs say [it shouldn't be implemented directly](https://doc.rust-lang.org/std/string/trait.ToString.html).

What do you think?

Stebalien:
`ToString` is automatically implemented for anything implementing the display. The correct way to implement a "to string" conversion is to implement `Display`.

Conclusion: Call `addr.to_string()` and assume that it'll format the address correctly as a string.

PR #355, opened 2022-03-02
prepare to publish crates; v0.1.0.

raulk:
- rename `ipld_{amt,hamt,car,bitfield}` to `fvm_ipld_...`. The `ipld_` namespace is taken by the original crates, and these forks depend on `fvm_shared`, so it's logical to namespace them under `fvm_`.
- add version numbers to all dependencies.
- add missing metadata to Cargo.toml.


ipld/amt/Cargo.toml
```diff
@@ -1,11 +1,11 @@
 [package]
-name = "ipld_amt"
+name = "fvm_ipld_amt"
 description = "Sharded IPLD Array implementation."
-version = "1.0.0"
+version = "0.1.0"
 license = "MIT OR Apache-2.0"
-authors = ["ChainSafe Systems <info@chainsafe.io>"]
+authors = ["Filecoin Core Devs"]
```
Stebalien, line 6: ChainSafe was the original author, so we need to keep then in the author list.

Stebalien (commented):
The dependency renames are reasonable, but we need to preserve author history.

ipld/amt/Cargo.toml
raulk: In my head, Filecoin Core Devs is a wider group that _includes_ ChainSafe Systems, but happy to make it additive.

Stebalien: Let's make it additive. It's good to keep history for stuff like this.

PR #360, opened 2022-03-07
Support nv15

Stebalien:
Rebased for builtin-actors extraction.

Closes https://github.com/filecoin-project/ref-fvm/issues/335.

PR #364, opened 2022-03-07
Tweak self-transfer behaviour to match v15

### LINKED ISSUE #143: make sure the Machine's transfer logic is level with nv15
Concretely: https://github.com/filecoin-project/lotus/pull/7637.

See context at: https://github.com/filecoin-project/fvm/pull/137#discussion_r770105151

arajasek:
Closes #143

Matches behaviour in https://github.com/filecoin-project/lotus/pull/7637.

The motivation for this change is mostly around UX -- it appears weird when people successfully send themselves 2B Filecoin, even though no actual transfer occurs. This change is slightly less efficient than exiting early on self-transfers, but that's not really a case we're looking to optimize.

### DIFF
--- a/fvm/src/machine/default.rs
+++ b/fvm/src/machine/default.rs
@@ -196,10 +196,6 @@ where
     }
 
     fn transfer(&mut self, from: ActorID, to: ActorID, value: &TokenAmount) -> Result<()> {
-        if from == to || value.is_zero() {
-            return Ok(());
-        }
-
         if value.is_negative() {
             return Err(syscall_error!(IllegalArgument;
                 "attempted to transfer negative transfer value {}", value)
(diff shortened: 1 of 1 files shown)


fvm/src/machine/default.rs
```diff
@@ -214,6 +210,15 @@ where
             .context("cannot transfer from non-existent sender")
             .or_error(ErrorNumber::InsufficientFunds)?;
 
+        if from_actor.balance.lt(value) {
```
Stebalien, line 213: Can we not just use `from_actor.balance < value`?

Stebalien (approved):
LGTM. Do we know of any such messages on-chain? I.e., should this be behind a version guard?

NOTE: we can always remove such version guards later, but it's still useful to be able to sync/debug against chain history for now.

jennijuju:
> LGTM. Do we know of any such messages on-chain? I.e., should this be behind a version guard?
> 
> NOTE: we can always remove such version guards later, but it's still useful to be able to sync/debug against chain history for now.

I.e https://m.filscout.com/zh/message/bafy2bzacebr2quxfina5pcy3uudh5lhrbapmbdaxd3l5qfkfhaun6yq44qxmu

raulk (changes requested):
Let's add a version guard. ref-fvm officially supports nv14 and above, and this would break support for nv14. (And we should extract a test vector from that message.)

arajasek:
Yeah, opened this before the discussion that resulted in https://github.com/filecoin-project/ref-fvm/pull/360#discussion_r821077448

arajasek:
Pulling into #360

PR #366, opened 2022-03-08
fix: use a different CBOR library

vmx:
Use a CBOR library that is tailored towards DAG-CBOR.

CIDs don't work as expected with Serde untagged enums, hence add a custom
implementation for the Pointer (de)serialization. That implementation works
similarly to the Serde untagged enum implementation.

First the data is deserialized into a common format that supports all types.
In our case that's `Ipld`, in Serde it's `Content`. Those are then converted
into the actual types the enum is using.

Directly converting from the serialized form into the concrete enum types is
not possible. One would need to try each of the possibilities sequentially
until one of them matches. The deserializer is consumed when deserializing,
hence it can only be used once. This is the reason to have the intermediate
step.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,9 +20,6 @@ default-members = [
     # "examples/*"
 ]
 
-[patch.crates-io]
-cid = { git = "https://github.com/multiformats/rust-cid", branch = "steb/cbor-hack" }
-
 [profile.actor]
 inherits = "release"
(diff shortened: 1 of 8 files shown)


vmx:
This PR depends on https://github.com/filecoin-project/builtin-actors/pull/65.

Stebalien (approved):

PR #382, opened 2022-03-11
fix: remove uses of Code::Identity and remove the multihash feature

Stebalien:
This feature is, to put it simply, very dangerous as it will panic when given more than 64 bytes of input. See https://github.com/multiformats/rust-multihash/issues/194

Given that all uses of the identity hash function should be special cased _anyways_, there's no reason to leave this foot-gun in place.


raulk (approved):
Nice find. Would be nice to add an `IDENTITY_MULTIHASH_CODEC` constant (or similar) instead of using an inline 0.

PR #388, opened 2022-03-11
feat(bitfield): add optimized last

Stebalien:
And refactor "first" to behave the same way.

Fixes the second point of https://github.com/filecoin-project/builtin-actors/issues/55.


Kubuxu (approved):
SGWM

PR #389, opened 2022-03-11
Remove EPOCH_DURATION_SECONDS

arajasek:
I don't think anything should be using this

### DIFF
--- a/shared/src/clock/mod.rs
+++ b/shared/src/clock/mod.rs
@@ -6,9 +6,6 @@ pub use quantize::*;
 
 const _ISO_FORMAT: &str = "%FT%X.%.9F";
 
-/// Duration of each tipset epoch.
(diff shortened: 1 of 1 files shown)


arajasek:
Hmm, there are a few more such constants that I don't think belong there. Hamt bitwidth, clock drift, expected number of blocks per epoch, etc.

I'd like to remove them, but maybe this needs a discussion first. Thoughts?

https://github.com/filecoin-project/ref-fvm/blob/a0942d80a97dc00e6414fc89955e5ffb845dc256/shared/src/lib.rs#L54

Stebalien (changes requested):
These belong in the actors, but the actors currently expect them to live here. They'll need to be moved first.

arajasek:
Good, I agree. I'll move them.

Issue #318, opened 2022-02-11
support execution traces


raulk:
ref-fvm needs to support Lotus-style execution traces. Not only does this facilitate debugging, but it's important for applications like Lily (Sentinel) that monitor the chain. See discussion here: https://github.com/filecoin-project/FIPs/pull/288#discussion_r802974107.

Notes:

- The `ExecutionTrace` is a complex data structure with recursive references: https://github.com/filecoin-project/lotus/blob/9d15a08d27a04fe987e9e4e530baa02dd8ab7024/chain/types/execresult.go#L12
- We'll need to figure out a serialization format for it to cross the FFI barrier. JSON or CBOR seem appropriate.
- The tracing system is rather sophisticated; it has various forms of magic including an optional wildcard "extra" field which different syscalls pack differently.
- The source/caller information is going to be tricky to get, if not impossible. Tracing happens at the kernel layer (Rust), but the source information required is of the actor.

PR #390, opened 2022-03-13
WIP: execution traces

arajasek:
Addresses #318 

The flow as prototyped in Lotus is:

- FVM-shared has a new execution trace object that the executor returns
- FFI serializes this object (CBOR) and returns the bytes to Lotus
- Lotus deserializes this `FvmExecutionTrace` object and puts the appropriate fields into its `ExecutionTrace` object

True completeness would be achieved when `FvmExecutionTrace` and `ExecutionTrace` coalesce into the same type.

### DIFF
--- a/fvm/src/executor/default.rs
+++ b/fvm/src/executor/default.rs
@@ -15,6 +15,7 @@ use num_traits::Zero;
 
 use super::{ApplyFailure, ApplyKind, ApplyRet, Executor};
 use crate::call_manager::{backtrace, CallManager, InvocationResult};
+use crate::executor::ExecutionTrace;
 use crate::gas::{GasCharge, GasOutputs};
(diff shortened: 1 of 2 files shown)


raulk:
- This seems like a good first stub-like direction.
- The machinery to populate these traces will ultimately live in the CallManager, so I expect further complexities to arise as we integrate deeper.
- We need to figure out how to _enable_ execution traces. Are you thinking of adding a boolean in the Machine Config? (Is that the right level of granularity that clients need, or do they to be able to ask for traces only for some messages inside the same execution?)
- Let's put the ExecutionTrace struct + members in its own top-level `trace` module.

arajasek:
Thanks for the feedback! Closing this PR until it's closer to ready.

PR #394, opened 2022-03-16
chore: update readme.

raulk:
- Remove reference to actors, which now live in https://github.com/filecoin-project/builtin-actors.


Stebalien (approved):

Stebalien:
I'd leave the MSRV where it is. That specifies the _minimum_ version.

PR #398, opened 2022-03-21
code and docs cleanup round

raulk:
Minor cleanup round to confuse auditors a bit less.


fvm/src/executor/default.rs
```diff
@@ -186,21 +180,19 @@ where
...
     // TODO: The return type here is very strange because we have three cases:
-    // 1. Continue (return actor ID & gas).
-    // 2. Short-circuit (return ApplyRet).
-    // 3. Fail (return an error).
-    //
-    // We could use custom types, but that would be even more annoying.
+    //  1. Continue (return actor ID & gas).
+    //  2. Short-circuit (return ApplyRet).
+    //  3. Fail (return an error).
+    //  We could use custom types, but that would be even more annoying.
```
Stebalien, line 186: I generally prefer adding line breaks between sections. Otherwise, it's hard to tell sections/paragraphs apart.

fvm/src/executor/default.rs
```diff
@@ -19,14 +19,7 @@ use crate::gas::{GasCharge, GasOutputs};
...
 
-/// The core of the FVM.
-///
-/// ## Generic types
-/// * B => Blockstore.
-/// * E => Externs.
-/// * K => Kernel.
-//
-// If the inner value is `None` it means the machine got poisend and is unusable.
+// If the inner value is `None` it means the machine got poisoned and is unusable.
```
Stebalien, line 22: Hm. We should document this... but the documentation was wrong anyways (I probably copied it from the Machine).

Stebalien (approved):

PR #400, opened 2022-03-22
update Manifest for builtin actor registry

vyzo:
For https://github.com/filecoin-project/builtin-actors/issues/122

### DIFF
--- a/shared/src/actor/builtin.rs
+++ b/shared/src/actor/builtin.rs
@@ -56,6 +56,41 @@ impl Type {
     pub fn is_principal(&self) -> bool {
         self == &Type::Account || self == &Type::Multisig
     }
+
+    /// Returns the canonical (versioned) actor name, as exists in the manifest
+    pub fn name(&self) -> String {
+        match self {
+            Type::System => String::from("fil/v7/system"),
+            Type::Init => String::from("fil/v7/init"),
+            Type::Cron => String::from("fil/v7/cron"),
+            Type::Account => String::from("fil/v7/account"),
+            Type::Power => String::from("fil/v7/storagepower"),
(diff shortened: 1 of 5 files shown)


vyzo:
conformance tests fail because the bundles are wrong.

shared/src/actor/builtin.rs
```diff
@@ -56,6 +56,41 @@ impl Type {
...
+            Type::Account => String::from("fil/v7/account"),
+            Type::Power => String::from("fil/v7/storagepower"),
+            Type::Miner => String::from("fil/v7/storageminer"),
+            Type::Market => String::from("fil/v7/storagemarket"),
+            Type::PaymentChannel => String::from("fil/v7/paymentchannel"),
+            Type::Multisig => String::from("fil/v7/multisig"),
+            Type::Reward => String::from("fil/v7/reward"),
+            Type::VerifiedRegistry => String::from("fil/v7/verifiedregistry"),
+        }
+    }
```
raulk, line 75: Hm, I don't think we want any of this. The manifest in the bundle will contain a `Map<String, Cid>` where String is one of these values: https://github.com/filecoin-project/FIPs/pull/328/files#diff-409cddcb6e5c156f98898f7fe707d292b13d1d25be87bfe2363c39409c131e72R142-R152 (not the entire legacy Code CID, which doesn't make sense to have here).

raulk (changes requested):
Hm, this seems like the wrong direction. Nothing in the inner FVM needs to change, only our parsing of the Manifest which goes from deserializing directly into the type used by the FVM to deserializing into `Map<String, Cid>` and then having to reverse it and convert String to the enum equivalent.

vyzo:
Ok, will try again.

PR #401, opened 2022-03-22
builtin actor registry in system actor

vyzo:
for https://github.com/filecoin-project/builtin-actors/issues/122
companion to https://github.com/filecoin-project/builtin-actors/pull/132

### DIFF
--- a/fvm/src/lib.rs
+++ b/fvm/src/lib.rs
@@ -135,7 +135,7 @@ mod test {
             Zero::zero(),
             fvm_shared::version::NetworkVersion::V14,
             root,
-            Some(manifest_cid),
+            (0, Some(manifest_cid)),
             bs,
             DummyExterns,
         )
--- a/fvm/src/lib.rs
+++ b/fvm/src/lib.rs
@@ -27,6 +27,7 @@ mod init_actor;
 mod market_actor;
 mod power_actor;
 mod reward_actor;
+mod system_actor;
 
 use cid::multihash::{Code, MultihashDigest};
 use cid::Cid;
--- a/fvm/src/lib.rs
+++ b/fvm/src/lib.rs
@@ -135,7 +135,7 @@ mod test {
             Zero::zero(),
             fvm_shared::version::NetworkVersion::V14,
             root,
-            manifest_cid,
+            Some(manifest_cid),
             bs,
             DummyExterns,
         )
--- a/fvm/src/machine/default.rs
+++ b/fvm/src/machine/default.rs
@@ -95,14 +95,7 @@ where
 
         // Load the built-in actors manifest.
         // TODO: Check that the actor bundle is sane for the network version.
-        let maybe_builtin_actors = load_manifest(&blockstore, &builtin_actors);
-        if maybe_builtin_actors.is_err() {
-            return Err(anyhow!(
-                "blockstore doesn't contain builtin actors index with CID {}",
-                &builtin_actors
-            ));
-        }
-        let builtin_actors = maybe_builtin_actors.unwrap();
+        let builtin_actors = load_manifest(&blockstore, &builtin_actors)?;
 
         // Preload any uncached modules.
         // This interface works for now because we know all actor CIDs
--- a/fvm/src/machine/default.rs
+++ b/fvm/src/machine/default.rs
@@ -2,7 +2,7 @@ use std::ops::RangeInclusive;
 
 use anyhow::{anyhow, Context as _};
 use cid::Cid;
-use fvm_shared::actor::builtin::{Manifest, load_manifest};
+use fvm_shared::actor::builtin::{load_manifest, Manifest};
 use fvm_shared::address::Address;
 use fvm_shared::blockstore::{Blockstore, Buffered};
 use fvm_shared::clock::ChainEpoch;
--- a/fvm/src/machine/default.rs
+++ b/fvm/src/machine/default.rs
@@ -57,7 +57,7 @@ where
         circ_supply: TokenAmount,
         network_version: NetworkVersion,
         state_root: Cid,
-        builtin_actors: Option<Cid>,
+        builtin_actors: (u32, Option<Cid>),
         blockstore: B,
         externs: E,
     ) -> anyhow::Result<Self> {
@@ -102,14 +102,15 @@ where
 
         // Load the built-in actors manifest.
         // TODO: Check that the actor bundle is sane for the network version.
-        let builtin_actors_cid = match builtin_actors {
+        let builtin_actors_cid = match builtin_actors.1 {
             Some(cid) => cid,
             None => {
                 let (state, _) = SystemActorState::load(&state_tree)?;
                 state.builtin_actors
             }
         };
-        let builtin_actors = load_manifest(state_tree.store(), &builtin_actors_cid)?;
+        let builtin_actors =
+            load_manifest(state_tree.store(), &builtin_actors_cid, builtin_actors.0)?;
 
         // Preload any uncached modules.
         // This interface works for now because we know all actor CIDs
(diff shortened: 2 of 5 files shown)


fvm/src/machine/default.rs
```diff
@@ -95,15 +95,14 @@ where
...
-        let builtin_actors: Manifest = blockstore
-            .get_cbor(&builtin_actors)
-            .context("failed to load built-in actor index")?
-            .ok_or_else(|| {
-                anyhow!(
-                    "blockstore doesn't contain builtin actors index with CID {}",
-                    &builtin_actors
-                )
-            })?;
+        let maybe_builtin_actors = load_manifest(&blockstore, &builtin_actors);
```
raulk, line 98: `builtin_actors` needs to be turned into an `Option<Cid>`. When `None`, we'll load it from the System Actor's state (after the `state_tree` has been constructed). When `Some`, we'll load it directly from the blockstore as we're doing now.

shared/src/actor/builtin.rs
```diff
@@ -82,5 +84,41 @@ impl TryFrom<&str> for Type {
...
+            Type::Account => String::from("account"),
+            Type::Power => String::from("storagepower"),
+            Type::Miner => String::from("storageminer"),
+            Type::Market => String::from("storagemarket"),
+            Type::PaymentChannel => String::from("paymentchannel"),
+            Type::Multisig => String::from("multisig"),
+            Type::Reward => String::from("reward"),
+            Type::VerifiedRegistry => String::from("verifiedregistry"),
+        }
+    }
```
raulk, line 102: I _think_ we can simplify this:

```suggestion
        match t {
            Type::System => "system",
            Type::Init => "init",
            Type::Cron => "cron",
            Type::Account => "account",
            Type::Power => "storagepower",
            Type::Miner => "storageminer",
            Type::Market => "storagemarket",
            Type::PaymentChannel => "paymentchannel",
            Type::Multisig => "multisig",
            Type::Reward => "reward",
            Type::VerifiedRegistry => "verifiedregistry",
        }.to_string()
    }
```

shared/src/actor/builtin.rs
```diff
@@ -82,5 +84,41 @@ impl TryFrom<&str> for Type {
...
+
+pub fn load_manifest<B: Blockstore>(bs: &B, root_cid: &Cid) -> Result<Manifest, String> {
+    let vec: Vec<(String, Cid)> = match bs.get_cbor(root_cid) {
+        Ok(Some(vec)) => vec,
+        Ok(None) => {
+            return Err("cannot find manifest root cid".to_string());
+        }
+        Err(what) => {
+            return Err(what.to_string());
+        }
```
raulk, line 116: Try using the `?` operator? Then we just need to deal with the Option.

fvm/src/machine/default.rs
raulk: We will need to copy the source for the system actor's state object into ref-fvm, just like we have for the init actor (and other implicitly coupled actors).

shared/src/actor/builtin.rs
vyzo: ah yes, indeed simpler.

shared/src/actor/builtin.rs
vyzo: ok.

vyzo: I anyhowized it, but the second "?" needed a match. anyhow, it looks a little better now.

fvm/src/machine/default.rs
vyzo: done.

shared/src/actor/builtin.rs
vyzo: done.

vyzo:
I think we have a problem with the conformance test machine instantiation, we pass it Some(cid) and it will try to load the manifest with the new code.... which will not end up good with the given bundles.

Or will it be ok because of the synthetic code cids being identity hashes?

vyzo:
so yeah, problem with the manifest data structure -- test bundles expect the old one and fail to load.

vyzo:
I versioned the manifest, passing a tuple `(version,` Option<Cid>)` to the machine constructor.

Stebalien:
> I versioned the manifest, passing a tuple (version,Option<Cid>) to the machine constructor.

I'd rather not externally version the manifest if possible. Can we not just backport the change to builtin-actors v7 (and drop v6 support)?

raulk:
@vyzo I'd rather monkey-patch the old manifest format _only inside_ the conformance test suite, than alter the format of the manifest itself for this n=1 case.

vyzo:
its not just the confromance test suite, regular tests fail to load it too.

fvm/src/machine/default.rs
```diff
@@ -56,7 +57,7 @@ where
         circ_supply: TokenAmount,
         network_version: NetworkVersion,
         state_root: Cid,
-        builtin_actors: Cid,
+        builtin_actors: (u32, Option<Cid>),
```
raulk, line 60: Just to confirm, passing this version number here is temporary to avoid breaking the tests? A follow-up PR will entirely remove the version number here? (Note that this affects the FFI layer too)

vyzo: for now yes, and we can certainly remove it once we have proper v1 bundles.

We might want to keep it for future proofing, but that is not rewuired given that there is no (and wont be for quite a while) version other than v1.

raulk: The reason I ask is that the version _really_ doesn't go here, at all.

- We only need it now to deal with the temporary situation of the schema change during development.
- In the future, if there are any System Actor schema changes, they would go coupled with a new network version.
- The builtin_actors CId argument is only here to support the canary mode (nv14 and nv15, where the system actor doesn't hold the manifest). It might eventually go as well.

Can you submit the follow-up PR to erase this version? We can make a release with only this PR, then release builtin-actors, then merge the new PR.

vyzo: one intermediate step is making the bundles, but yes of course i will make that pr.

vyzo: agreed in either case

vyzo: follow up in #404

fvm/src/lib.rs
```diff
@@ -134,7 +135,7 @@ mod test {
             Zero::zero(),
             fvm_shared::version::NetworkVersion::V14,
             root,
-            manifest_cid,
+            (0, Some(manifest_cid)),
```
raulk, line 138: Temporary so that we can make a release to integrate in builtin-actors; removed by #404.

shared/src/actor/builtin.rs
```diff
@@ -82,5 +85,61 @@ impl TryFrom<&str> for Type {
...
+        _ => Err(anyhow!("unknown manifest version {}", ver)),
+    }
+}
+
+pub fn load_manifest_v0<B: Blockstore>(bs: &B, root_cid: &Cid) -> anyhow::Result<Manifest> {
+    match bs.get_cbor::<Manifest>(root_cid)? {
+        Some(mf) => Ok(mf),
+        None => Err(anyhow!("cannot find manifest root cid {}", root_cid)),
+    }
+}
```
raulk, line 123: Temporary, needed to avoid build breakage across modules. Removed by #404.

raulk (approved):

Stebalien:
This change has caused quite a bit of pain, so I'm going to post a small personal retro:

- We should avoid merging things on Friday right before you go to bed.
- We should be extra careful about releasing breaking changes in patch releases (e.g., I just did this today with an accidental bundler release...)
- In general, avoid making "temporary" breaking changes unless we're _REALLY_ sure we need them.

With 20/20 hindsight, we could have just:

1. Made the changes to `fvm_shared`, and released them. We _didn't_ have to merge the rest of the changes.
2. Made the changes to the actors bundle.
3. Finally, consumed those changes here.

There probably would have been fallout, but we could have avoided a broken RC and a lot of angst.

Stebalien:
**NOTE**: to be clear, it wasn't _just_ this change and the pain wasn't really anyone's fault. It's just that this lead to a comedy of errors so I wanted to try to find ways in which we could have avoided these issues.

PR #404, opened 2022-03-24
builtin actor registry followup

vyzo:
Follow up to #401 
Blocked on getting v1 bundles for conformance tests.


Stebalien:
This won't work.

Stebalien:
New version in https://github.com/filecoin-project/ref-fvm/pull/416.

PR #405, opened 2022-03-24
detailed gas and fuel tracing.

raulk:
Introduces a new Cargo feature `tracing` that enables detailed gas tracing.

Traces are accumulated in the GasTracer, owned by the CallManager.

Gas traces specify the context, point of tracing, consumption stats, and timing stats.

We currently support these tracing points:

- {Pre,Post}Syscall
- {Pre,Post}Extern
- {Enter,Exit}Call
- Start (of the call stack)
- Finish (of the call stack)

Traces are currently serialized to JSON and printed directly on stdout.

There's a bit of duplication the syscall binding macros that I'm not happy about, and perhaps we could make the tracepoints in the CallManager a bit DRYer.

---

Example of a gas trace that this produces for a test vector: https://gist.github.com/raulk/f2207675000571e283bc2e20e43a5463


raulk:
Superseded by https://github.com/filecoin-project/ref-fvm/pull/465.

PR #409, opened 2022-03-25
bitfield improvements

Stebalien:


Kubuxu (approved):

Kubuxu (approved):

Stebalien:
Per @anorth's review, I've changed `set` to _panic_ and introduced a `try_set`. We treat these like "index out of bounds" errors.

The main downside is that the behavior is less self-documenting, but we provide a `try_set` to help with that.

Kubuxu (approved):

PR #411, opened 2022-03-26
Add clusterfuzzlite integration

Kubuxu:
It will fuzz on PR (if relevant code is changed), batch fuzz in the background,
produce fuzzing coverage reports and prune the corpus.

The aim of ClusterfuzzLite is to detect newly introduced regressions caused by code changes.

### DIFF
--- a/.clusterfuzzlite/build.sh
+++ b/.clusterfuzzlite/build.sh
@@ -1,0 +1,45 @@
+#!/bin/bash -eu
+# Copyright 2021 Google LLC.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#      http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+################################################################################
+
+cd "$SRC"
+
+declare -a PROJECTS=(amt hamt)
+declare -A PROJECT_PATHS=(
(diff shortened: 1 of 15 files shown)


.clusterfuzzlite/build.sh
```diff
@@ -0,0 +1,45 @@
...
+)
+
+export CARGO_TARGET_DIR="$SRC/target"
+FUZZ_TARGET_OUTPUT_DIR="$CARGO_TARGET_DIR/x86_64-unknown-linux-gnu/release"
+
+
+for project in "${PROJECTS[@]}"; do
+	pushd "${PROJECT_PATHS[$project]}"
+
+	cargo +nightly fuzz build -O --debug-assertions
```
raulk, line 33: Shouldn't we be using the same version as the tool chain?

.github/workflows/cflite_batch.yml
```diff
@@ -0,0 +1,33 @@
...
+    - cron: '0 0/12 * * *'  # Every 12th hour. Change this to whatever is suitable.
+permissions: read-all
+jobs:
+  BatchFuzzing:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        sanitizer:
+        - address
```
raulk, line 13: This looks odd with no indentation.

raulk (approved):
I believe you know what you're doing, but it would be great to document what's happening, where, and when on the README as part of this PR.

.clusterfuzzlite/build.sh
Kubuxu: Unfortunately, we can't. `cargo fuzz` uses some options which are only available in unstable builds. I think it will change once 1.59.0 is adopted by cargo-fuzz but I might be wrong (and we are on 1.58 either way).

Kubuxu:
I've added a README, I will merge it now so tomorrow I know if cron tasks work. Worst case we can revert.

Kubuxu:
@raulk now thinking about it the README should go into `.clusterfuzzlite` I will fix it in future PR. I blame Sunday brain on it.

PR #412, opened 2022-03-27
Implement execution traces for the FVM

### LINKED ISSUE #318: support execution traces

arajasek:
Resolves #318 

This gets us most of the way to matching Lotus execution traces -- fine gas tracing isn't implemented yet, but it will be impossible to meet that level of detail in M1 anyway.

The ExecutionTrace is created and populated by the call manager, which returns it on `finish`.

### DIFF
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -11,10 +11,12 @@ use num_traits::Zero;
 
 use super::{Backtrace, CallManager, InvocationResult, NO_DATA_BLOCK_ID};
 use crate::call_manager::backtrace::Frame;
+use crate::call_manager::FinishRet;
 use crate::gas::GasTracker;
 use crate::kernel::{ClassifyResult, ExecutionError, Kernel, Result};
 use crate::machine::Machine;
 use crate::syscalls::error::Abort;
+use crate::trace::{ExecutionEvent, ExecutionTrace, SendParams};
 use crate::{account_actor, syscall_error};
 
 /// The default [`CallManager`] implementation.
@@ -40,6 +42,8 @@ pub struct InnerDefaultCallManager<M> {
     call_stack_depth: u32,
     /// The current chain of errors, if any.
     backtrace: Backtrace,
+    /// The current execution trace.
+    exec_trace: ExecutionTrace,
 }
 
 #[doc(hidden)]
@@ -73,6 +77,7 @@ where
             num_actors_created: 0,
             call_stack_depth: 0,
             backtrace: Backtrace::default(),
+            exec_trace: vec![],
         })))
     }
 
@@ -87,6 +92,16 @@ where
     where
         K: Kernel<CallManager = Self>,
     {
+        if self.machine.context().tracing {
+            self.exec_trace.push(ExecutionEvent::Call(SendParams {
+                from,
+                to,
+                method,
+                params: params.clone(),
+                value: value.clone(),
+            }));
+        }
+
         // We check _then_ set because we don't count the top call. This effectivly allows a
         // call-stack depth of `max_call_depth + 1` (or `max_call_depth` sub-calls). While this is
         // likely a bug, this is how NV15 behaves so we mimic that behavior here.
@@ -108,6 +123,11 @@ where
         self.call_stack_depth += 1;
         let result = self.send_unchecked::<K>(from, to, method, params, value);
         self.call_stack_depth -= 1;
+
+        if self.machine.context().tracing {
+            self.exec_trace.push(ExecutionEvent::Return(result.clone()));
+        }
+
         result
     }
 
@@ -124,12 +144,19 @@ where
         res
     }
 
-    fn finish(mut self) -> (i64, Backtrace, Self::Machine) {
+    fn finish(mut self) -> (FinishRet, Self::Machine) {
         let gas_used = self.gas_tracker.gas_used().max(0);
 
         let inner = self.0.take().expect("call manager is poisoned");
         // TODO: Having to check against zero here is fishy, but this is what lotus does.
-        (gas_used, inner.backtrace, inner.machine)
+        (
+            FinishRet {
+                gas_used,
+                backtrace: inner.backtrace,
+                exec_trace: inner.exec_trace,
+            },
+            inner.machine,
+        )
     }
 
     // Accessor methods so the trait can implement some common methods by default.
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -4,7 +4,7 @@ use fvm_ipld_encoding::{RawBytes, DAG_CBOR};
 use fvm_shared::actor::builtin::Type;
 use fvm_shared::address::{Address, Protocol};
 use fvm_shared::econ::TokenAmount;
-use fvm_shared::error::ExitCode;
+use fvm_shared::error::{ErrorNumber, ExitCode};
 use fvm_shared::version::NetworkVersion;
 use fvm_shared::{ActorID, MethodNum, METHOD_SEND};
 use num_traits::Zero;
@@ -13,7 +13,7 @@ use super::{Backtrace, CallManager, InvocationResult, NO_DATA_BLOCK_ID};
 use crate::call_manager::backtrace::Frame;
 use crate::call_manager::FinishRet;
 use crate::gas::GasTracker;
-use crate::kernel::{ClassifyResult, ExecutionError, Kernel, Result};
+use crate::kernel::{ClassifyResult, ExecutionError, Kernel, Result, SyscallError};
 use crate::machine::Machine;
 use crate::syscalls::error::Abort;
 use crate::trace::{ExecutionEvent, ExecutionTrace, SendParams};
@@ -125,7 +125,16 @@ where
         self.call_stack_depth -= 1;
 
         if self.machine.context().tracing {
-            self.exec_trace.push(ExecutionEvent::Return(result.clone()));
+            self.exec_trace.push(ExecutionEvent::Return(match result {
+                Err(ref e) => Err(match e {
+                    ExecutionError::OutOfGas => {
+                        SyscallError::new(ErrorNumber::Forbidden, "out of gas")
+                    }
+                    ExecutionError::Fatal(_) => SyscallError::new(ErrorNumber::Forbidden, "fatal"),
+                    ExecutionError::Syscall(s) => s.clone(),
+                }),
+                Ok(ref v) => Ok(v.clone()),
+            }));
         }
 
         result
--- a/fvm/src/call_manager/mod.rs
+++ b/fvm/src/call_manager/mod.rs
@@ -11,11 +11,14 @@ use crate::state_tree::StateTree;
 use crate::Kernel;
 
 pub mod backtrace;
+
 pub use backtrace::Backtrace;
 
 mod default;
 
 pub use default::DefaultCallManager;
+
+use crate::trace::ExecutionTrace;
 
 /// BlockID representing nil parameters or return data.
 pub const NO_DATA_BLOCK_ID: u32 = 0;
@@ -59,8 +62,8 @@ pub trait CallManager: 'static {
         f: impl FnOnce(&mut Self) -> Result<InvocationResult>,
     ) -> Result<InvocationResult>;
 
-    /// Finishes execution, returning the gas used and the machine.
-    fn finish(self) -> (i64, backtrace::Backtrace, Self::Machine);
+    /// Finishes execution, returning the gas used, machine, and exec trace if requested.
+    fn finish(self) -> (FinishRet, Self::Machine);
 
     /// Returns a reference to the machine.
     fn machine(&self) -> &Self::Machine;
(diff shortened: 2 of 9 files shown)


fvm/src/call_manager/default.rs
```diff
@@ -86,14 +92,59 @@ where
...
             return Err(
                 syscall_error!(LimitExceeded, "message execution exceeds call depth").into(),
             );
         }
         self.call_stack_depth += 1;
         let result = self.send_unchecked::<K>(from, to, method, params, value);
         self.call_stack_depth -= 1;
+
+        trace.msg_receipt = match result {
+            Ok(ref i) => Receipt {
```
Stebalien, line 117:
```suggestion
        trace.msg_receipt = match &result {
            Ok(i) => Receipt {
```

fvm/src/call_manager/default.rs
```diff
@@ -86,14 +92,59 @@ where
...
+                ExecutionError::OutOfGas => Receipt {
+                    exit_code: ExitCode::SysErrOutOfGas,
+                    return_data: Default::default(),
+                    gas_used: 0,
+                },
+                ExecutionError::Syscall(s) => Receipt {
+                    exit_code: match s.1.try_into() {
+                        Ok(e) => e,
+                        _ => ErrPlaceholder,
+                    },
```
Stebalien, line 132: Hm. Any chance we could change the format? We're intentionally distinguishing between syscall errors and exit codes.

fvm/src/call_manager/default.rs
```diff
@@ -86,14 +92,59 @@ where
...
+                },
+                _ => Default::default(),
+            },
+        };
+
+        if self.machine.config().debug {
+            if self.exec_trace.is_none() {
+                self.exec_trace = Some(trace);
+            } else {
+                self.exec_trace.as_mut().unwrap().subcalls.push(trace);
```
Stebalien, line 144: won't this put all sub-calls in the same level?

fvm/src/call_manager/default.rs
```diff
@@ -86,14 +92,59 @@ where
     where
         K: Kernel<CallManager = Self>,
     {
+        let mut trace = ExecutionTrace::new(
```
Stebalien, line 95: ideally we'd avoid doing any work unless debugging is enabled.

fvm/src/call_manager/default.rs
Stebalien: Basically, receipts are really for recording results on-chain, but aren't the best for debugging. In this case, it's also causing us to drop the error message.

fvm/src/call_manager/default.rs
```diff
@@ -39,6 +42,8 @@ pub struct InnerDefaultCallManager<M> {
     call_stack_depth: u32,
     /// The current chain of errors, if any.
     backtrace: Backtrace,
+    /// The current execution trace.
+    exec_trace: Option<ExecutionTrace>,
```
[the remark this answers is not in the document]
arajasek: Sg, thanks!

shared/src/receipt.rs
```diff
@@ -13,4 +13,14 @@ pub struct Receipt {
     pub gas_used: i64,
 }
 
+impl Default for Receipt {
```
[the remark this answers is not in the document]
arajasek: Not without deriving Default for ExitCode which I didn't want to do

fvm/src/call_manager/default.rs
```diff
@@ -110,12 +138,19 @@ where
         res
     }
 
-    fn finish(mut self) -> (i64, Backtrace, Self::Machine) {
+    fn finish(mut self) -> (FinishResult, Self::Machine) {
```
raulk, line 141: nit: `FinishResult` sounds like a specialized `Result` type. Can we change this type name to `FinishRet`, parallel with `ApplyRet`?

fvm/src/call_manager/mod.rs
```diff
@@ -140,3 +146,9 @@ impl InvocationResult {
         }
     }
 }
+
+pub struct FinishResult {
```
raulk, line 150: nit: docs.

raulk, line 150: Maybe just make the type `pub(crate)` instead of its fields? It's a little bit useless to have a public method-less struct with all `pub(crate)` fields.

fvm/src/executor/default.rs
```diff
@@ -149,12 +153,15 @@ where
         };
 
         match apply_kind {
-            ApplyKind::Explicit => self.finish_message(msg, receipt, failure_info, gas_cost),
+            ApplyKind::Explicit => {
+                self.finish_message(msg, receipt, failure_info, gas_cost, exec_trace)
```
raulk, line 157: Hm, this was confusing to me. I didn't expect `finish_message` to operate on the `exec_trace`, so I didn't understand why we passed it in. And indeed it doesn't.

Did you consider:

```rust
.map(|apply_ret| {
    apply_ret.exec_trace = exec_trace;
    apply_ret
})
```

It's a few more lines of code but it's less surprising (for me). Alternatively we could use a one-liner:

```rust
.inspect(|apply_ret| { apply_ret.exec_trace = exec_trace })
```

But it requires us to enable an experimental API: https://github.com/rust-lang/rust/issues/91345.

fvm/src/kernel/error.rs
```diff
@@ -26,7 +28,7 @@ macro_rules! syscall_error {
     };
 }
 
-// NOTE: this intentionally does not implemnent error so we can make the context impl work out
+// NOTE: this intentionally does not implement error so we can make the context impl work out
 // below.
 #[derive(Display, Debug)]
```
raulk, line 33: Is it not possible to derive `Clone`?

fvm/src/trace/mod.rs
```diff
@@ -0,0 +1,26 @@
...
+use fvm_shared::{ActorID, MethodNum};
+
+use crate::call_manager::InvocationResult;
+use crate::kernel::Result;
+
+/// Execution Trace, only for informational and debugging purposes.
+pub type ExecutionTrace = Vec<ExecutionEvent>;
+
+#[derive(Clone, Debug)]
+pub enum ExecutionEvent {
```
raulk, line 13: Does this not need to derive serialize so that can filecoin-ffi can CBOR-serialize it over the FFI boundary? (Same for SendParams).

fvm/src/call_manager/default.rs
```diff
@@ -86,6 +91,21 @@ where
...
+                to,
+                method,
+                params: params.clone(),
+                value: value.clone(),
+            });
+            if self.exec_trace.is_none() {
+                self.exec_trace = Some(vec![call_event]);
+            } else {
+                self.exec_trace.as_mut().unwrap().push(call_event);
+            }
```
dignifiedquire, line 106: should be written as 

```rust
if let Some(ref mut exec_trace) = self.exec_trace {
  exec_trace.push(call_event);
} else {
  self.exec_trace = Some(vec![call_event]);
}
```

fvm/src/call_manager/default.rs
```diff
@@ -94,6 +114,14 @@ where
         self.call_stack_depth += 1;
         let result = self.send_unchecked::<K>(from, to, method, params, value);
         self.call_stack_depth -= 1;
+
+        if self.machine.config().debug {
+            self.exec_trace
+                .as_mut()
+                .unwrap()
```
dignifiedquire, line 121: this should be at least an `.expect` explaining why this is already checked

fvm/src/call_manager/default.rs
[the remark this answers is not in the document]
dignifiedquire: 👌

fvm/src/kernel/error.rs
arajasek: No, because anyhow::Error doesn't feel inspired to be cloned.

fvm/src/call_manager/default.rs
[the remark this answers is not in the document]
arajasek: Y'all and this entire language are outta control. But that's cool to see.

fvm/src/executor/default.rs
```diff
@@ -79,8 +79,11 @@ where
 
                 Ok(ret)
             });
-            let (gas_used, backtrace, machine) = cm.finish();
-            (Ok((result, gas_used, backtrace)), machine)
+            let (res, machine) = cm.finish();
+            (
+                Ok((result, res.gas_used, res.backtrace, res.exec_trace)),
```
Stebalien, line 84:
```suggestion
                Ok(((result, res), machine)),
```

?

fvm/src/kernel/error.rs
```diff
@@ -35,6 +37,16 @@ pub enum ExecutionError {
...
 
+impl Clone for ExecutionError {
+    fn clone(&self) -> Self {
+        match self {
+            OutOfGas => OutOfGas,
+            Syscall(v) => Syscall(v.clone()),
+            Fatal(v) => Fatal(anyhow::Error::msg(v.to_string())),
+        }
+    }
+}
```
Stebalien, line 48: Hm... this isn't really a clone. I'd handle these cases separately:

1. fatal/out-of-gas: nothing? We can just cut the trace short because it doesn't really matter.
2. Syscall error -> push that.

fvm/src/machine/mod.rs
```diff
@@ -178,18 +179,26 @@ pub struct MachineContext {
     ///
     /// DEFAULT: Total FIL supply (likely not what you want).
     pub circ_supply: TokenAmount,
+
+    pub tracing: bool,
```
Stebalien, line 183: docs

fvm/src/machine/mod.rs
```diff
@@ -178,18 +179,26 @@ pub struct MachineContext {
...
 
-    /// Set [`EpochContext::circ_supply`].
+    /// Set [`MachineContext::circ_supply`].
     pub fn set_circulating_supply(&mut self, amt: TokenAmount) -> &mut Self {
         self.circ_supply = amt;
         self
     }
+
+    /// Set [`MachineContext::tracing`].
+    pub fn enable_tracing(&mut self) -> &mut Self {
```
Stebalien, line 200: `enable_tracing(&mut self, enable: bool)` to match `enable_debug`. Or change `enable_debug`?

Stebalien (commented):
Mostly LGTM with some comments.

fvm/src/kernel/error.rs
Stebalien: Basically, record `Result<InvocationResult, SyscallError>` instead of `Result<InvocationResult, ExecutionError>`.

fvm/src/machine/mod.rs
arajasek: Changing enable_debug

fvm/src/executor/default.rs
arajasek: Eh, but we use `gas_used` so many times, and they're not suuuper related concepts, and I don't like it.

fvm/src/call_manager/default.rs
```diff
@@ -125,7 +125,16 @@ where
         self.call_stack_depth -= 1;
 
         if self.machine.context().tracing {
-            self.exec_trace.push(ExecutionEvent::Return(result.clone()));
+            self.exec_trace.push(ExecutionEvent::Return(match result {
+                Err(ref e) => Err(match e {
```
arajasek, line 129: This combination of `ref`s and `clone`s appeased the checker, but I feel like there's something better that can be done?

fvm/src/executor/default.rs
Stebalien: We can destructure this on the outside. I'm just suggesting a way to slightly reduce the amount of code.

Stebalien: (but it doesn't really matter)

fvm/src/call_manager/default.rs
Stebalien: Not really...

Stebalien (approved):
LGTM.

fvm/src/call_manager/default.rs
Stebalien: Well, you can use `&result` and remove the `ref e` and friends, but you still need to clone. On the other hand, this will only happen when tracing is enabled.

raulk (approved):
LGTM.

PR #415, opened 2022-03-28
Improve AMT equivalance fuzzer

Kubuxu:
Remove simple AMT fuzzing as the equivalence fuzzer now does its work better.


Stebalien (approved):
LGTM. Conformance tests should be fixed by https://github.com/filecoin-project/ref-fvm/pull/416.

PR #416, opened 2022-03-29
feat: switch to the new versioned manifest.

Stebalien:
1. If no manifest CID is passed, load the manifest from the system actor and assume "v1" for now. In the future, we'll likely use the network version to infer the manifest version.
2. If a manifest CID _is_ passed, assume it's a versioned manifest and use the specified version.

replaces #404

### DIFF
--- a/fvm/src/lib.rs
+++ b/fvm/src/lib.rs
@@ -127,6 +127,8 @@ mod test {
             bs.put_cbor(&manifest, Code::Blake2b256).unwrap()
         };
 
+        let actors_cid = bs.put_cbor(&(0, manifest_cid), Code::Blake2b256).unwrap();
+
         let machine = DefaultMachine::new(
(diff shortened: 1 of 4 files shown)


arajasek (approved):

fvm/src/lib.rs
```diff
@@ -127,6 +127,8 @@ mod test {
             bs.put_cbor(&manifest, Code::Blake2b256).unwrap()
         };
 
+        let actors_cid = bs.put_cbor(&(0, manifest_cid), Code::Blake2b256).unwrap();
```
vyzo, line 130: wherent we supposed to eliminate v0? The original pr fixed this to use the v1 structure.

Stebalien: This was just a test but I'm happy to get rid of it.

PR #417, opened 2022-03-29
fix: don't leak underlying library error

vmx:
The error of the underlying CBOR library is an implementation detail,
it shouldn't leak into the public API. Instead use the error that
wraps it into the one specified by the module.


Stebalien (approved):
Looking at the builtin actors, this should be fine. The builtin actors will return an `ErrSerialization` exit code for _either_ error.

Stebalien (commented):
Well, actually, this _will_ break things. But that's not the end of the world. We just need a patch release.

Issue #253, opened 2022-01-12
Reject actor abort with a system exit code


Stebalien:
Unfortunately, we need to disable this for now to fix #245 and #243 (and others).

PR #431, opened 2022-03-31
Implement ExitCode as u32 wrapper, and use new spec codes

### LINKED ISSUE #430: VM rejects any exit code it doesn't have hard-coded in enum
When receiving an exit code from WASM during an internal send, the VM attempts to interpret it as one of the ExitCode enum values, but overwrites it with  `SysErrIllegalActor` if it can't. The VM should accept any u32 exit code here. An enum is generally not an appropriate representation.

In send.rs:
```
        let exit_code = ExitCode::from_u32(exit_code).unwrap_or(ExitCode::SysErrIllegalActor);
```

anorth:
This is for discussion first, following initial discussion on Slack. I've implemented ExitCode as a wrapper over a u32, not an enum. It thus accepts any value. I've then added constants for the system and standard values, updating them to match the FVM [spec](https://github.com/filecoin-project/fvm-specs/blob/main/07-errors.md).

Possible changes:
- Some way to use ExitCode values as values of an enum, rather than as a bag of constants.
- Rename back to CamelCase and explicitly ignore Clippy

The worst bit of this change, IMO, is renaming to ALL_CAPS to follow Rust convention for constants that are not enum values. I did try leaving the SystemExitCode values as an enum of integers, with a `From` implementation, but that results in a bunch of `.into()` wherever they are used. I opted to use constant values of `ExitCode` type instead.

I considered using an enum as a discriminated union, with cases for ok, system code, standard code, and others, each containing a u32 value with explicit range checking on creation. I think that would be trying much to hard. There's no call to `match` over those, and not much call here for leaning on type safety for range bounds. In the built-in actors, we might not even use the ExitCode type. An enum of integers on that side could work well.

**Questions**

~The spec removed SysErrIllegalActor, but I've restored it as the code to which the syscall error number `IllegalOperation` maps. Please let me know if that should instead map to `SysErrIllegalInstruction` (i.e. panic) or something else. I didn't yet trace all the places that could generate it.~

The only use of the (non-system) StandardExitCode values in this repo is the `params_cbor` function, which is actually unused here or in builtin-actors, so could be removed (it's just sugar). We could then remove the StandardExitCode values entirely from this repo over to built-in actors, as part of the built-in actor calling convention. WDYT? They might still be useful here for user actors who call into the builtin ones (depending how we export their APIs).

I shortened the names, since they're always qualified. Happy to restore the prefixes on request.

Includes resolution of #253 if we update the conformance tests.
Fixes #430 plus some other TODOs

FYI @ZenGround0

### DIFF
--- a/fvm/src/executor/default.rs
+++ b/fvm/src/executor/default.rs
@@ -7,7 +7,7 @@ use fvm_shared::actor::builtin::Type;
 use fvm_shared::address::Address;
 use fvm_shared::bigint::{BigInt, Sign};
 use fvm_shared::econ::TokenAmount;
-use fvm_shared::error::{ErrorNumber, ExitCode};
+use fvm_shared::error::{ErrorNumber, ExitCode, SystemExitCode};
 use fvm_shared::message::Message;
 use fvm_shared::receipt::Receipt;
 use fvm_shared::ActorID;
@@ -88,7 +88,7 @@ where
             Ok(InvocationResult::Return(return_data)) => {
                 backtrace.clear();
                 Receipt {
-                    exit_code: ExitCode::Ok,
+                    exit_code: ExitCode::OK,
                     return_data,
                     gas_used,
                 }
@@ -104,16 +104,16 @@ where
                 }
             }
             Err(ExecutionError::OutOfGas) => Receipt {
-                exit_code: ExitCode::SysErrOutOfGas,
+                exit_code: SystemExitCode::OUT_OF_GAS,
                 return_data: Default::default(),
                 gas_used,
             },
             Err(ExecutionError::Syscall(err)) => {
                 let exit_code = match err.1 {
-                    ErrorNumber::IllegalOperation => ExitCode::SysErrIllegalActor,
-                    ErrorNumber::AssertionFailed => ExitCode::SysErrIllegalArgument,
-                    ErrorNumber::InsufficientFunds => ExitCode::SysErrInsufficientFunds,
-                    ErrorNumber::NotFound => ExitCode::SysErrInvalidReceiver,
+                    ErrorNumber::IllegalOperation => SystemExitCode::ILLEGAL_ACTOR,
+                    ErrorNumber::AssertionFailed => SystemExitCode::ASSERTION_FAILED,
+                    ErrorNumber::InsufficientFunds => SystemExitCode::INSUFFICIENT_FUNDS,
+                    ErrorNumber::NotFound => SystemExitCode::INVALID_RECEIVER,
                     code => {
                         return Err(anyhow!(
                             "unexpected syscall error when processing message: {} ({})",
@@ -207,7 +207,7 @@ where
                 // Verify the cost of the message is not over the message gas limit.
                 if inclusion_total > msg.gas_limit {
                     return Ok(Err(ApplyRet::prevalidation_fail(
-                        ExitCode::SysErrOutOfGas,
+                        SystemExitCode::OUT_OF_GAS,
                         format!("Out of gas ({} > {})", inclusion_total, msg.gas_limit),
                         &self.context().base_fee * inclusion_total,
                     )));
@@ -227,7 +227,7 @@ where
             Some(id) => id,
             None => {
                 return Ok(Err(ApplyRet::prevalidation_fail(
-                    ExitCode::SysErrSenderInvalid,
+                    SystemExitCode::SENDER_INVALID,
                     "Sender invalid",
                     miner_penalty_amount,
                 )))
@@ -246,7 +246,7 @@ where
             Some(act) => act,
             None => {
                 return Ok(Err(ApplyRet::prevalidation_fail(
-                    ExitCode::SysErrSenderInvalid,
+                    SystemExitCode::SENDER_INVALID,
                     "Sender invalid",
                     miner_penalty_amount,
                 )))
@@ -262,7 +262,7 @@ where
 
         if !sender_is_account {
             return Ok(Err(ApplyRet::prevalidation_fail(
-                ExitCode::SysErrSenderInvalid,
+                SystemExitCode::SENDER_INVALID,
                 "Send not from account actor",
                 miner_penalty_amount,
             )));
@@ -271,7 +271,7 @@ where
         // Check sequence is correct
         if msg.sequence != sender.sequence {
             return Ok(Err(ApplyRet::prevalidation_fail(
-                ExitCode::SysErrSenderStateInvalid,
+                SystemExitCode::SENDER_STATE_INVALID,
                 format!(
                     "Actor sequence invalid: {} != {}",
                     msg.sequence, sender.sequence
@@ -284,7 +284,7 @@ where
         let gas_cost: TokenAmount = msg.gas_fee_cap.clone() * msg.gas_limit;
         if sender.balance < gas_cost {
             return Ok(Err(ApplyRet::prevalidation_fail(
-                ExitCode::SysErrSenderStateInvalid,
+                SystemExitCode::SENDER_STATE_INVALID,
                 format!(
                     "Actor balance less than needed: {} < {}",
                     sender.balance, gas_cost
--- a/fvm/src/executor/default.rs
+++ b/fvm/src/executor/default.rs
@@ -7,7 +7,7 @@ use fvm_shared::actor::builtin::Type;
 use fvm_shared::address::Address;
 use fvm_shared::bigint::{BigInt, Sign};
 use fvm_shared::econ::TokenAmount;
-use fvm_shared::error::{ErrorNumber, ExitCode, SystemExitCode};
+use fvm_shared::error::{ErrorNumber, ExitCode};
 use fvm_shared::message::Message;
 use fvm_shared::receipt::Receipt;
 use fvm_shared::ActorID;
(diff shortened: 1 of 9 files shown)


fvm/src/executor/default.rs
```diff
@@ -104,16 +104,16 @@ where
...
                 return_data: Default::default(),
                 gas_used,
             },
             Err(ExecutionError::Syscall(err)) => {
                 let exit_code = match err.1 {
-                    ErrorNumber::IllegalOperation => ExitCode::SysErrIllegalActor,
-                    ErrorNumber::AssertionFailed => ExitCode::SysErrIllegalArgument,
-                    ErrorNumber::InsufficientFunds => ExitCode::SysErrInsufficientFunds,
-                    ErrorNumber::NotFound => ExitCode::SysErrInvalidReceiver,
+                    ErrorNumber::IllegalOperation => SystemExitCode::ILLEGAL_ACTOR,
```
anorth, line 113: Here is where I need ILLEGAL_ACTOR (dropped by the spec), or an alternative.

shared/src/error/mod.rs
```diff
@@ -1,101 +1,133 @@
...
+}
 
-    /// Indicates a message execution is forbidden for the caller.
-    SysErrForbidden = 8,
+impl<'de> serde::Deserialize<'de> for ExitCode {
+    fn deserialize<D>(_deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        Ok(ExitCode { value: 0 }) // FIXME figure this out
```
anorth, line 60: I still need to implement this.

sdk/src/message.rs
```diff
@@ -78,7 +78,7 @@ pub fn params_cbor<T: Cbor>(id: BlockId) -> SyscallResult<T> {
     match fvm_shared::encoding::from_slice(raw.as_slice()) {
         Ok(v) => Ok(v),
         Err(e) => vm::abort(
-            ExitCode::ErrSerialization as u32,
+            StandardExitCode::SERIALIZATION.value(),
```
anorth, line 81: This is the only use of StandardExitCode

anorth:
The conformance tests fail because this changes behaviour, fixing #253. In the conformance tests, actors exit with code 8 (was `SysErrForbidden`, now `ILLEGAL_ACTOR`), but that's now trapped and converted to `ILLEGAL_EXIT_CODE` (code 9).

I can "fix" this by flipping the codes for ILLEGAL_ACTOR and ILLEGAL_EXIT_CODE so that the bad exit code signal is 8, but that's not the "right" fix. Alternatively, I can disable the lines that fix #253.

fvm/src/executor/default.rs
```diff
@@ -104,16 +104,16 @@ where
...
+                    ErrorNumber::AssertionFailed => SystemExitCode::ASSERTION_FAILED,
+                    ErrorNumber::InsufficientFunds => SystemExitCode::INSUFFICIENT_FUNDS,
+                    ErrorNumber::NotFound => SystemExitCode::INVALID_RECEIVER,
                     code => {
                         return Err(anyhow!(
```
Stebalien, line 118: Future comment: We should probably change this to return `SysErrAssertionFailed` rather than a fatal error.

fvm/src/executor/default.rs
Stebalien: Send _shouldn't_ be able to return this, I think. If it can, I believe the correct exit code would likely be `SysErrInvalidReceiver`.

Send should only return:

https://github.com/filecoin-project/fvm-specs/blob/724650bf92b239bb7d2079cde7d5bad121d7c79f/07-errors.md#sendsend

shared/src/error/mod.rs
```diff
@@ -1,101 +1,116 @@
...
-)]
-pub enum ExitCode {
-    Ok = 0,
-
-    /// Indicates failure to find an actor in the state tree.
-    SysErrSenderInvalid = 1,
+/// ExitCode defines the exit code from the VM invocation.
+#[derive(PartialEq, Eq, Debug, Clone, Copy, Serialize, Deserialize)]
+#[serde(transparent)]
+pub struct ExitCode {
```
Stebalien, line 10: We should probably implement `From` on this.

Stebalien, line 10: We should also mark this as `#[repr(transparent)]`. That will ensure that the rust compiler represents it as a `u32` (in memory) and will make it safe to use in syscalls (as arguments & return values).

Stebalien: That would let us replace all `ExitCode::new(...)` with `ExitCall::from(...)` (which is generally a bit more idiomatic).

shared/src/error/mod.rs
```diff
@@ -1,101 +1,116 @@
...
+    pub const ILLEGAL_EXIT_CODE: ExitCode = ExitCode::new(9);
+    /// Indicates an internal VM assertion failed.
+    pub const ASSERTION_FAILED: ExitCode = ExitCode::new(10);
+    /// Indicates the actor returned a block handle that doesn't exist
+    pub const MISSING_RETURN: ExitCode = ExitCode::new(11);
+
+    pub const RESERVED_12: ExitCode = ExitCode::new(12);
+    pub const RESERVED_13: ExitCode = ExitCode::new(13);
+    pub const RESERVED_14: ExitCode = ExitCode::new(14);
+    pub const RESERVED_15: ExitCode = ExitCode::new(15);
```
Stebalien, line 82: I'd document but remove these. Otherwise, using them would be a breaking change.

shared/src/error/mod.rs
```diff
@@ -1,101 +1,116 @@
...
-    /// - failing to invoke caller validation
-    /// - aborting with a reserved exit code (including success or a system error).
-    SysErrIllegalActor = 9,
+/// Enumerates exit codes which originate inside the VM.
+/// These values may not be used by actors when aborting.
+pub struct SystemExitCode {}
 
-    /// Indicates an invalid argument passed to a runtime method.
-    SysErrIllegalArgument = 10,
+impl SystemExitCode {
```
Stebalien, line 51: Any reason not to attach these to `ExitCode`? I.e.

- `SYS_...`
- `USR_...`

The namespacing isn't great, but:

- Shorter names.
- Fewer imports.
- No constant-only structs (sometimes useful, but a bit strange).

fvm/src/executor/default.rs
anorth: I'll remove the case for now (so it becomes an unhandled error) so I can drop ILLEGAL_ACTOR. I'll then follow-up to address all possible error numbers explicitly. #438.

fvm/src/executor/default.rs
anorth: #438

shared/src/error/mod.rs
anorth: I implemented `From<u32>` but left `new` in place as I can't call `from()` from a `const` definition (and can't annotate it `const`).

shared/src/error/mod.rs
anorth: Ok, done.

anorth:
I've removed the single use of the serialization error, and would suggest for moving those exit code definitions into the builtin-actors repo soon. This would prevent the SDK in this repo from using any of them, though.

fvm/src/syscalls/vm.rs
```diff
@@ -18,11 +17,14 @@ pub fn abort(
...
-        //.filter(|c| !c.is_system_error())
-        .unwrap_or(ExitCode::SysErrIllegalActor); // TODO: will become "illegal exit"
+    let code = ExitCode::new(code);
+    // Uncomment to fix https://github.com/filecoin-project/ref-fvm/issues/253
+    // if code.is_system_error() {
+    //     return Err(Abort::Exit(
+    //         SystemExitCode::ILLEGAL_EXIT_CODE,
+    //         format!("actor aborted with code {}", code),
+    //     ));
+    // }
```
Stebalien, line 27: can we uncomment this with a network version check (for now)?

Stebalien: Eh, we can leave it till a future PR.

shared/src/error/mod.rs
Stebalien: Note: it's common to use `pub struct ExitCode(u32)` in cases like this, although I don't have strong opinions here.

Stebalien (approved):
I like it :+1:.

raulk (approved):
LGTM for a start.

I played with the type system to try to attain compile-time safety for ExitCode ranges in APIs. This does not eliminate the need to perform runtime correctness checks at untrusted/trusted boundaries.

I _think_ the `repr` here would work and translate to a single u32 value, but I'm not sure how to test it. And there might be ways to get rid of the PhantomData, but I didn't try too hard.

It also allows us to namespace the system and user codes better.

```rust
#[derive(PartialEq, Eq, Debug, Clone, Copy)]
#[repr(transparent)]
pub struct ExitCode<T = ()> {
    value: u32,
    _phantom: std::marker::PhantomData<T>,
}

impl ExitCode {
    pub const fn new<T>(value: u32) -> ExitCode<T> {
        ExitCode { value, _phantom: std::marker::PhantomData::<T> }
    }
}

#[derive(Debug)]
pub struct System;

#[derive(Debug)]
pub struct User;

impl ExitCode<System> {
    pub const A: ExitCode<System> = ExitCode::new(1);
    pub const B: ExitCode<System> = ExitCode::new(2);
}

impl ExitCode<User> {
    pub const C: ExitCode<User> = ExitCode::new(3);
    pub const D: ExitCode<User> = ExitCode::new(4);
}

fn main() {
    dbg!(ExitCode::<System>::A);
    dbg!(ExitCode::<System>::B);
    dbg!(ExitCode::<User>::C);
    dbg!(ExitCode::<User>::D);
}
```

In case you wanna play with it: https://play.rust-lang.org/?version=stable&mode=debug&edition=2021&gist=cb2577095b0a195e23a8e196949b345e

anorth:
Thanks @raulk, I'm interested to follow up exploring this later, but my Rust type-fu isn't strong enough yet. So I'll merge this and we can play with it.

dignifiedquire:
@raulk what use case are you trying to make work exactly/model?

Stebalien:
(I like the simplicity and lack of ::<>:: in the current version)

raulk:
@dignifiedquire I was just curious about other was in which the Thor system could help classify errors strongly, but weaker than the categorical enum (enum for error ranges).

PR #435, opened 2022-03-31
Magik audit clueanups

magik6k:


rust-toolchain
```diff
@@ -1 +1 @@
-stable
\ No newline at end of file
+nightly
```
Stebalien, line 1: Fixes build on aarch64 (for now). We can revert to stable once we've updated bellperson.

Stebalien (approved):

PR #436, opened 2022-03-31
doc: explain call-stack-depth logic

Stebalien:
The ordering of the checks here, and the comparison, are confusing (and
arguably incorrect but we're following the spec).

### DIFF
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -86,6 +86,19 @@ where
     where
         K: Kernel<CallManager = Self>,
     {
+        // We check _then_ set because we don't count the top call. This effectivly allows a
+        // call-stack depth of `max_call_depth + 1` (or `max_call_depth` sub-calls). While this is
+        // likely a bug, this is how NV15 behaves so we mimic that behavior here.
+        //
+        // By example:
+        //
+        // 1. If the max depth is 0, call_stack_depth will be 1 and the top-level message won't be
+        //    able to make sub-calls (1 > 0).
+        // 2. If the max depth is 1, the call_stack_depth will be 1 in the top-level message, 2 in
+        //    sub-calls, and said sub-calls will not be able to make further subcalls (2 > 1).
+        //
+        // NOTE: Unlike the FVM, Lotus adds _then_ checks. It does this because the
+        // `call_stack_depth` in lotus is 0 for the top-level call, unlike in the FVM where it's 1.
         if self.call_stack_depth > self.machine.config().max_call_depth {
             return Err(
                 syscall_error!(LimitExceeded, "message execution exceeds call depth").into(),


Stebalien:
Alternatively, we _could_ change this in nv16. I.e., a max call depth of 0 would mean that calls were forbidden, etc.

fvm/src/call_manager/default.rs
```diff
@@ -86,6 +86,18 @@ where
...
+        //    able to make sub-calls (1 > 0).
+        // 2. If the max depth is 1, the call_stack_depth will be 1 in the top-level message, 2 in
+        //    sub-calls, and said sub-calls will not be able to make further subcalls (2 > 1).
+        //
+        // NOTE: This means we can actually have a full "depth" of max_call_depth + 1, but that's
+        // how the network currently operates.
+        //
+        // NOTE: While lotus adds then checks (unlike the FVM which checks then adds), the
+        // call_stack_depth starts at 0 for the first-level while it's 1 in the FVM.
         if self.call_stack_depth > self.machine.config().max_call_depth {
```
raulk, line 101: @tchataigner This is one of the correctness behaviours I'd like to test via the integration test framework.

fvm/src/call_manager/default.rs
```diff
@@ -86,6 +86,18 @@ where
...
     {
+        // We check _then_ set because we don't count the top call. By example:
+        //
+        // 1. If the max depth is 0, call_stack_depth will be 1 and the top-level message won't be
+        //    able to make sub-calls (1 > 0).
+        // 2. If the max depth is 1, the call_stack_depth will be 1 in the top-level message, 2 in
+        //    sub-calls, and said sub-calls will not be able to make further subcalls (2 > 1).
+        //
+        // NOTE: This means we can actually have a full "depth" of max_call_depth + 1, but that's
+        // how the network currently operates.
```
raulk, line 97: We => FVM?
Network => Lotus?
I'm not following. Does nv15 allow an effective depth of 4095?

fvm/src/call_manager/default.rs
```diff
@@ -86,6 +86,18 @@ where
...
+        //
+        // NOTE: While lotus adds then checks (unlike the FVM which checks then adds), the
+        // call_stack_depth starts at 0 for the first-level while it's 1 in the FVM.
```
raulk, line 100: It's great to document this, but I find it a bit confusing. Because this is very much a correctness thing, I suspect some formality would help. Maybe:
- define terminology, e.g. top-level message, first-level message, top-call
- define implementation behaviours: what Lotus does and when, what FVM does and when
- define network behaviour pre-FVM and post-FVM
- elaborate on tradeoffs

fvm/src/call_manager/default.rs
Stebalien: >  I'm not following. Does nv15 allow an effective depth of 4095?

Yes. I've clarified that.

fvm/src/call_manager/default.rs
Stebalien: Is the new comment clearer?

The underlying tradaeoff is really: the code is different and the lotus model doesn't work here for "reasons". The goal of this comment is to explain how the FVM's version is correct (or, at least, matches lotus).

fvm/src/call_manager/default.rs
```diff
@@ -86,6 +86,19 @@ where
...
+        //
+        // By example:
+        //
+        // 1. If the max depth is 0, call_stack_depth will be 1 and the top-level message won't be
+        //    able to make sub-calls (1 > 0).
+        // 2. If the max depth is 1, the call_stack_depth will be 1 in the top-level message, 2 in
+        //    sub-calls, and said sub-calls will not be able to make further subcalls (2 > 1).
+        //
+        // NOTE: Unlike the FVM, Lotus adds _then_ checks. It does this because the
+        // `call_stack_depth` in lotus is 0 for the top-level call, unlike in the FVM where it's 1.
```
raulk, line 101: This is the key insight here; the equivalence between:

- FVM: 1-based with check-then-add
- Lotus: 0-based with add-then-check

raulk (approved):

PR #441, opened 2022-04-01
chore: randomness: Use array instead of Vec

magik6k:


Stebalien (approved):

Stebalien:
Marking this as "feature freeze" because we might need to land it. We can drop it from that list later if it doesn't make sense.

Kubuxu (approved):

Stebalien:
Closing as outdated. Half of this has been implemented (the sdk/kernel side). The rest of it will get punted to M2.2

PR #442, opened 2022-04-01
Clean up and simplify gas functionality

arajasek:
- Remove `verify_seal`, pretty sure this has been unused since pre-mainnet
- Replace the `charge_gas` syscall with a specialized `on_submit_verify_seal` call (better name welcome). This was the only use of the syscall in actors code, and I really don't like the idea of actors code asking the VM to charge arbitrary amounts of gas on a whim.

### DIFF
--- a/fvm/src/gas/price_list.rs
+++ b/fvm/src/gas/price_list.rs
@@ -136,6 +136,7 @@ lazy_static! {
 
         verify_consensus_fault: 495422,
         verify_replica_update: 36316136,
+        verify_seal_proof: 34721049,
         verify_post_lookup: [
             (
                 RegisteredPoStProof::StackedDRGWindow512MiBV1,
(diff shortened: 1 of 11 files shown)


raulk:
Agree with the removal of the charge_gas(units) syscall and the motivation for that, but the replacement is overly specific. One solution is to only inject this syscall in built-in actors. On the flip side, that creates precedent for a double standard when it comes to actors. I'd like to consider other possible solutions here like attempting to rejigger the syscall to move the gas calculation to the kernel.

Stebalien:
> On the flip side, that creates precedent for a double standard when it comes to actors

This isn't the only place where we'll need to do this. At a minimum, we'll need to restrict some syscalls to the init actor.

I'd also like to restrict some of the miner syscalls, honestly (to reduce the attack surface area).

raulk:
I've opened https://github.com/filecoin-project/ref-fvm/issues/468 to track a privilege system for syscalls. It isn't necessary for M1, though.

@arajasek could you scale back this PR to only remove the `verify_seal` pricelist item?

raulk:
Based on discuission above, this is not something we'll merge. Final solution will come by way of https://github.com/filecoin-project/ref-fvm/issues/468 in M2.

PR #443, opened 2022-04-02
Optimize RLE+ decoding by 20%

Kubuxu:
```
decode                  time:   [15.350 us 15.388 us 15.440 us]                    
                        change: [-19.388% -19.130% -18.910%] (p = 0.00 < 0.05)
                        Performance has improved.

decode_encode           time:   [30.458 us 30.542 us 30.659 us]                           
                        change: [-13.335% -13.014% -12.640%] (p = 0.00 < 0.05)
                        Performance has improved.
```

### DIFF
--- a/ipld/bitfield/src/rleplus/reader.rs
+++ b/ipld/bitfield/src/rleplus/reader.rs
@@ -9,16 +9,13 @@ const VARINT_MAX_BYTES: usize = 10;
 /// A `BitReader` allows for efficiently reading bits from a byte buffer, up to a byte at a time.
 ///
 /// It works by always storing at least the next 8 bits in `bits`, which lets us conveniently
-/// and efficiently read bits that cross a byte boundary. It's filled with the bits from `next_byte`
-/// after every read operation, which is in turn replaced by the next byte from `bytes` as soon
-/// as the next read might read bits from `next_byte`.
+/// and efficiently read bits that cross a byte boundary.
 pub struct BitReader<'a> {
     /// The bytes that have not been read from yet.
     bytes: &'a [u8],
-    /// The next byte from `bytes` to be added to `bits`.
-    next_byte: u8,
     /// The next bits to be read.
-    bits: u16,
+    bits: u64,
+
     /// The number of bits in `bits` from bytes that came before `next_byte` (at least 8, at most 15).
     num_bits: u32,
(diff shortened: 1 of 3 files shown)


Kubuxu:
I spent few hours learning about profiling in Rust, this is the result.

Kubuxu:
It might also not make sense due to fuel accounting.

raulk:
Mind elaborating on the latter comment? Are you seeing much variability, and did you run the profiling on Wasm bytecode?

raulk:
Also, how invasive did you find the profiler to be?

Kubuxu:
Regarding the fuel comment, in the fuel world you are incentivised to optimise for fuel not for speed.

In this case I profiled out of WASM, believing that better/faster x86 will result in better llvm-or and thus faster WASM.

Profiler only runs when you pass the —profile-time flag to criterion. Thinking about it more maybe I should add feature flag for it as people might be missing gperftools shared lib.

ipld/bitfield/src/rleplus/reader.rs
```diff
@@ -31,51 +28,59 @@ impl<'a> BitReader<'a> {
...
         self.num_bits -= num_bits;
 
-        // this unconditionally adds the next byte to `bits`,
-        // regardless of whether there's enough space or not. the
-        // point is to make sure that `bits` always contains
-        // at least the next 8 bits to be read
-        self.bits |= (self.next_byte as u16) << self.num_bits;
+        // not sure why this being outside of the if improves the performance
+        // bit it does, probably related to keeping caches warm
+        let byte = self.bytes.get(0).unwrap_or(&0);
```
Stebalien, line 67:
```suggestion
        let byte = self.bytes.first().unwrap_or(&0);
```

Stebalien (approved):

Stebalien:
@Kubuxu when you optimized, what compiler flags did you pass? I ask because we usually use Oz/Os in actors. We _might_ need to add a few `#[inline(always)]` comments here given that we're adding a few new function calls.

Kubuxu:
I used the default release profile that is in ref-fvm. I guess it might behave a bit different with Oz, I can retest.

Stebalien:
Then either add it or don't, up to you, but merge when ready.

PR #446, opened 2022-04-04
Make signature properties public

rllola:
In order to use `#[serde(remote = "Signature")]` it is required to have its fields defined as public.

I need to be able to define how to serialize/deserialize message in json and for that I use serde remote trick. The Signature is part of a SignedMessage struct that I want to serialize/deserialize.


raulk:
@Stebalien I suspect `#[serde(with = "")]` won't work in this case because @rllola is trying to avoid creating newtypes for every single type they want to be able to serialize in JSON (basically _every_ type that can appear in messages -- and possibly state?).

The problem is that the serde strategy at origin is tuples, so my understanding is that @rllola is trying to override that strategy "remotely".

@rllola is that correct?

PR #449, opened 2022-04-04
Use repr(packed, C) for syscall types

Stebalien:
I was hoping to _not_ do this as it might have a perf implication, but:

- Padding is undefined so we could end up leaking rust memory (and causing non-determinism at runtime).
- Requiring that everything be "packed" simplifies the spec. We don't need to explain alignment, just:
    - little endian byte order
    - packed
- In practice, there should be no performance issues (with existing syscalls). Specifically, we can re-order fields to avoid unaligned loads. This PR takes care of that.

Part of https://github.com/filecoin-project/fvm-specs/issues/63.

### DIFF
--- a/shared/src/sys/out.rs
+++ b/shared/src/sys/out.rs
@@ -11,8 +11,8 @@ pub mod actor {
     #[derive(Debug, Copy, Clone)]
     #[repr(packed, C)]
     pub struct ResolveAddress {
-        pub resolved: i32,
         pub value: u64,
+        pub resolved: i32,
     }
 }
 
--- a/shared/src/sys/out.rs
+++ b/shared/src/sys/out.rs
@@ -7,6 +7,13 @@
 //!
 //! Read more at https://github.com/rust-lang/rust/issues/73755.
 
+// NOTE: When possible, pack fields such that loads will be power-of-two aligned. Un-aligned loads
+// _can_ be done (LLVM will generate the appropriate code) but are slower.
+//
+// Read up on https://doc.rust-lang.org/reference/type-layout.html for more information.
+//
+// Also, please also read the docs on super::SyscallSafe before modifying any of these types.
+
 pub mod actor {
     #[derive(Debug, Copy, Clone)]
     #[repr(packed, C)]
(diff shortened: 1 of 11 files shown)


shared/src/sys/out.rs
```diff
@@ -8,26 +8,25 @@
 //! Read more at https://github.com/rust-lang/rust/issues/73755.
 
 pub mod actor {
-    #[repr(C)]
     #[derive(Debug, Copy, Clone)]
+    #[repr(packed, C)]
     pub struct ResolveAddress {
         pub resolved: i32,
         pub value: u64,
```
raulk, line 15: Should we invert these too so value is left aligned?

shared/src/sys/out.rs
```diff
@@ -8,26 +8,25 @@
...
     }
 }
 
 pub mod ipld {
-    #[repr(C)]
     #[derive(Debug, Copy, Clone)]
+    #[repr(packed, C)]
     pub struct IpldOpen {
```
raulk, line 22: For all structs, can we add a note aimed as SDK implementers to explain that this is not necessarily the recommended argument order for SDK methods, but rather an alignment-optimised order? The first intuition is to expand structs to SDK function arguments honouring the order, which can lead to awkward APIs.

raulk (approved):

raulk:
@Stebalien clippy is angry because of a missing doc comment.

Stebalien:
Unfortunately, CI is not going to be green for this merge.

Release plan:

1. Merge this PR. CI = :angry: 
2. Release shard/sdk.
3. Update actors v6 & v7 to the new shared/sdk releases.
4. Release actors v6 & v7.
5. Update actors v6 & v7 in the conformance tests (here). CI = :smile: 
6. Release fvm v0.7.0.
7. Update the FFI to fvm v0.7.0

Stebalien:
> Unfortunately, CI is not going to be green for this merge.

Basically, the issue is that we have a "workspace" here, so the conformance tests will always use the latest version of the fvm_shared crate, even if it's incompatible.

arajasek:
> 6\. Release fvm v0.7.0.

Sounds like a reasonable incantation.

PR #452, opened 2022-04-05
Implement FIP-0032

arajasek:


fvm/src/call_manager/default.rs
```diff
@@ -294,14 +330,15 @@ where
...
-            let result: std::result::Result<RawBytes, Abort> = (|| {
+            let mut result: std::result::Result<RawBytes, Abort> = (|| {
                 // Lookup the invoke method.
                 let invoke: wasmtime::TypedFunc<(u32,), u32> = instance
                     .get_typed_func(&mut store, "invoke")
                     // All actors will have an invoke method.
                     .map_err(Abort::Fatal)?;
 
                 // Invoke it.
+                // TODO: I'm not sure how to actually check the Trap here for OutOfFuel (as opposed to any other failure).
```
arajasek, line 341: This is the biggest open question.

arajasek: This is the underlying method in Wasmtime, but hooowww can I sniff this out 

```
    fn consume_fuel(&mut self, fuel: u64) -> Result<u64> {
        let consumed_ptr = unsafe { &mut *self.interrupts.fuel_consumed.get() };
        match i64::try_from(fuel)
            .ok()
            .and_then(|fuel| consumed_ptr.checked_add(fuel))
        {
            Some(consumed) if consumed < 0 => {
                *consumed_ptr = consumed;
                Ok(u64::try_from(-consumed).unwrap())
            }
            _ => bail!("not enough fuel remaining in store"),
        }
    }
```

fvm/src/call_manager/default.rs
```diff
@@ -153,6 +160,22 @@ where
         self.num_actors_created += 1;
         ret
     }
+
+    fn get_fuel_consumed_snapshot(&self) -> u64 {
```
Stebalien, line 164: You should be able to store these on the `InvocationContext` itself. We _cannot_ set them on the CallManager as sub-calls will clobber the values stored by parents.

Stebalien: It also means you can avoid modifying the kernel and call manager.

fvm/src/call_manager/default.rs
Stebalien: This is where fuel is consumed when _you_ manually call `consume_fuel`. When the actor runs out of fuel, I believe you get:

https://github.com/bytecodealliance/wasmtime/blob/6c8c94723a842258a4ead8e92eeb0a745a57aa64/crates/wasmtime/src/store.rs#L1805

But looking a the code... it looks like wasmtime will:

1. Consume the amount of fuel necessary to perform some operation.
2. Check to make sure it didn't go over the budget.
3. If it did, it'll abort.

So I _think_ we can just check `fuel_consumed > gas_to_fuel(gas_remaining)`. Specifically, we need to check that when the wasm call _fails_.

fvm/src/call_manager/default.rs
arajasek: Ah, yes, we discussed then in person, but it slipped my mind. Thanks!

fvm/src/gas/price_list.rs
```diff
@@ -32,9 +32,117 @@ lazy_static! {
...
+                RegisteredPoStProof::StackedDRGWindow64GiBV1,
+                ScalingCost {
+                    flat: 117680921,
+                    scale: 43780,
+                },
+            ),
+        ]
+        .iter()
+        .copied()
+        .collect(),
```
dignifiedquire, line 119: what’s the type you are collecting into here?

arajasek: I copy-pastad here, will check in an hour-ish.

fvm/src/call_manager/default.rs
```diff
@@ -294,7 +295,23 @@ where
             };
 
             // Make a store.
+            let gas_available = kernel.gas_available();
+            let fuel_to_add = match kernel.network_version() {
+                NetworkVersion::V14 => u64::MAX,
+                NetworkVersion::V15 => u64::MAX,
```
Stebalien, line 301:
```suggestion
                NetworkVersion::V14|NetworkVersion::V15 => u64::MAX,
```

fvm/src/call_manager/default.rs
```diff
@@ -309,18 +326,40 @@ where
             // From this point on, there are no more syscall errors, only aborts.
             let result: std::result::Result<RawBytes, Abort> = (|| {
                 // Lookup the invoke method.
-                let invoke: wasmtime::TypedFunc<(u32,), u32> = instance
+                let invoke: wasmtime::TypedFunc<(u32, ), u32> = instance
```
Stebalien, line 329: ?

fvm/src/call_manager/default.rs
```diff
@@ -309,18 +326,40 @@ where
...
                     .get_typed_func(&mut store, "invoke")
                     // All actors will have an invoke method.
                     .map_err(Abort::Fatal)?;
 
                 // Invoke it.
-                let return_block_id = invoke.call(&mut store, (param_id,))?;
+                let res = invoke.call(&mut store, (param_id, ));
```
Stebalien, line 335: Does rustfmt not hate you?

fvm/src/call_manager/default.rs
```diff
@@ -309,18 +326,40 @@ where
...
+                let fuel_consumed = match store
+                    .fuel_consumed() {
+                    Some(v) => v,
+                    None => 0,
+                };
```
Stebalien, line 340: This will be the fuel consumed over the course of the entire execution (_including_ fuel manually consumed to account for syscalls). We need to charge for the fuel consumed since the last syscall.

fvm/src/syscalls/bind.rs
```diff
@@ -94,6 +94,35 @@ fn memory_and_data<'a, K: Kernel>(
...
+    caller: &mut Caller<InvocationData<K>>,
+) -> Result<(), Trap> {
+    let gas_available = caller.data().kernel.gas_available();
+    let gas_used = gas_available - caller.data().gas_available_snapshot;
+    let fuel_used = caller.data().kernel.price_list().gas_to_fuel(gas_used);
+    caller.consume_fuel(fuel_used)?;
+
+    caller.data_mut().gas_available_snapshot = gas_available;
+    Ok(())
+}
```
Stebalien, line 107: Just make this a method on the invocation data, then call `consume_fuel` afterwards.

fvm/src/syscalls/bind.rs
```diff
@@ -94,6 +94,35 @@ fn memory_and_data<'a, K: Kernel>(
     Ok((Memory::new(mem), data))
 }
 
+fn update_gas_available_snapshot<K: Kernel>(
```
Stebalien, line 97: This name isn't really accurate.

Stebalien (changes requested):

fvm/src/call_manager/default.rs
arajasek: It seems to...want this?

Stebalien: Not _my_ rustfmt...

fvm/src/gas/price_list.rs
Stebalien: A map.

fvm/src/call_manager/default.rs
arajasek: #NotMyRustfmt

fvm/src/call_manager/default.rs
```diff
@@ -294,7 +295,19 @@ where
             };
 
             // Make a store.
+            let gas_available = kernel.gas_available();
```
raulk, line 298: Prefer to get the gas available from the gas tracker directly (owned by the CallManager) outside the closure. Getting it from the kernel at this stage is confusing.

fvm/src/call_manager/default.rs
```diff
@@ -294,7 +295,19 @@ where
...
+            let fuel_to_add = match kernel.network_version() {
+                NetworkVersion::V14 | NetworkVersion::V15 => u64::MAX,
```
raulk, line 300: Ouch, I don't like this one bit. Is it not possible to keep fuel metering on but not set a fuel limit (i.e. we meter fuel, but we never halt)?

Kubuxu: u64::MAX should be large enough to be impossible to reach by incrementing but I understand where you are coming from.

fvm/src/call_manager/default.rs
```diff
@@ -315,12 +328,26 @@ where
                     .map_err(Abort::Fatal)?;
 
                 // Invoke it.
-                let return_block_id = invoke.call(&mut store, (param_id,))?;
+                let res = invoke.call(&mut store, (param_id,));
+
+                // First, charge gas for the "latest" fuel use (all the fuel used since the most recent syscall)
+                let fuel_consumed = store
+                    .fuel_consumed()
+                    // TODO: Check whether this ever returns None (eg if 0 fuel has been consumed)
```
raulk, line 336: This can only return `None` when fuel consumption is disabled in the configuration.

fvm/src/syscalls/mod.rs
```diff
@@ -29,17 +30,56 @@ pub struct InvocationData<K> {
...
     pub last_error: Option<backtrace::Cause>,
+
+    /// A snapshot is _taken_ every time execution returns _from_ WASM.
+    /// This happens at the start of a syscall invocation as well as the end of WASM invocation.
+    /// The snapshot is _updated_ every time execution moves _into_ WASM.
+    /// This happens at the first WASM invocation, and at the end of a syscall invocation.
+    /// Before a snapshot is _updated_, fuel is consumed based on the delta.
+    pub gas_available_snapshot: i64,
+
+    /// A snapshot is _taken_ every time execution moves _into_ WASM.
```
raulk, line 41: These comments are mildly confusing. After reading them it's not clear to me what the difference between taking a snapshot and updating a snapshot is?

fvm/src/syscalls/mod.rs
```diff
@@ -29,17 +30,56 @@ pub struct InvocationData<K> {
...
+    pub fuel_consumed_snapshot: u64,
 }
 
 impl<K> InvocationData<K> {
-    pub(crate) fn new(kernel: K) -> Self {
+    pub(crate) fn new(kernel: K, gas_available: i64) -> Self {
         Self {
             kernel,
             last_error: None,
+            gas_available_snapshot: gas_available,
```
raulk, line 54: If you move this constructor to the impl below, you should be able to get the gas available from the call manager -> gas tracker without having to pass it in explicitly.

fvm/src/gas/price_list.rs
```diff
@@ -244,6 +357,17 @@ pub struct PriceList {
     pub(crate) verify_post_discount: bool,
     pub(crate) verify_consensus_fault: i64,
     pub(crate) verify_replica_update: i64,
+    // 1 Fuel = gas_per_fuel * 1 Gas
+    pub(crate) gas_per_fuel: i64,
+
+    pub(crate) extern_traversal_cost: i64,
+    // TODO: Some of these will probably need per-byte multipliers
+    pub(crate) block_open: i64,
```
raulk, line 365: This will need a per byte multiplier since the open will retain the block content in Rust memory.

fvm/src/gas/price_list.rs
```diff
@@ -244,6 +357,17 @@ pub struct PriceList {
...
+    pub(crate) block_read: i64,
+    pub(crate) block_create: i64,
```
raulk, line 367: This will need a per byte multiplier since the write will retain the block content in Rust memory until the execution finishes.

fvm/src/gas/price_list.rs
```diff
@@ -244,6 +357,17 @@ pub struct PriceList {
...
+    pub(crate) block_link_base: i64,
```
raulk, line 368: Note: The base needs to account for the cost of CID calculation.

fvm/src/gas/price_list.rs
```diff
@@ -244,6 +357,17 @@ pub struct PriceList {
...
+    pub(crate) block_link_per_byte: i64,
```
raulk, line 369: This writes the blockstore from the block registry to the buffered blockstore (memory). It won't ultimately write to the disk blockstore until Machine flush, at which point we do copy_rec and only write it if it's reachable.

fvm/src/gas/price_list.rs
```diff
@@ -244,6 +357,17 @@ pub struct PriceList {
     pub(crate) verify_post_discount: bool,
     pub(crate) verify_consensus_fault: i64,
     pub(crate) verify_replica_update: i64,
+    // 1 Fuel = gas_per_fuel * 1 Gas
+    pub(crate) gas_per_fuel: i64,
+
+    pub(crate) extern_traversal_cost: i64,
+    // TODO: Some of these will probably need per-byte multipliers
```
raulk, line 364: I would introduce two variables here:

- `block_memcpy_per_byte_cost`
- `block_io_per_byte_cost`

These two factors composed appropriately at the syscall site should be enough to apply the appropriate dynamic cost portion per syscall.

fvm/src/gas/price_list.rs
```diff
@@ -416,9 +526,74 @@ impl PriceList {
     pub fn on_verify_consensus_fault(&self) -> GasCharge<'static> {
         GasCharge::new("OnVerifyConsensusFault", self.verify_consensus_fault, 0)
     }
+
+    /// Returns the gas required for the specified amount of fuel.
+    #[inline]
+    pub fn on_consume_fuel(&self, fuel: u64) -> GasCharge<'static> {
+        GasCharge::new("OnConsumeFuel", self.gas_per_fuel * fuel as i64, 0)
```
raulk, line 533: Can we call this "execution units" (like in FIP-0032) (maybe abbreviate to ExecUnits). Fuel is a wasmtime-specific term which won't apply to other engines.

We may want to remove the term `fuel` from the interface of our components too -- we should only use this term privately in the context of wasmtime.

fvm/src/gas/price_list.rs
```diff
@@ -416,9 +526,74 @@ impl PriceList {
...
+    }
+
+    /// Converts the specified fuel into equivalent units of gas
+    #[inline]
+    pub fn fuel_to_gas(&self, fuel: u64) -> i64 {
+        (self.gas_per_fuel * fuel as i64) as i64
```
raulk, line 539: Needs overflow check.

fvm/src/syscalls/bind.rs
raulk: @arajasek I think you solved this in a follow up commit, but the commit only says "implement fip-0032". Could you please loop back here?

fvm/src/call_manager/default.rs
Stebalien: Wasmtime actually only supports up to i64::MAX fuel. If you pass u64::MAX, it'll treat it as "infinity".

raulk:
Sorry for the stream of tiny reviews. It's VS Code's GitHub extension's fault. Haven't gotten a chance to review the fuel => gas and vice versa conversions and snapshotting logic yet.

fvm/src/syscalls/bind.rs
arajasek: Yes, it was solved.

fvm/src/call_manager/default.rs
arajasek: I don't think that works because we store the params inside the kernel which can lead to a change in gas_available -- am I missing something?

fvm/src/call_manager/default.rs
arajasek: Yeah, I figured it was fine based on [this](https://github.com/bytecodealliance/wasmtime/blob/main/crates/wasmtime/src/store.rs#L1290-L1296), and of course this is only relevant for testing on v14/15.

fvm/src/call_manager/default.rs
arajasek: Good, so long as we avoid doing that, we're good.

fvm/src/syscalls/mod.rs
arajasek: Clarifying, thanks for the feedback.

fvm/src/syscalls/mod.rs
arajasek: SG, thanks!

fvm/src/gas/price_list.rs
arajasek: This is gonna get hard to shoehorn to fit both v15 and v16 -- in v15 we only have the per-byte cost for link. If we replace `block_link_per_byte` with `block_io_per_byte`, and start using `block_io_per_byte` for operations like `block_open`, we'll break consensus with v15.

I'm gonna leave `block_link_per_byte` in for now to get around this, but otherwise incorporate your suggestions.

fvm/src/call_manager/mod.rs
```diff
@@ -11,8 +11,11 @@ use crate::state_tree::StateTree;
 use crate::Kernel;
 
 pub mod backtrace;
+
```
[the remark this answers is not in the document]
arajasek: i...think so? but rustfmt doesn't seem to be putting it back in if i take it out.

arajasek:
Addressed most of the review comments. I think it might be useful to work through a somewhat complicated message with multiple `Send`s and syscalls, just to make sure things are always tallied correctly.

A test would be great, but Idk if there's a nice framework for that.

fvm/src/gas/price_list.rs
```diff
@@ -416,9 +537,91 @@ impl PriceList {
...
+            0,
+        ))
+    }
+
+    /// Converts the specified gas into equivalent exec_units
+    #[inline]
+    pub fn gas_to_exec_units(&self, gas: i64) -> u64 {
+        match self.gas_per_exec_unit {
+            0 => 0,
+            v => (gas / v) as u64,
```
Stebalien, line 561: So, rounding here is tricky. Especially if we want to be accurate:

1. Round down when setting the fuel limit.
2. When charging gas for fuel:
    1. Convert fuel to gas, rounding up. Charge for that.
    2. Set the fuel snapshot to `gas_to_exec_units(gas_CHARGED)`. This will be larger than the actual fuel used, but will prevent us from double-charging for fuel.

Now, maybe that's overkill. Alternatively, we could just "round up" in case 2 and leave it at that. If the actor uses 1 fuel, makes a syscall, then uses 1 fuel, we'd charge 2 gas (instead of 1 gas). But maybe that doesn't matter?

fvm/src/gas/price_list.rs
```diff
@@ -416,9 +537,91 @@ impl PriceList {
...
+        GasCharge::new("OnExternTraversal", self.extern_traversal_cost, 0)
+    }
+
+    /// Returns the gas required for loading an object.
+    #[inline]
+    pub fn on_block_open(&self, data_size: usize) -> GasCharge<'static> {
+        // TODO: Should we also throw on a memcpy cost here (see https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0032.md#ipld-state-management-fees)
+        GasCharge::new(
+            "OnBlockOpen",
+            self.block_open_base + data_size as i64 * self.block_io_per_byte_cost,
```
Stebalien, line 577: overflow checks here and everywhere.

fvm/src/gas/price_list.rs
```diff
@@ -416,9 +537,91 @@ impl PriceList {
...
 
 /// Returns gas price list by NetworkVersion for gas consumption.
-pub fn price_list_by_network_version(_: NetworkVersion) -> PriceList {
-    OH_SNAP_PRICES.clone()
+pub fn price_list_by_network_version(network_version: NetworkVersion) -> PriceList {
+    match network_version {
+        NetworkVersion::V14 => OH_SNAP_PRICES.clone(),
+        NetworkVersion::V15 => OH_SNAP_PRICES.clone(),
+        _ => SKYR_PRICES.clone(),
+    }
```
Stebalien, line 626: Hm. Can we return `&'static PriceList`?

fvm/src/call_manager/default.rs
```diff
@@ -315,12 +328,26 @@ where
...
+                    .ok_or_else(|| Abort::Fatal(anyhow!("expected to find fuel consumed")))?;
```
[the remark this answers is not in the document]
arajasek: Ack

arajasek:
@Stebalien 

> That should be included in the per-byte cost.

Need them for v15 and earlier, me thinks?

fvm/src/kernel/default.rs
```diff
@@ -130,20 +130,27 @@ where
...
-            .or_fatal()?; // because the state should exist.
+            .or_fatal()?;
+
+        if charge_gas {
+            self.call_manager.charge_gas(
+                self.call_manager
+                    .price_list()
+                    .on_block_open(state_block.len()),
+            )?;
+        }
```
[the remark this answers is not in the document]
arajasek: We can, but that's a detail for the FIP, I think?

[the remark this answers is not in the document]
Stebalien: Not really, we'll run out of gas either way (doesn't matter for consensus). The question is how much extra unpaid work we do.

fvm/src/kernel/mod.rs
```diff
@@ -218,9 +219,16 @@ pub trait CircSupplyOps {
...
+    /// GasAvailable return the gas available for the transaction.
+    fn gas_available(&self) -> i64;
+
     /// ChargeGas charges specified amount of `gas` for execution.
     /// `name` provides information about gas charging point
     fn charge_gas(&mut self, name: &str, compute: i64) -> Result<()>;
+
+    fn charge_exec_units(&mut self, exec_units: u64) -> Result<()>;
+
+    fn price_list(&self) -> &PriceList;
```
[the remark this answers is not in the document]
arajasek: Fair point.

fvm/src/syscalls/bind.rs
```diff
@@ -94,6 +94,27 @@ fn memory_and_data<'a, K: Kernel>(
     Ok((Memory::new(mem), data))
 }
 
+fn charge_exec_units_for_gas<K: Kernel>(
+    caller: &mut Caller<InvocationData<K>>,
+) -> Result<(), Trap> {
```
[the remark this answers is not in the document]
arajasek: Yessir.

Stebalien:
> Need them for v15 and earlier, me thinks?

No? We didn't charge for hashing per-byte before nv15 (we just had a lot of strict limits).

fvm/src/gas/price_list.rs
arajasek: Yessir.

fvm/src/syscalls/bind.rs
```diff
@@ -114,26 +135,32 @@ macro_rules! impl_bind_syscalls {
                 if mem::size_of::<Ret::Value>() == 0 {
                     // If we're returning a zero-sized "value", we return no value therefore and expect no out pointer.
                     self.func_wrap(module, name, move |mut caller: Caller<'_, InvocationData<K>> $(, $t: $t)*| {
-                        let (mut memory, mut data) = memory_and_data(&mut caller)?;
+                        let caller_mut = &mut caller;
```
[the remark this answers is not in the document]
arajasek: Fair enough, no.

fvm/src/syscalls/bind.rs
```diff
@@ -114,26 +135,32 @@ macro_rules! impl_bind_syscalls {
...
+                        charge_gas_for_exec_units(caller_mut)?;
+                        let (mut memory, mut data) = memory_and_data(caller_mut)?;
                         let ctx = Context{kernel: &mut data.kernel, memory: &mut memory};
                         Ok(match syscall(ctx $(, $t)*).into()? {
                             Ok(_) => {
                                 log::trace!("syscall {}::{}: ok", module, name);
                                 data.last_error = None;
+                                charge_exec_units_for_gas(caller_mut)?;
```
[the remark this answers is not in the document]
arajasek: thank youuuuu this was driving me insane

arajasek:
@Stebalien Ah, sorry, GitHub bad UI made me think you were starting a new discussion, not responding to Raul.

fvm/src/kernel/default.rs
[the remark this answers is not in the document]
arajasek: Still TODO.

fvm/src/gas/price_list.rs
arajasek: Still TODO.

fvm/src/gas/price_list.rs
```diff
@@ -540,17 +539,14 @@ impl PriceList {
...
-        exec_units: u64,
-    ) -> Result<GasCharge<'static>, anyhow::Error> {
-        Ok(GasCharge::new(
+    pub fn on_consume_exec_units(&self, exec_units: u64) -> GasCharge<'static> {
+        GasCharge::new(
             "OnConsumeExecUnits",
             self.gas_per_exec_unit
                 .checked_mul(exec_units as i64)
-                .ok_or_else(|| anyhow!("overflow when consuming exec_units"))?,
+                .unwrap_or(i64::MAX),
```
Stebalien, line 547: https://doc.rust-lang.org/std/primitive.i64.html#method.saturating_mul

fvm/src/gas/price_list.rs
```diff
@@ -562,6 +558,12 @@ impl PriceList {
         }
     }
 
+    /// Converts the specified exec_units into equivalent gas
+    #[inline]
+    pub fn exec_units_to_gas(&self, exec_units: u64) -> Option<i64> {
+        self.gas_per_exec_unit.checked_mul(exec_units as i64)
```
Stebalien, line 564: Also saturate?

arajasek:
Ready for another round of review.

fvm/src/call_manager/default.rs
```diff
@@ -315,12 +328,26 @@ where
                     .map_err(Abort::Fatal)?;
 
                 // Invoke it.
-                let return_block_id = invoke.call(&mut store, (param_id,))?;
+                let res = invoke.call(&mut store, (param_id,));
+
+                // First, charge gas for the "latest" use of execution units (all the exec units used since the most recent syscall)
```
Stebalien, line 333: We should clarify that the variable below is "total units consumed".

fvm/src/gas/price_list.rs
```diff
@@ -122,6 +234,26 @@ lazy_static! {
...
+
+        block_memcpy_per_byte_cost: 4,
+        block_io_per_byte_cost: 2,
+        block_link_per_byte_cost: 1,
+        // TODO: PARAM_FINISH
+
+        // TODO: PARAM_FINISH
+        gas_per_exec_unit: 2,
+        // TODO: PARAM_FINISH
+        extern_traversal_cost: 1,
```
Stebalien, line 246: traverse what?

fvm/src/gas/price_list.rs
```diff
@@ -416,9 +536,104 @@ impl PriceList {
     pub fn on_verify_consensus_fault(&self) -> GasCharge<'static> {
         GasCharge::new("OnVerifyConsensusFault", self.verify_consensus_fault, 0)
     }
+
+    /// Returns the gas required for the specified exec_units.
+    #[inline]
+    pub fn on_consume_exec_units(&self, exec_units: u64) -> GasCharge<'static> {
+        GasCharge::new(
+            "OnConsumeExecUnits",
+            self.gas_per_exec_unit.saturating_mul(exec_units as i64),
```
Stebalien, line 545: to be safe, we should either take `exec_units` as an i64, or do a saturating conversion. I mean, this is almost _certainly_ safe... but we don't want to wrap around.

fvm/src/gas/price_list.rs
```diff
@@ -416,9 +536,104 @@ impl PriceList {
...
+                true => gas.div_floor(v) as u64,
+                false => gas.div_ceil(v) as u64,
+            },
+        }
+    }
+
+    /// Converts the specified exec_units into equivalent gas
+    #[inline]
+    pub fn exec_units_to_gas(&self, exec_units: u64) -> i64 {
+        self.gas_per_exec_unit.saturating_mul(exec_units as i64)
```
Stebalien, line 565: Again, I'd be slightly careful about wrapping.

fvm/src/kernel/default.rs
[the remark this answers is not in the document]
Stebalien: Although we do have to be careful about errors... But I think it's safe to charge for "access" up-front, then charge per-byte afterwards.

fvm/src/kernel/default.rs
```diff
@@ -295,8 +299,14 @@ where
             // to be in the state-tree.
             .or_fatal()?;
 
-        // We charge on open, not read, to emulate the current gas model.
         let block = Block::new(cid.codec(), data);
+
+        // We charge on open, not read, to emulate the current gas model.
```
Stebalien, line 304: Remove comment. We now charge on both.

fvm/src/kernel/default.rs
```diff
@@ -342,8 +355,12 @@ where
         Ok(k)
     }
 
-    fn block_read(&self, id: BlockId, offset: u32, buf: &mut [u8]) -> Result<u32> {
+    fn block_read(&mut self, id: BlockId, offset: u32, buf: &mut [u8]) -> Result<u32> {
         let data = self.blocks.get(id).or_illegal_argument()?.data();
+
+        self.call_manager
+            .charge_gas(self.call_manager.price_list().on_block_read(data.len()))?;
```
Stebalien, line 362: We've already charged to "load" the block, now we just want to charge for the amount of data read (the length we return from this function).

fvm/src/kernel/default.rs
```diff
@@ -558,6 +578,8 @@ where
         h2: &[u8],
         extra: &[u8],
     ) -> Result<Option<ConsensusFault>> {
+        self.call_manager
+            .charge_gas(self.call_manager.price_list().on_extern_traversal())?;
```
Stebalien, line 582: @raulk we should discuss this. The fact that this traverses an extern is an implementation detail and the actual _cost_ of that depends on the implementation. We should bundle the cost into the verify consensus fault syscall itself.

fvm/src/kernel/default.rs
```diff
@@ -743,11 +779,13 @@ where
...
     #[allow(unused)]
     fn get_randomness_from_tickets(
-        &self,
+        &mut self,
         personalization: DomainSeparationTag,
         rand_epoch: ChainEpoch,
         entropy: &[u8],
     ) -> Result<[u8; RANDOMNESS_LENGTH]> {
+        self.call_manager
+            .charge_gas(self.call_manager.price_list().on_extern_traversal())?;
```
Stebalien, line 788: @raulk we need to discuss this as well. We need to be charging a randomness fee (base + multiplier * randomness_size) not an extern traversal fee.

fvm/src/lib.rs
```diff
@@ -1,10 +1,13 @@
+#![feature(int_roundings)]
```
Stebalien, line 1: We shouldn't use nightly features.

fvm/src/lib.rs
```diff
@@ -1,10 +1,13 @@
...
 //! (Proper package docs coming shortly; for now this is a holding pen for items
 //! we must mention).
 //!
 //! ## Logging
 //!
 //! This package emits logs using the log façade. Configure the logging backend
 //! of your choice during the initialization of the consuming application.
+extern crate anyhow;
```
Stebalien, line 9: This isn't necessary.

fvm/src/syscalls/bind.rs
```diff
@@ -114,9 +134,10 @@ macro_rules! impl_bind_syscalls {
                 if mem::size_of::<Ret::Value>() == 0 {
                     // If we're returning a zero-sized "value", we return no value therefore and expect no out pointer.
                     self.func_wrap(module, name, move |mut caller: Caller<'_, InvocationData<K>> $(, $t: $t)*| {
+                        charge_gas_for_exec_units(&mut caller)?;
                         let (mut memory, mut data) = memory_and_data(&mut caller)?;
                         let ctx = Context{kernel: &mut data.kernel, memory: &mut memory};
-                        Ok(match syscall(ctx $(, $t)*).into()? {
+                        let result = match syscall(ctx $(, $t)*).into()? {
```
Stebalien, line 140: Note to self: If we trap here, we _don't_ need to "charge exec units for gas", because we won't be executing anything else.

fvm/src/syscalls/bind.rs
```diff
@@ -144,20 +169,25 @@ macro_rules! impl_bind_syscalls {
...
                                 let code = err.1;
                                 log::trace!("syscall {}::{}: fail ({})", module, name, code as u32);
                                 data.last_error = Some(backtrace::Cause::new(module, name, err));
+                                charge_exec_units_for_gas(&mut caller)?;
                                 code as u32
                             },
-                        })
+                        };
+
+                        charge_exec_units_for_gas(&mut caller)?;
```
Stebalien, line 189: we're calling this twice.

fvm/src/syscalls/mod.rs
```diff
@@ -29,15 +30,63 @@ pub struct InvocationData<K> {
...
+    /// 3) updates the available_gas and exec_units_consumed snapshots. The exec_units_consumed_snapshot is optimistically updated, assuming the value calculated in 2 will be consumed
+    /// 4) returns the value calculated in 2) for its caller to actually consume that exec_units_consumed
+    pub(crate) fn calculate_exec_units_for_gas(&mut self) -> Result<u64> {
+        let gas_available = self.kernel.gas_available();
+        let gas_used = self.gas_available_snapshot - gas_available;
+        if gas_used < 0 {
+            return Err(ExecutionError::Syscall(SyscallError(
+                String::from("used more gas than available"),
+                fvm_shared::error::ErrorNumber::IllegalOperation,
+            )));
```
Stebalien, line 68: On reflection, I don't think this is correct. Technically we can refund gas (and this _definitely_ isn't a syscall error).

I _think_ we need to add fuel (which we can do). And I _don't_ think that causes any other problems.... but we'll have to be careful.

fvm/src/syscalls/mod.rs
```diff
@@ -29,15 +30,63 @@ pub struct InvocationData<K> {
...
+        }
+        let exec_units_to_consume = self.kernel.price_list().gas_to_exec_units(gas_used, false);
+        self.gas_available_snapshot = gas_available;
+        self.exec_units_consumed_snapshot += exec_units_to_consume;
```
Stebalien, line 72: We should likely just "set" this to the fuel consumed as reported by wasmtime? Otherwise, we're going to have rounding hell.

fvm/src/call_manager/default.rs
arajasek: Ack.

fvm/src/gas/price_list.rs
arajasek: Adding a comment.

fvm/src/lib.rs
arajasek: Should I just manually do the rounding for floor/ceil then? Or is there some other builtin way to specify round up/down when dividing?

fvm/src/syscalls/bind.rs
arajasek: oops don't write code while falling asleep in a car

fvm/src/syscalls/mod.rs
arajasek: Ye..ee....eeeess.

fvm/src/syscalls/mod.rs
Stebalien: > it means the gas_available somehow increased as execution went on, so I'm p sure it can only be programmer error

Nope. We have negative gas charges in some very special cases.

fvm/src/lib.rs
Stebalien: Hm. That's annoying. Yeah, I think we should do this manually. It looks like the implementation is a bit unstable and will actually be changing in the near future.

fvm/src/kernel/default.rs
Stebalien: AH, no. I mean we need to charge for the amount of data read which might, technically, be slightly different from the size of the block.

fvm/src/kernel/default.rs
arajasek: Discussing this in sync: We're leaning towards having a flat_fee (that includes the extern cost) + a per_byte_fee.

fvm/src/kernel/default.rs
```diff
@@ -295,8 +307,14 @@ where
             // to be in the state-tree.
             .or_fatal()?;
 
-        // We charge on open, not read, to emulate the current gas model.
         let block = Block::new(cid.codec(), data);
+
+        self.call_manager.charge_gas(
+            self.call_manager
+                .price_list()
+                .on_block_open_per_byte(block.size().try_into().or_illegal_argument()?),
```
Stebalien, line 315: You know, this is a case where I'd just cast. The block size limit will be well below this.

fvm/src/kernel/default.rs
```diff
@@ -323,7 +344,7 @@ where
         self.call_manager.charge_gas(
```
Stebalien, line 344: The comment above is out of date.

fvm/src/kernel/default.rs
```diff
@@ -342,18 +363,29 @@ where
...
-            len as u32
-        })
+            let copy_len = buf.len().min(data.len());
+            buf.copy_from_slice(&data[offset as usize..][..copy_len]);
+            copy_len
+        };
+
+        self.call_manager
+            .charge_gas(self.call_manager.price_list().on_block_read_per_byte(len))
+            .map(|_| len as u32)
```
Stebalien, line 382: Unlike `open`, we _should_ charge this up-front. I mean, there's no observable difference, but we save ourselves a bit of work if the gas doesn't cover the copy.

fvm/src/syscalls/bind.rs
```diff
@@ -94,6 +94,38 @@ fn memory_and_data<'a, K: Kernel>(
...
 
+fn charge_exec_units_for_gas(caller: &mut Caller<InvocationData<impl Kernel>>) -> Result<(), Trap> {
+    let exec_units = caller
+        .data_mut()
+        .calculate_exec_units_for_gas()
+        .map_err(|_| Trap::new("failed to calculate exec_units"))?;
+    if exec_units.is_negative() {
+        caller.add_fuel(u64::try_from(exec_units.saturating_neg()).unwrap_or(0))?;
+    } else {
+        caller.consume_fuel(u64::try_from(exec_units).unwrap_or(0))?;
```
Stebalien, line 105: Or max, I think. We want to consume _all_ fuel if exec_units are positive and don't fit, right?

fvm/src/syscalls/bind.rs
```diff
@@ -94,6 +94,38 @@ fn memory_and_data<'a, K: Kernel>(
     Ok((Memory::new(mem), data))
 }
 
+fn charge_exec_units_for_gas(caller: &mut Caller<InvocationData<impl Kernel>>) -> Result<(), Trap> {
+    let exec_units = caller
+        .data_mut()
+        .calculate_exec_units_for_gas()
+        .map_err(|_| Trap::new("failed to calculate exec_units"))?;
+    if exec_units.is_negative() {
+        caller.add_fuel(u64::try_from(exec_units.saturating_neg()).unwrap_or(0))?;
```
Stebalien, line 103: Technically... we want to add infinite fuel in the `unwrap_or` case? That seems like a bad idea. I think we need to turn this into a fatal error somehow?

Stebalien: I mean, it can't happen, but it likely means we wrapped somewhere.

Stebalien (approved):
Some nits, but LGTM.

fvm/src/syscalls/bind.rs
arajasek: I don't think it can not fit, right? If exec_units (and i64) is positive, it _must_ fit in a u64? We should probably just `as u64` it?

fvm/src/syscalls/bind.rs
arajasek: Idk, I'm pretty comfortable saying "do nothing" if something literally impossible like this happens.

Stebalien (approved):

PR #454, opened 2022-04-05
fix(shared): limit BigInt encoding and decoding sizes

dignifiedquire:
The limit is taken from https://github.com/filecoin-project/go-state-types/blob/6027e22b77fd07693bf782401d10215057af9aad/big/int.go#L13

### DIFF
--- a/shared/src/bigint/bigint_ser.rs
+++ b/shared/src/bigint/bigint_ser.rs
@@ -75,7 +75,7 @@ mod tests {
 
     #[test]
     fn test_bigiint_max() {
-        let max_limbs = MAX_BIGINT_SIZE / 4; // u32 -> u8
+        let max_limbs = MAX_BIGINT_SIZE / 4; // 32bit limbs to bytes
         let good = BigInt::new(Sign::Plus, vec![u32::MAX; max_limbs - 1]);
         let good_neg = BigInt::new(Sign::Minus, vec![u32::MAX; max_limbs - 1]);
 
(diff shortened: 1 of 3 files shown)


shared/src/bigint/bigint_ser.rs
```diff
@@ -53,5 +59,58 @@ where
...
+
+#[cfg(test)]
+mod tests {
+    use fvm_ipld_encoding::{from_slice, to_vec};
+
+    use super::*;
+
+    #[test]
+    fn test_bigiint_max() {
+        let max_limbs = MAX_BIGINT_SIZE / 4; // u32 -> u8
```
vmx, line 78: It took me a while to get the comment. Proposal for another comment: "// 32-bit limbs"

shared/src/bigint/bigint_ser.rs
```diff
@@ -53,5 +59,58 @@ where
...
+
+        let good_bytes = to_vec(&BigIntSer(&good)).expect("should be good");
+        let good_back: BigIntDe = from_slice(&good_bytes).unwrap();
+        assert_eq!(good_back.0, good);
+
+        let good_neg_bytes = to_vec(&BigIntSer(&good_neg)).expect("should be good");
+        let good_neg_back: BigIntDe = from_slice(&good_neg_bytes).unwrap();
+        assert_eq!(good_neg_back.0, good_neg);
+
+        let bad1 = BigInt::new(Sign::Plus, vec![u32::MAX; max_limbs]);
```
vmx, line 90: It took me a while to understand why this is a bad `BigInt`. So perhaps it would be good to add a comment that the sign adds an additional byte when serialized (feel free to ignore if you think that's only me).

Stebalien (approved):
Merge when you feel ready.

shared/src/bigint/bigint_ser.rs
dignifiedquire: added a comment

shared/src/bigint/bigint_ser.rs
dignifiedquire: updaetd the comment

vmx (approved):

PR #460, opened 2022-04-05
fix(shared): impl IntoIterator for Address

dignifiedquire:
This allows writing `rt.validate_immediate_caller_is(&*ACTOR_ADDR)`, without having to
manually wrap them.


Stebalien (changes requested):
This exceeds my magic limit. It's not _that_ hard to just write `iter::once` or `[...]`.

dignifiedquire:
fiiiine

PR #461, opened 2022-04-05
Reject system exit codes from abort

### LINKED ISSUE #253: Reject actor abort with a system exit code

anorth:
Fixes #253.

I'm a little confused why I need a network version check here (to have conformance tests pass). I thought the master branch was for nv16 already. Should I instead update the conformance tests? How do I do that?

### DIFF
--- a/fvm/src/syscalls/vm.rs
+++ b/fvm/src/syscalls/vm.rs
@@ -1,4 +1,5 @@
 use fvm_shared::error::ExitCode;
+use fvm_shared::version::NetworkVersion;
 
 use super::error::Abort;
 use super::Context;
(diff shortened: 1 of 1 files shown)


Stebalien:
> I'm a little confused why I need a network version check here (to have conformance tests pass). I thought the master branch was for nv16 already.

Conformance tests are extracted from the network itself, and they're a big part of our test suite at the moment.

Stebalien (approved):

raulk (commented):
Post-merge approve

raulk:
Our current support level is nv15+. Not only for conformance tests, but also to support the experimental FVM mode against today's mainnet. We may drop support for nv15 since those code paths are technically only transitory.

anorth:
I thought that only the v7 branch was expected to sync mainnet. We'll have to drop that compatibility eventually (e.g. gas).

arajasek:
@anorth The plan is to keep that compatibility, even for gas (using network version checks as appropriate). Maintaining v15 compatibility is pretty important to our current testing process (a _lot_ of our confidence comes from continuing to sync the chain).

Having a separate branch for v15 is possible, but gets painful fast.

PR #462, opened 2022-04-06
Map all executor errors to exit codes.

### LINKED ISSUE #438: Default executor should map all syscall error numbers to exit codes
The executor handles some syscall error numbers by propagating an error result, which means there is no receipt to persist to reflect the failure. I.e. a "fatal" error. It should instead specify an exit code for each error number.

anorth:
Fixed #438.

Please review each mapping and consider whether it makes sense. Recall that the context is errors that take place during top-level dispatch.

FYI @dignifiedquire

### DIFF
--- a/fvm/src/executor/default.rs
+++ b/fvm/src/executor/default.rs
@@ -109,21 +109,22 @@ where
                 gas_used,
             },
             Err(ExecutionError::Syscall(err)) => {
-                // Errors indicate the message couldn't be dispatched at all.
-                // Some of these errors are mapped to exit codes that persist on chain.
-                // The remainder propagate in the Result as error numbers and cannot be persisted.
-                // TODO: map them all to exit codes, https://github.com/filecoin-project/ref-fvm/issues/438
+                // Errors indicate the message couldn't be dispatched at all
+                // (as opposed to failing during execution of the receiving actor).
+                // These errors are mapped to exit codes that persist on chain.
                 let exit_code = match err.1 {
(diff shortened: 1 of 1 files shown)


fvm/src/executor/default.rs
```diff
@@ -109,21 +109,22 @@ where
...
                     ErrorNumber::NotFound => ExitCode::SYS_INVALID_RECEIVER,
-                    code => {
-                        return Err(anyhow!(
-                            "unexpected syscall error when processing message: {} ({})",
-                            code,
-                            code as u32
-                        ))
-                    }
+
+                    ErrorNumber::IllegalArgument => ExitCode::SYS_ASSERTION_FAILED,
```
Stebalien, line 119: You don't _have_ to enumerate these, but I guess it doesn't hurt.

Stebalien: Also note: you _can_ combine patterns with `|`. I.e.:

```rust
use ErrorNumber::*;

match err.1 {
    ...
    IllegalOparation | LimitExceeded | ...
}
```

But the result isn't any more readable.

Stebalien (approved):

fvm/src/executor/default.rs
anorth: > You don't have to enumerate these, but I guess it doesn't hurt

Thanks. I really think we should enumerate them, or we'll easily miss mappings that aren't right.

PR #463, opened 2022-04-06
Use concrete error types for ipld/* and shared

dignifiedquire:
This is the first step to be able to remove the usage of downcasting in actors. 
It removes the usage of anyhow in lower level parts.

### DIFF
--- a/fvm/src/blockstore/buffered.rs
+++ b/fvm/src/blockstore/buffered.rs
@@ -1,7 +1,6 @@
 // Copyright 2019-2022 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
 
-use anyhow::{anyhow, Result};
 use byteorder::{BigEndian, ByteOrder, ReadBytesExt};
 use cid::Cid;
 use fvm_ipld_blockstore::{Blockstore, Buffered};
@@ -38,6 +37,14 @@ where
     }
 }
 
+#[derive(thiserror::Error, Debug)]
+pub enum Error<E> {
+    #[error("flush: {0}")]
+    Flush(#[from] FlushError),
+    #[error("blockstore: {0}")]
+    Blockstore(E),
+}
+
 impl<BS> Buffered for BufferedBlockstore<BS>
 where
     BS: Blockstore,
@@ -45,16 +52,44 @@ where
     /// Flushes the buffered cache based on the root node.
     /// This will recursively traverse the cache and write all data connected by links to this
     /// root Cid.
-    fn flush(&self, root: &Cid) -> Result<()> {
+    fn flush(&self, root: &Cid) -> Result<(), Error<BS::Error>> {
         let mut buffer = Vec::new();
         let mut s = self.write.borrow_mut();
         copy_rec(&s, *root, &mut buffer)?;
 
-        self.base.put_many_keyed(buffer)?;
+        self.base
+            .put_many_keyed(buffer)
+            .map_err(Error::Blockstore)?;
         *s = Default::default();
 
         Ok(())
     }
+}
+
+#[derive(thiserror::Error, Debug)]
+pub enum FlushError {
+    #[error("io: {0}")]
+    Io(#[from] std::io::Error),
+    #[error("cid: {0}")]
+    Cid(#[from] cid::Error),
+    #[error("cbor input was not canonical (lval 24 with value < 24)")]
+    HeaderLval24,
+    #[error("cbor input was not canonical (lval 25 with value <= MaxUint8)")]
+    HeaderLval25,
+    #[error("cbor input was not canonical (lval 26 with value <= MaxUint16)")]
+    HeaderLval26,
+    #[error("cbor input was not canonical (lval 27 with value <= MaxUint32)")]
+    HeaderLval27,
+    #[error("invalid header cbor_read_header_buf")]
+    HeaderInvalid,
+    #[error("expected cbor type byte string in input")]
+    UnexpectedByteString,
+    #[error("string in cbor input too long")]
+    StringTooLong,
+    #[error("Invalid link ({0}) in flushing buffered store")]
+    InvalidLink(Cid),
+    #[error("unhandled cbor type: {0}")]
+    UnhandledCborType(u8),
 }
 
 /// Given a CBOR encoded Buffer, returns a tuple of:
@@ -64,7 +99,10 @@ where
 /// This was implemented because the CBOR library we use does not expose low
 /// methods like this, requiring us to deserialize the whole CBOR payload, which
 /// is unnecessary and quite inefficient for our usecase here.
-fn cbor_read_header_buf<B: Read>(br: &mut B, scratch: &mut [u8]) -> anyhow::Result<(u8, usize)> {
+fn cbor_read_header_buf<B: Read>(
+    br: &mut B,
+    scratch: &mut [u8],
+) -> Result<(u8, usize), FlushError> {
     let first = br.read_u8()?;
     let maj = (first & 0xe0) >> 5;
     let low = first & 0x1f;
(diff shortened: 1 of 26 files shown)


fvm/src/blockstore/buffered.rs
```diff
@@ -38,33 +37,72 @@ where
...
+    #[error("cid: {0}")]
+    Cid(#[from] cid::Error),
+    #[error("cbor input was not canonical (lval 24 with value < 24)")]
+    HeaderLval24,
+    #[error("cbor input was not canonical (lval 25 with value <= MaxUint8)")]
+    HeaderLval25,
+    #[error("cbor input was not canonical (lval 26 with value <= MaxUint16)")]
+    HeaderLval26,
+    #[error("cbor input was not canonical (lval 27 with value <= MaxUint32)")]
+    HeaderLval27,
```
vmx, line 82: Without knowing what they are really about, I'd collapse them into `HeaderLval(u8)`.

Stebalien:
We switched to anyhow because:

1. These errors aren't useful in practice. We can't actually _do_ anything other than to give the user a nice error message telling them what's wrong.
2. Converting between and handling errors became a _real_ pain.

dignifiedquire:
@Stebalien well they are very useful, because you don't need to downcast anymore, you can actually to type level checks in actors with this

Stebalien:
> @Stebalien well I would argue they are very useful, because you don't need to downcast anymore, you can actually do type level checks in actors with this

Sorry, I should have been specific. I meant having many detailed errors rather than just a simple "here's what went wrong" error message.

On the other hand, they don't doesn't really _hurt_, I guess.

But please propagate before merging here. I agree with you in theory, but enumerated errors became a _real_ pain in practice.

- Anyhow allowed us to just combine arbitrary errors. This makes propagating errors _really_ easy.
- Error enums means you need to either:
    - Define a different enum for every case.
    - Define a single enum that includes every sub-error in your crate, which is pretty useless (no way to tell which errors can _actually_ happen in any given case).
- Making the blockstore generic over the error tainted _everything_ with additional generics.

On the other hand, we started with an unholy mix of enumerated errors, string errors, and anyhow. You may have better luck now.

anorth (approved):
Looks great to me, but please wait for someone else's review too.

ipld/amt/src/error.rs
```diff
@@ -20,49 +17,65 @@ pub enum Error {
...
-impl From<&'static str> for Error {
-    fn from(e: &'static str) -> Self {
-        Self::Dynamic(anyhow::anyhow!(e))
-    }
+#[derive(Debug, Error)]
+pub enum EitherError<U, E> {
+    #[error("user: {0}")]
+    User(U),
+    #[error("hamt: {0}")]
+    Hamt(#[from] Error<E>),
```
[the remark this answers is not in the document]
dignifiedquire: that’s what I get for copy pasting 😬

dignifiedquire:
once the exit code lands on builtin-actors, I will make a matchinh PR, to make sure things work as intended based on this.

ipld/amt/src/error.rs
```diff
@@ -20,49 +17,65 @@ pub enum Error {
...
+        }
     }
 }
 
-impl From<&'static str> for Error {
-    fn from(e: &'static str) -> Self {
-        Self::Dynamic(anyhow::anyhow!(e))
-    }
+#[derive(Debug, Error)]
+pub enum EitherError<U, E> {
```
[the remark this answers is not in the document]
dignifiedquire: added some docs

ipld/amt/src/amt.rs
```diff
@@ -243,23 +246,23 @@ where
...
+    ) -> Result<bool, Error<BS::Error>> {
         // TODO: optimize this
         let mut modified = false;
 
         // Iterate sorted indices. Sorted to safely optimize later.
         for i in sorted(iter) {
             let found = self.delete(i)?.is_none();
             if strict && found {
-                return Err(anyhow!("no such index {} in Amt for batch delete", i).into());
+                return Err(Error::BatchDelteNotFound(i));
```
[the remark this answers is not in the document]
dignifiedquire: fixed

fvm/src/blockstore/buffered.rs
```diff
@@ -38,33 +37,66 @@ where
...
+    #[error("cbor input was not canonical (lval {0} with value < {1})")]
+    HeaderNotCanonical(usize, &'static str),
+    #[error("invalid header cbor_read_header_buf")]
+    HeaderInvalid,
+    #[error("expected cbor type byte string in input")]
+    UnexpectedByteString,
+    #[error("string in cbor input too long")]
+    StringTooLong,
+    #[error("Invalid link ({0}) in flushing buffered store")]
+    InvalidLink(Cid),
```
Stebalien, line 84: MissingBlock?

fvm/src/blockstore/buffered.rs
```diff
@@ -38,33 +37,66 @@ where
...
+            .put_many_keyed(buffer)
+            .map_err(Error::Blockstore)?;
         *s = Default::default();
 
         Ok(())
     }
 }
 
+#[derive(thiserror::Error, Debug)]
+pub enum FlushError {
```
Stebalien, line 70: I'd reduce this to:

- MissingBlock (can't find a block)
- IPLD (encoding error)

(how can we even have an IO error that's not a blockstore error?)

If we're super-specific, we'll have to break this error type every time we change some detail (switch to libipld's link enumeration function, support new IPLD codecs, etc.).

ipld/blockstore/src/lib.rs
```diff
@@ -15,26 +14,28 @@ pub use block::*;
 ///
 /// The cgo blockstore adapter implements this trait.
 pub trait Blockstore {
+    type Error: std::error::Error + std::fmt::Debug + Send + Sync + 'static;
```
Stebalien, line 17: This is what we started with, but it was a _huge_ pain.

1. This is incompatible with anyhow, because anyhow doesn't implement `std::error::Error`.
2. Repeating this kind of constraint wherever we want to be generic is _really_ annoying.

What if we just _don't_ constrain this? Really, only _applications_ care about enforcing constraints on this type.

ipld/hamt/src/pointer.rs
```diff
@@ -111,7 +112,7 @@ where
 
     /// Internal method to cleanup children, to ensure consistent tree representation
     /// after deletes.
-    pub(crate) fn clean(&mut self) -> Result<(), Error> {
+    pub(crate) fn clean<BS: Blockstore>(&mut self) -> Result<(), Error<BS::Error>> {
```
Stebalien, line 115: Can you use `!` or `convert::Infallable` as the blockstore error?

fvm/src/blockstore/buffered.rs
dignifiedquire: fixed

ipld/blockstore/src/lib.rs
dignifiedquire: That is true, the main reason I constrained it is because `anyhow::Context` requires this, and so it gets painful there, as I can't easily define a `where` clause on `Machine::Blockstore::Error` without having to write it in a __LOT__ of places.

ipld/hamt/src/pointer.rs
dignifiedquire: If I do that, the error propagation stops working when I call it. If you have an idea how to work around it, I am all ears

ipld/blockstore/src/lib.rs
Stebalien: Yeah, I guess that's fine. We just need to make sure we have patches for the actors before we merge.

ipld/hamt/src/pointer.rs
Stebalien: Oh. I see.

maciejwitowski:
@Stebalien This is assigned to M2.1 now. Is it correct?

Stebalien:
It probably won't land in M2.1. I'd like it to, but I expect we'll handle this in M2.2.

PR #464, opened 2022-04-06
Derive NetworkVersion from strings

connormullett:
Currently the Forest project would like functionality to allow the ability of Deserializing a NetworkVersion from a String rather than a `u32`. This will allow us to increase our type safety and result in less panics at runtime when our chain configuration is ran at the start of the node.

This PR adds the functionality stated above.


connormullett:
> Have you considered using `serde(with)`? Or wrapping the type in APIs?

It's been considered to use a wrapper, however we came to the conclusion that it would best be with the original type. I understand the need to make sure the types deserialize/serialize symmetrically, there was no prior deserialization implementation. What would you recommend to solve this?

Stebalien:
Can you link to the affected code? A `#[serde(with)]` specification is likely the _simplest_ solution.

Stebalien:
@connormullett any updates?

PR #465, opened 2022-04-06
feat: measure fuel usage and timing stats

Kubuxu:


raulk:
Rude to remove me as an author @Kubuxu 🙃🤣

Kubuxu:
It wasn't on purpose 🙃 no idea why it picked up steb instead of you as the author.

Kubuxu:
Fixed

raulk (commented):
This tracing framework is useful to inspect running logic in detail, how it interacts with the environment, debugging execution flow, understand the instruction complexity of ranges, and get a _relative_ sense of where time is spent.

However, this should come with a massive warning sign: the timing numbers are considerably skewed by several factors:

1. Non-negligible overhead of getting time from the system.
    - According to [minstant benchmarks](https://crates.io/crates/minstant), `std::time::Instant::now()` can have an overhead of as much as 30ns on the specified platform.
    - Using the TSC is a solution, but even with minstant there's an overhead of 10ns which needs to be accounted for.
    - Potential solution: correct for overhead by benchmarking getting time, and applying a negative offset to all duration readings.
2. Additional instructions + cache thrashing due to the tracing logic, which is not observed in real executions.

PR #467, opened 2022-04-06
chore: run github actions on pull requests from forks

Stebalien:


galargh (approved):

vmx:
I highly prefer running CI on branches before I open a PR. I want to reduce noise for other folks. I usually push on a branch and only PR once it is green. This workflow doesn't work anymore.

galargh:
Myself, I usually use draft PRs to handle in-progress work.

In a lot of PL Go repositories, we run CI on both all `push`es and `pull_request`s. You could do that here too but, unfortunately, since we don't have a way to monitor GitHub Actions usage yet, it'll be hard to tell what it means in terms of runners availability. You can always check in practice if such setup affects you negatively.

Issue #373, opened 2022-03-09
Reduce monomorphization by using shared, non-generic helper functions


raulk:
So only the outer generic function gets monomorphized.

raulk:
@Stebalien might explaining this a little bit more and providing links to places where we'd need to refactor?

Stebalien:
Rust will create a new "copy" of every generic function for every type it's called with. To reduce the amount of code, one can:

1. Convert the generic parameter into some common type.
2. Call a non-generic helper function, moving most of the code into that function.

That way only the _generic_ part gets copied.

To find code bloat (especially from monomorphizations), use `twiggy`. Specifically, `twiggy monos my_actor.wasm`.


For example, the miner actor:

```
 Apprx. Bloat Bytes │ Apprx. Bloat % │ Bytes   │ %      │ Monomorphizations
────────────────────┼────────────────┼─────────┼────────┼───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
             529243 ┊         10.53% ┊  535778 ┊ 10.66% ┊ cs_serde_cbor::de::Deserializer<R>::parse_value
                    ┊                ┊    6535 ┊  0.13% ┊     cs_serde_cbor::de::Deserializer<R>::parse_value::hb4eea98bd31e9c74
                    ┊                ┊    6474 ┊  0.13% ┊     cs_serde_cbor::de::Deserializer<R>::parse_value::h149b2147018c4e8f
                    ┊                ┊    6418 ┊  0.13% ┊     cs_serde_cbor::de::Deserializer<R>::parse_value::ha76c2cb4ae0d0e03
                    ┊                ┊    6398 ┊  0.13% ┊     cs_serde_cbor::de::Deserializer<R>::parse_value::h84eab51809962d91
                    ┊                ┊    6318 ┊  0.13% ┊     cs_serde_cbor::de::Deserializer<R>::parse_value::h00a48fe07203d3ca
                    ┊                ┊    6318 ┊  0.13% ┊     cs_serde_cbor::de::Deserializer<R>::parse_value::h0208eea32461d279
                    ┊                ┊    6318 ┊  0.13% ┊     cs_serde_cbor::de::Deserializer<R>::parse_value::h076c98e03f124303
                    ┊                ┊    6318 ┊  0.13% ┊     cs_serde_cbor::de::Deserializer<R>::parse_value::h082cfe60d482e5e9
[...]

PR #472, opened 2022-04-08
ipld/encoding: switch to cbor4ii

vmx:
TODOs:

 - [x] Do not fork `cbor4ii`, but use it's core features and just have our own Serde implementation
 - [x] Integrate that work into `serde_ipld_dagcbor` and release it as 0.2.

### DIFF
--- a/ipld/blockstore/Cargo.toml
+++ b/ipld/blockstore/Cargo.toml
@@ -10,10 +10,9 @@ repository = "https://github.com/filecoin-project/ref-fvm"
 [dependencies]
 cid = { version = "0.8.2", default-features = false, features = ["serde-codec", "std"] }
 anyhow = "1.0.51"
-
-
-[dev-dependencies]
-multihash = { version = "0.16.1", default-features = false, features = ["blake2b", "multihash-impl"] }
+# multihash is also re-exported by `cid`. Having `multihash` here as a
+# depdendency is needed to enable the features of the re-export.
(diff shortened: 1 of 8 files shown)


Stebalien:
We also need to enforce minimal encoding.

vmx:
> We also need to enforce minimal encoding.

What do you mean with "minimal encoding". That integers are in the smallest representation?

vmx:
That's already the case, it should be covered by the [IPLD codec fixtures](https://github.com/ipld/codec-fixtures):

raulk:
Addresses https://github.com/filecoin-project/ref-fvm/issues/373 to a fair degree.

ipld/blockstore/Cargo.toml
```diff
@@ -10,9 +10,8 @@ repository = "https://github.com/filecoin-project/ref-fvm"
 [dependencies]
 cid = { version = "0.8.2", default-features = false, features = ["serde-codec", "std"] }
 anyhow = "1.0.51"
-
-
-[dev-dependencies]
+# multihash is also re-exported by `cid`. Having `multihash` here as a
+# depdendency is needed to enable the features of the re-export.
 multihash = { version = "0.16.1", default-features = false, features = ["blake2b", "multihash-impl"] }
```
Stebalien, line 15: We don't want to enable blake2b by default as a non-dev dependency.

Stebalien (approved):

PR #476, opened 2022-04-11
Refactor consume/take

Stebalien:
This refactor makes two significant changes:

1. Renames consume/take into `into_*` less ambiguous `into_*` functions.
2. Boxes the "inner" call manager to reduce the amount of copying as we move up and down the callstack.

### DIFF
--- a/fvm/src/blockstore/buffered.rs
+++ b/fvm/src/blockstore/buffered.rs
@@ -33,7 +33,7 @@ where
         }
     }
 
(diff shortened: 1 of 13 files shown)


Stebalien:
Note: this isn't just a spur of the moment "I feel like refactoring things". It's just something that has been on my mind, but the time was never right.

raulk (approved):
WFM. This is more idiomatic, although the consume/take terminology is more straight-to-the-point.

PR #481, opened 2022-04-11
fvm_shared: fix RegisteredSealProof incorrect order/values for 8MiB and 512MiB.

raulk:
Now matching go-state-types: https://github.com/filecoin-project/go-state-types/blob/master/abi/sector.go#L71-L83.

These have not caused a problem so far because we haven't upgraded any live network that uses these sector sizes to the FVM.


anorth (approved):

magik6k (commented):
This doesn't work because of the `Invalid(i64)` in the enum (I think)

PR #484, opened 2022-04-11
doc: document syscalls & errors

Stebalien:
I've decided to document these in the SDK because this is the cleanest, low-level API definition we have. My plan is to link to the documentation from the spec instead of duplicating the syscall definitions in both places.

Rendered: https://bafybeif5utc6rkqubx3iiikkpn6ubompnd7xdowliznndocyaqgozy2uni.ipfs.dweb.link/fvm_sdk/sys/index.html

### DIFF
--- a/sdk/src/sys/actor.rs
+++ b/sdk/src/sys/actor.rs
@@ -1,25 +1,40 @@
+//! Syscalls for creating and resolving actors.
+
+#[doc(inline)]
+pub use fvm_shared::sys::out::actor::*;
+
 super::fvm_syscalls! {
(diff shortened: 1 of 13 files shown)


anorth (approved):
Documenting here looks good. I didn't review the sets of errors described for each syscall here in detail (I assume copied faithfully from spec).

Stebalien:
> I assume copied faithfully from spec

Mostly. I cleaned up some descriptions (mostly removed trailing periods).

I also fixed https://github.com/filecoin-project/fvm-specs/pull/83.

Issue #444, opened 2022-04-04
prototype minimal Wasm actor deployment flow


raulk:
## Context

Having a minimal, experimental protoype of an actor deployment flow will unlock several downstream community projects. What's described here is most certainly not the definitive flow, but it's doubly useful:

- It's a sufficiently functional PoC to allow FVM devs to spin up devnets, and node operators to provide testnets, where actors can be deployed.
- Can be regarded like a spike to explore the problem and dissipate unknowns.

This implementation should sit in a branch, ideally behind a Cargo feature flag, and should carry prominent EXPERIMENTAL warnings.

## InitActor enhancements

See https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0030.md#actor-deployment for reference.

### State object

Add an `actor_registry: Hamt<Cid, ActorSpec>` field to the state object, where:

```rs
struct ActorSpec {
    // The actor that installed this actor.
    creator: ActorID,
    // The originator of the message that triggered creator to install this actor.
    // Probably needs https://github.com/filecoin-project/ref-fvm/issues/300
    originator: ActorID,
    // The version of the invocation container.
    container_version: u32,
}
```

### Operations

Add an `InstallActor(bytecode: RawBytes) -> Result<InstallActorReturn, ActorError>` method, where:

```rs
struct InstallActorReturn {
    // The CodeCID with which this actor can be constructed.
    code_cid: Cid,
    // Whether the bytecode was effectively installed in this call or not.
    // False when the CodeCID already existed.
    // Alternatively we could fail, but that would break API idempotency.
    installed: bool,
}
```

This method:

1. Computes the Cid of the parameter block (Wasm bytecode).
2. Checks if an entry already exists in`actor_registry`. If yes, it shortcircuits and returns the CodeCID and `installed=false`.
3. Adds the bytecode to the blockstore.
4. Calls the `actor::install_actor(block_id) -> Result<Cid>` syscall to apply the installation. Aborts on failure.
5. Adds an entry in `actor_registry`, filling in the active container version, creator, and originator.
6. Returns the CodeCID and `installed=true`

### `actor::install_actor` syscall

1. Persists the Wasm bytecode to the blockstore, and remembers the resulting Cid as the `code_cid`.
2. Charge the actor installation gas defined in https://github.com/filecoin-project/FIPs/pull/317. (Can be 0 for now)
[...]

PR #485, opened 2022-04-12
prototype for installation of user actors (including `install_actor` syscall)

vyzo:
This introduces the ability to deploy custom actors, in the manner specified in https://github.com/filecoin-project/ref-fvm/issues/444. Concretely:

- It adds an `install_actor` syscall that's only wired into the actor container if, and only if, the `m2-native` Cargo feature is enabled. This syscall won't be necessary for M2 EVM, that's why it's behind a feature flag.
- It adds the `IPLD_RAW` (0x55) codec to the allowlist for the IPLD subsystem.
- This still keeps compiled modules in memory.

Requires https://github.com/filecoin-project/builtin-actors/pull/216 to actually be useful.


raulk (commented):
LGTM so far.

vyzo:
rebased on master.

vyzo:
Converted back to draft as #658 is a clean rebased form that has feature gating.

raulk:
Superseded by #658.

PR #487, opened 2022-04-13
Add a base and per-byte cost for getting randomness

arajasek:


Stebalien (approved):
LGTM, needs a FIP update.

Stebalien:
So, actually, these are already TODOs in the FIP. We _do_ need to update the FIP, but the changes here don't change anything.

PR #489, opened 2022-04-13
chore: sector: Cleanup registered_proof

magik6k:
builtin-actors build with this, but I'm not 100% sure that this isn't a breaking change

Related to https://github.com/filecoin-project/ref-fvm/pull/481


Stebalien:
This may be dead code, but it doesn't really hurt anything. Closing to reduce breaking changes at this stage.

PR #490, opened 2022-04-13
refactor: combine/refactor context/config/machine arguments

### LINKED ISSUE #249: Simplify machine constructor
Mainly, it takes too many arguments. Many of these (epoch, base fee, state root, etc.) should be moved to a new struct describing the current Filecoin state.

Stebalien:
1. Reduce the number of arguments to the constructor.
2. Renames "debug" to actor_debugging to make it clear what it does, and document that it's
   consensus critical.
3. Split the config into "network-wide" and "epoch-specific" sections.

This change also makes it clear that these options are all consensus critical.

fixes #249

### DIFF
--- a/fvm/src/machine/mod.rs
+++ b/fvm/src/machine/mod.rs
@@ -1,4 +1,5 @@
 use cid::Cid;
+use derive_more::{Deref, DerefMut};
 use fvm_ipld_blockstore::Blockstore;
 use fvm_shared::actor::builtin::Manifest;
 use fvm_shared::address::Address;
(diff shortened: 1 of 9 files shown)


fvm/src/machine/mod.rs
```diff
@@ -88,23 +86,110 @@ pub trait Machine: 'static {
...
+    }
+}
+
+/// Per-epoch machine context.
+#[derive(Clone, Debug, Deref, DerefMut)]
 pub struct MachineContext {
+    /// Network-level settings.
+    #[deref]
+    #[deref_mut]
+    pub network: NetworkConfig,
```
Stebalien, line 162: It's unclear if this should go inside the context (my first attempt had it as a second machine parameter), but it's convenient to specify all "context" information in the same place.

PR #494, opened 2022-04-14
feat: fvm-wasm-instrument stack limiter and gas

magik6k:


Stebalien:
This looks like the correct approach, but do we need to do any pre-validation first?

magik6k:
Depends on https://github.com/filecoin-project/fvm-wasm-instrument/pull/3

fvm/src/gas/mod.rs
```diff
@@ -10,34 +10,46 @@ mod charge;
 mod outputs;
 mod price_list;
 
+/// bits of precision for fractional gas
+/// 16 bits means 65_536 fractional gas per unit of gas
+pub const FRGAS_PRECISION: i64 = 16;
+
 pub struct GasTracker {
-    gas_available: i64,
+    // TODO: convert to frgas
```
magik6k, line 18: Separate PR, or do we want to do this here?

fvm/src/syscalls/bind.rs
```diff
@@ -95,34 +96,35 @@ fn memory_and_data<'a, K: Kernel>(
...
+
+    let frgas = match global.unwrap().get(&mut *caller) {
+        Val::I64(g) => Ok(g),
+        _ => Err(Trap::new("failed to get wasm gas")),
+    }?;
+
+    let available_gas = gas::frgas_to_gas(frgas, false);
 
+    // todo do we have consts for charge names anywhere?
+    // todo make sure that we handle traps/out-of-gas here correctly
```
magik6k, line 121: I'm 60% convinced that we're not doing this correctly

fvm/src/gas/mod.rs
```diff
@@ -58,6 +104,24 @@ impl GasTracker {
...
+#[inline]
+pub fn frgas_to_gas(frgas: i64, round_up: bool) -> i64 {
+    let p = 1 << FRGAS_PRECISION;
+
+    let mut div_result = frgas / p;
+    if round_up && frgas % p != 0 {
+        div_result = div_result.saturating_add(1);
+    }
+    div_result
+}
```
Stebalien, line 123: Instead of providing conversion methods, I'd consider adding methods to the gas tracker:

- `charge_milligas`, `available_milligas`, etc.

fvm/src/call_manager/default.rs
```diff
@@ -357,7 +343,14 @@ where
                 .or_fatal()
             {
                 Ok(ret) => ret,
-                Err(err) => return (Err(err), store.into_data().kernel.into_call_manager()),
+                Err(err) => {
+                    store
+                        .data_mut()
+                        .kernel
+                        .set_available_miligas("getinstance_fail", initial_miligas)
+                        .unwrap();
```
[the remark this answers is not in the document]
Stebalien: Ah, because we're "returning" the gas?

fvm/src/machine/engine.rs
```diff
@@ -125,26 +128,67 @@ impl Engine {
...
         Ok(())
     }
 
     /// Load some wasm code into the engine.
-    pub fn load_bytecode(&self, k: &Cid, wasm: &[u8]) -> anyhow::Result<Module> {
+    pub fn load_bytecode(
+        &self,
+        k: &Cid,
+        wasm: &[u8],
+        mctx: &MachineContext,
```
Stebalien, line 142: So, this engine was supposed to be independent of machine, epoch, pricelist, etc. We're probably going to need to move to a model where we construct a `NetworkConfig` (contains the network version, pricelist, etc.).

fvm/src/syscalls/bind.rs
```diff
@@ -95,34 +95,31 @@ fn memory_and_data<'a, K: Kernel>(
...
-        .ok_or_else(|| Trap::new("expected to find exec_units consumed"))?;
+fn wasmgas_to_gastracker(caller: &mut Caller<InvocationData<impl Kernel>>) -> Result<(), Trap> {
+    let global = caller.data_mut().avail_gas_global;
 
+    let miligas = match global.unwrap().get(&mut *caller) {
+        Val::I64(g) => Ok(g),
+        _ => Err(Trap::new("failed to get wasm gas")),
+    }?;
+
+    // todo do we have consts for charge names anywhere?
```
Stebalien, line 115: no, they're just for debugging anyways.

fvm/src/syscalls/bind.rs
Stebalien: We're not. If we're out of gas, we need to return an OutOfGas error. We can do that by converting an OutOfGas `Abort` into a `Trap`.

Stebalien (commented):
partial review

fvm/src/call_manager/default.rs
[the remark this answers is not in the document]
magik6k: Correct

fvm/src/gas/price_list.rs
```diff
@@ -257,6 +259,9 @@ lazy_static! {
         block_link_base: 1,
         // TODO: PARAM_FINISH
         block_stat: 1,
+
+        exec_instruction_cost_fr: (gas::MILIGAS_PRECISION / 2) as u64,
```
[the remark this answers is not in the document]
magik6k: Just a weird way to write 500

fvm/src/kernel/mod.rs
```diff
@@ -218,6 +218,7 @@ pub trait CircSupplyOps {
 ///  some built-in actors currently charge gas explicitly for concrete actions.
 ///  In the future (Phase 1), this should disappear and be replaced by gas instrumentation
 ///  at the WASM level.
+/// TODO remove the todo above
```
Stebalien, line 221: yes?

fvm/src/machine/engine.rs
```diff
@@ -101,15 +98,19 @@ impl From<wasmtime::Engine> for Engine {
 
 struct Cache<K> {
     linker: wasmtime::Linker<InvocationData<K>>,
-    instances: HashMap<Cid, wasmtime::InstancePre<InvocationData<K>>>,
```
Stebalien, line 104: This will add a ~70-200k gas overhead. But, well, we _can't_ feasibly cache instances in M2 anyways unless we find some way to persist said cache.

fvm/src/machine/engine.rs
```diff
@@ -125,26 +126,66 @@ impl Engine {
...
         Ok(())
     }
 
     /// Load some wasm code into the engine.
-    pub fn load_bytecode(&self, k: &Cid, wasm: &[u8]) -> anyhow::Result<Module> {
+    pub fn load_bytecode(
+        &self,
+        k: &Cid,
+        wasm: &[u8],
+        nc: &NetworkConfig,
```
Stebalien, line 140: We still need a per-network engine.

magik6k:
Related ffi PR: https://github.com/filecoin-project/filecoin-ffi/pull/273

fvm/src/syscalls/bind.rs
```diff
@@ -148,30 +153,35 @@ macro_rules! impl_bind_syscalls {
...
-                        charge_gas_for_exec_units(&mut caller)?;
+                        wasmgas_to_gastracker(&mut caller)?;
+
                         let (mut memory, mut data) = memory_and_data(&mut caller)?;
                         let ctx = Context{kernel: &mut data.kernel, memory: &mut memory};
-                        let result = match syscall(ctx $(, $t)*).into()? {
-                            Ok(_) => {
+                        let out = syscall(ctx $(, $t)*).into();
+
+                        let result = match out {
```
Stebalien, line 162: I think we can re-combine these lines now.

fvm/src/syscalls/bind.rs
```diff
@@ -148,30 +153,35 @@ macro_rules! impl_bind_syscalls {
...
-                            Err(err) => {
+                            Ok(Err(err)) => {
                                 let code = err.1;
                                 log::trace!("syscall {}::{}: fail ({})", module, name, code as u32);
                                 data.last_error = Some(backtrace::Cause::new(module, name, err));
-                                code as u32
+                                Ok(code as u32)
                             },
+                            Err(e) => Err(e.into()),
                         };
```
Stebalien, line 175: I would _consider_ wrapping all of this in a closure (and immediately calling it) to make trapping errors a bit easier.

testing/integration/src/builtin.rs
```diff
@@ -16,9 +16,10 @@ use crate::error::Error::{
     FailedToLoadManifest, FailedToSetActor, FailedToSetState, MultipleRootCid, NoCidInManifest,
 };
 
-const BUNDLES: [(NetworkVersion, &[u8]); 2] = [
+const BUNDLES: [(NetworkVersion, &[u8]); 3] = [
     (NetworkVersion::V14, actors_v6::BUNDLE_CAR),
     (NetworkVersion::V15, actors_v7::BUNDLE_CAR),
+    (NetworkVersion::V16, actors_v7::BUNDLE_CAR), // todo bad hack
```
Stebalien, line 22: ?

Stebalien (approved):
nits, but the logic LGTM.

fvm/src/gas/price_list.rs
```diff
@@ -648,3 +634,14 @@ pub fn price_list_by_network_version(network_version: NetworkVersion) -> &'stati
         _ => &SKYR_PRICES,
     }
 }
+
+impl Rules for &WasmGasPrices {
+    fn instruction_cost(&self, _instruction: &Instruction) -> Option<u64> {
+        Some(self.exec_instruction_cost_milli)
```
raulk, line 640: There are a few wasm instructions we don't charge for: https://github.com/filecoin-project/FIPs/pull/368/files#diff-399066a5694eb92e4ac1648a75e6e683f9fc26afce520e5b5e17d6131cf695d5R109-R110

Should this handle the filtering?

fvm/src/kernel/mod.rs
```diff
@@ -226,6 +221,12 @@ pub trait GasOps {
     /// `name` provides information about gas charging point
     fn charge_gas(&mut self, name: &str, compute: i64) -> Result<()>;
 
+    /// Returns available gas.
+    fn borrow_milligas(&mut self) -> Result<i64>;
+
+    /// Sets available gas to a new value, creating a gas charge if needed
+    fn return_milligas(&mut self, name: &str, newgas: i64) -> Result<()>;
+
```
raulk, line 229: These traits usually map 1:1 with syscall namespaces. Can you move these up to the Kernel trait itself? Actually: do they even _need_ to be part of the trait? Can't they just be implemented in DefaultKernel?

fvm/src/machine/engine.rs
```diff
@@ -1,25 +1,72 @@
...
+use crate::machine::NetworkConfig;
 use crate::syscalls::{bind_syscalls, InvocationData};
 use crate::Kernel;
 
 /// A caching wasmtime engine.
 #[derive(Clone)]
 pub struct Engine(Arc<EngineInner>);
 
+#[derive(Clone)]
+pub struct MultiEngine(Arc<Mutex<HashMap<EngineConfig, Engine>>>);
```
raulk, line 22: Docs, please!

fvm/src/machine/engine.rs
```diff
@@ -55,23 +104,20 @@ pub fn default_wasmtime_config() -> wasmtime::Config {
     // > not enabled by default.
     c.cranelift_nan_canonicalization(true);
 
-    // c.cranelift_opt_level(Speed); ?
+    // Set to something much higher than the instrumented limiter.
+    c.max_wasm_stack(64 << 20).unwrap();
 
-    c.consume_fuel(true);
```
raulk, line 60: We should probably keep this, set it to false, and add a note.

fvm/src/machine/engine.rs
```diff
@@ -55,23 +104,20 @@ pub fn default_wasmtime_config() -> wasmtime::Config {
     // > not enabled by default.
     c.cranelift_nan_canonicalization(true);
 
-    // c.cranelift_opt_level(Speed); ?
+    // Set to something much higher than the instrumented limiter.
+    c.max_wasm_stack(64 << 20).unwrap();
```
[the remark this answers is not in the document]
raulk: Is this just a backstop? Why do we need it? I guess because it can't be disabled entirely?

fvm/src/machine/engine.rs
```diff
@@ -55,23 +104,20 @@ pub fn default_wasmtime_config() -> wasmtime::Config {
...
 
-impl Default for Engine {
-    fn default() -> Self {
-        Engine::new(&default_wasmtime_config()).unwrap()
-    }
-}
-
 struct EngineInner {
     engine: wasmtime::Engine,
+    dummy_gas_global: Global,
```
raulk, line 117: Docs, pls.

fvm/src/machine/engine.rs
```diff
@@ -83,25 +129,30 @@ impl Deref for Engine {
...
-        Engine(Arc::new(EngineInner {
+    /// Create a new Engine from a wasmtime config.
+    pub fn new(c: &wasmtime::Config, ec: EngineConfig) -> anyhow::Result<Self> {
+        let engine = wasmtime::Engine::new(c)?;
+
+        let mut dummy_store = wasmtime::Store::new(&engine, ());
+        let gg_type = GlobalType::new(ValType::I64, Mutability::Var);
+        let dummy_gg = Global::new(&mut dummy_store, gg_type, Val::I64(0))
+            .expect("failed to create dummy gas global");
+
```
raulk, line 144: Now that we're doing even more at engine init, we should really be pooling these engines.

fvm/src/machine/engine.rs
```diff
@@ -183,35 +270,42 @@ impl Engine {
         let mut instance_cache = self.0.instance_cache.lock().expect("cache poisoned");
 
         let cache = match instance_cache.entry() {
-            anymap::Entry::Occupied(e) => e.into_mut(),
+            anymap::Entry::Occupied(e) => e.into_mut(), // todo what about gas here?
             anymap::Entry::Vacant(e) => e.insert({
                 let mut linker = Linker::new(&self.0.engine);
+                linker.allow_shadowing(true);
```
raulk, line 276: Why is this necessary? Mind adding a note?

fvm/src/machine/mod.rs
```diff
@@ -98,6 +98,10 @@ pub struct NetworkConfig {
     /// DEFAULT: 4096
     pub max_call_depth: u32,
 
+    /// The maximum nubmber of elements on wasm stack
```
raulk, line 101:
```suggestion
    /// The maximum number of elements on wasm stack
```

fvm/src/machine/mod.rs
```diff
@@ -121,6 +125,7 @@ impl NetworkConfig {
         NetworkConfig {
             network_version,
             max_call_depth: 4096,
+            max_wasm_stack: 64 * 1024,
```
raulk, line 128: Are we sure this is enough for built-in actors?

fvm/src/syscalls/bind.rs
raulk: (We'll split them again if we merge #465.)

testing/integration/src/builtin.rs
raulk: I guess this is because we still don't have a release for v8.

fvm/src/gas/mod.rs
```diff
@@ -3,59 +3,130 @@
...
+
+    /// returns available milligas; makes the gas tracker block gas charges until
+    /// set_available_gas is called
+    pub fn borrow_milligas(&mut self) -> Result<i64> {
+        if !self.own_limit {
+            return Err(ExecutionError::Fatal(anyhow::Error::msg(
+                "borrow_milligas called on GasTracker which doesn't own gas limit",
+            )));
+        }
+        self.own_limit = false;
```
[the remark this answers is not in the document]
Stebalien: Unfortunately, no. But we could probably avoid this dance by remembering the starting gas in the `InvocationContext`?

fvm/src/gas/mod.rs
```diff
@@ -3,59 +3,130 @@
...
     /// Safely consumes gas and returns an out of gas error if there is not sufficient
     /// enough gas remaining for charge.
-    pub fn charge_gas(&mut self, charge: GasCharge) -> Result<()> {
-        let to_use = charge.total();
-        match self.gas_used.checked_add(to_use) {
+    fn charge_milligas(&mut self, name: &str, to_use: i64) -> Result<()> {
+        if !self.own_limit {
+            return Err(ExecutionError::Fatal(anyhow::Error::msg(
+                "charge_gas called when gas_limit owned by execution",
+            )));
```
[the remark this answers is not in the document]
Stebalien: `anyhow!("foo").or_fatal()`

fvm/src/kernel/mod.rs
Stebalien: - They need to be a part of the trait because we call them in the CallManager and InvocationContext.
- IMO, splitting gas logic into two places would be more confusing (although I wish we didn't need these functions here).

fvm/src/machine/engine.rs
Stebalien: We pool them in the FFI (using the `MultiEngine` as a "pool").

fvm/src/gas/mod.rs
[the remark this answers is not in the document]
magik6k: I wish we could, but I don't see a great way to do this, especially with the dance that happens in syscall handlers.

fvm/src/gas/price_list.rs
magik6k: Need to look into this more, but those instructions are 'structural', so we may already not be charging for those.

fvm/src/machine/engine.rs
magik6k: That's what the MultiEngine is doing, and we were already re-using the engine in filecoin-ffi.

testing/integration/src/builtin.rs
magik6k: Yeah, the integration test runner needs a bundle to load, even if we don't use any actor code. Ideally we could run integration tests without actor bundles - that would also make those tests run much faster.

Kubuxu:
I would love to see a test case of an empty loop because it is a quite uncomfortable edge case (if no gas gets charged within the loop).

raulk (approved):
Thanks for addressing feedback.

fvm/src/machine/engine.rs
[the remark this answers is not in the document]
magik6k: As far as I can tell wasm execution uses the native stack - when this is set to a high value the out_of_stack test will fail with:

```
thread 'out_of_stack' has overflowed its stack
fatal runtime error: stack overflow
```

Given that, I don't think we're allocating here

PR #497, opened 2022-04-15
fix: make sure we can read CIDs at the end of memory

Stebalien:


fvm/src/syscalls/context.rs
```diff
@@ -67,3 +79,75 @@ impl Memory {
...
 }
+
+#[cfg(test)]
+mod test {
+    use cid::multihash::MultihashDigest;
+
+    use super::*;
+
+    // TODO: move this somewhere more useful.
+    macro_rules! expect_syscall_err {
```
Stebalien, line 90: We should probably be writing a _lot_ of kernel/syscall tests with this.

raulk (approved):
This is really awesome! These tests are also good potential targets for fuzzing.

PR #499, opened 2022-04-18
chore: release 0.7.1

arajasek:
Not sure whether we want patch release changelogs as well.

### DIFF
--- a/fvm/CHANGELOG.md
+++ b/fvm/CHANGELOG.md
@@ -2,6 +2,13 @@
 
 Changes to the reference FVM implementation.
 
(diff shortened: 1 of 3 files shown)


Stebalien (commented):
Let's cut a minor version. Technically, we've changed the 'actor_debugging` function (so this is a breaking change). It's not huge, but there's no real cost to cutting a minor version (at least of this crate).

Stebalien:
Eh, whatever. Nobody's using that API yet anyways.

Stebalien (approved):

PR #501, opened 2022-04-19
fix: gas: charge for execunits based on snapshot

arajasek:
Commit #1: I think this got clobbered in one of the review periods. 
Commit #2: The GasTracker's gas_available doesn't actually get updated through execution, so we should snapshot and consume based on gas_used instead.

Thanks to AWade for the catches.

### DIFF
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -1,3 +1,5 @@
+use std::cmp::max;
+
 use anyhow::Context;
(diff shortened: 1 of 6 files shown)


arajasek:
I feel like it might be useful to add a `gas_remaining` method on the `GasTracker` and maybe rename `gas_available` and its getter to something like `inital_gas`.

magik6k (approved):
Looks reasonably correct.

Issue #506, opened 2022-04-21
optimizable syscalls


raulk:
## Context

Using the execution tracing framework introduced in #405, we have obtained detailed traces of Wasm execution on nv15. Those traces indicate there's significant repetitiveness in certain syscall paths. Concretely:

- ipld::open => ipld::read (sequence used to read a block from the store)
- ipld::stat => ipld::read (sequence used to read the parameters block)
- ipld::create => ipld::cid (sequence used to create a block and immediately obtain its CID)
- self::root => ipld::open (sequence used to read our state)
- send::send => ipld::stat (sequence used to read a return value)
- paths involving message::* syscalls
- paths involving network::* syscalls

We propose ways to optimize these paths to reduce context switching and potential associated gas costs.

## Optimization 1: Static contextual data

We currently have dedicated syscalls to obtain every one of this fields. Most actors will want to query these fields, so grouping them behind a single syscall, or injecting them into an exported, statically-allocated global, would make this data available without the multiple-syscall overhead.

- caller actor id (u64)
- receiver actor id (u64)
- method number (u64)
- value received (u64, u64)
- current epoch (i64; why isn't this a u64?)
- network version (u32)

Optionally (these attributes are hardly ever used by actors directly and in most cases would go unused):

- base fee (u64, u64)
- circulating supply (u64, u64)

## Optimization 2: optimistically passing buffers to ipld::read

Instead of performing an ipld::stat to first obtain the length of the block to allocate a buffer actor-side, we could allocate an arbitrarily sized buffer. If we guessed the size correctly, we're good. If we guessed it incorrectly, the FVM would return the _real_ size so we can expand the buffer before calling ipld::read again.

In the optimistic outcome, we perform 1 syscall. In the pessimistic outcome, we perform 2 syscalls (just like today) and an extra memory operation.

This is probably worth the tradeoff in the case of message parameters.

## Optimization 3: ipld::open_read

Similar to optimization 2, this new syscall would open a block and perform the read into an optimistically sized buffer. If we guessed the size correctly, we're good. If not, we would get the block handle back with its correct size, so we can allocate an appropriate buffer and call ipld::read.

## Optimization 4: pre-opening the state root

Assuming that most actors want to access their state, the FVM could pre-open their state root block and place it inside a static block handle, which is then set on the static context data from optimization 1.

[...]

PR #509, opened 2022-04-22
feat: optimize static message information into a single syscall

dignifiedquire:
ref https://github.com/filecoin-project/ref-fvm/issues/506


sdk/src/sys/message.rs
```diff
@@ -1,33 +1,23 @@
...
-
-    /// Returns the value that was received.
+    /// - The caller's actor ID.
+    /// - The receiver's actor ID (i.e. ourselves).
+    /// - The method number from the message.
+    /// - The value that was received.
+    /// - The current epoch.
+    /// - The network version.
+    /// - The base fee for the current epoch.
+    /// - The circulating supply.
```
Kubuxu, line 17: Maybe add note that these values are constant during execution and should be cached by sdks.

dignifiedquire:
@Kubuxu @Stebalien @raulk any ideas/pointers what needs to be fixed for the conformance tests?

fvm/src/syscalls/message.rs
```diff
@@ -1,26 +1,33 @@
...
+            .try_into()
+            .context("base-fee exceeds u128 limit")
+            .or_fatal()?,
+        circulating_supply: context
+            .kernel
+            .total_fil_circ_supply()?
+            .try_into()
+            .context("circulating supply exceeds u128 limit")
+            .or_fatal()?,
+    })
```
Stebalien, line 32: It's going to be difficult to charge gas for this. I'd collapse this all into a single kernel method (if possible?). Or maybe a few kernel methods.

fvm/src/syscalls/message.rs
```diff
@@ -1,26 +1,33 @@
...
-}
-
-pub fn value_received(context: Context<'_, impl Kernel>) -> Result<sys::TokenAmount> {
-    context
-        .kernel
-        .msg_value_received()
-        .try_into()
-        .context("invalid token amount")
-        .or_fatal()
+pub fn details(context: Context<'_, impl Kernel>) -> Result<MessageDetails> {
```
Stebalien, line 7: These aren't _just_ message details, they're invocation "context". Maybe:

1. Rename the `vm` module to `invocation`.
2. Rename this to `metadata` or `context`.

sdk/src/sys/message.rs
Stebalien: I would also move this documentation to the `MessageDetails` struct itself. Otherwise, it's guaranteed to get out of sync.

sdk/src/sys/message.rs
```diff
@@ -1,33 +1,23 @@
 //! Syscalls for reading message metadata.
 
+use fvm_shared::sys::out::message::MessageDetails;
```
Stebalien, line 3:
```suggestion
#[doc(inline)]
pub use fvm_shared::sys::out::message::MessageDetails;
```

Stebalien:
> @Kubuxu @Stebalien @raulk any ideas/pointers what needs to be fixed for the conformance tests?

1. Merge this.
2. Update the builtin actors (v6/v7) to the latest sdk/shared.
3. Update them here.

At the moment, there's no way to make this work without breaking master for a time. In the _future_, this won't be an issue because we'll need to preserve backwards compatibility in the syscall ABIs.

Stebalien:
https://github.com/filecoin-project/ref-fvm/issues/512

shared/src/sys/out.rs
```diff
@@ -62,3 +62,22 @@ pub mod crypto {
...
+
+    #[derive(Debug, Copy, Clone)]
+    #[repr(packed, C)]
+    pub struct MessageDetails {
+        pub caller: ActorID,
+        pub receiver: ActorID,
+        pub method_number: MethodNum,
+        pub value_received: TokenAmount,
+        pub curr_epoch: ChainEpoch,
+        pub version: u32,
```
Stebalien, line 79: This is the network version? I'd change it to `network_version` to be clear.

Stebalien: Also, let's put this _last_. That way we can avoid unaligned loads (due to `repr(packed)`).

Really, I'd order this as:

```rust
pub struct MessageDetails {
    pub base_fee: TokenAmount,
    pub circulating_supply: TokenAmount,
    pub value_received: TokenAmount,
    pub caller: ActorID,
    pub receiver: ActorID,
    pub method_number: MethodNum,
    pub network_curr_epoch: ChainEpoch,
    pub version: u32,
}
```

That way everything is nice and aligned.

Stebalien:
Ok, one thing I realized here is that adding, removing, and changing metadata is a breaking change.

Given that, we should consider slimming this down to the "core" metdata. Basically:

- caller
- receiver
- method number (well, maybe..., but changing this will be a big change anyways)
- value received
- current epoch
- network version

However:

1. I don't even _want_ to expose `circ_supply` to user actors. The circulating supply model is
   likely going to need to change in M2 because we'll have additional sources of "lockup".
2. We very rarely care about the base fee. And we may have _multiple_ base fees in the future.

shared/src/sys/out.rs
dignifiedquire: fixed

sdk/src/sys/message.rs
dignifiedquire: fixed manually

sdk/src/network.rs
```diff
@@ -5,22 +5,17 @@ use fvm_shared::econ::TokenAmount;
 use fvm_shared::version::NetworkVersion;
 
 use crate::sys;
+use crate::vm::MESSAGE_DETAILS;
```
raulk, line 8: Can we unify the naming here? Should this be Invocation Context (how it's called kernel side)?

sdk/src/vm.rs
```diff
@@ -1,6 +1,78 @@
...
+use fvm_shared::sys::{BlockId, Codec};
+use fvm_shared::{ActorID, MethodNum};
+
+use crate::{sys, SyscallResult};
+
+/// BlockID representing nil parameters or return data.
+pub const NO_DATA_BLOCK_ID: u32 = 0;
+
+lazy_static::lazy_static! {
+    pub(crate) static ref MESSAGE_DETAILS: InvocationContext = {
```
raulk, line 16:
```suggestion
    pub(crate) static ref INVOCATION_CONTEXT: InvocationContext = {
```

sdk/src/vm.rs
```diff
@@ -1,6 +1,78 @@
...
+        unsafe {
+            sys::vm::context().expect("failed to lookup message details")
```
raulk, line 18:
```suggestion
            sys::vm::context().expect("failed to look up invocation context")
```

shared/src/sys/out.rs
```diff
@@ -62,3 +62,26 @@ pub mod crypto {
...
+        /// The value that was received.
+        pub value_received: TokenAmount,
+        /// The caller's actor ID.
+        pub caller: ActorID,
+        /// The receiver's actor ID (i.e. ourselves).
+        pub receiver: ActorID,
+        /// The method number from the message.
+        pub method_number: MethodNum,
+        /// The current epoch.
+        pub network_curr_epoch: ChainEpoch,
```
raulk, line 83: Can drop the `network_` prefix here as there's no namespacing needed any longer, nor is there ambiguity.

raulk (approved):
* Ideally I'd like to see the kernel-side methods unified too. Happy to do it in a follow-up PR to prevent the diff from growing too much.
* But please do replace `MESSAGE_DETAILS` with `INVOCATION_CONTEXT`  before merging.

fvm/src/syscalls/vm.rs
```diff
@@ -45,3 +47,21 @@ pub fn abort(
...
+        method_number: context.kernel.msg_method_number(),
+        value_received: context
+            .kernel
+            .msg_value_received()
+            .try_into()
+            .context("invalid token amount")
+            .or_fatal()?,
+        network_curr_epoch: context.kernel.network_epoch(),
+        network_version: context.kernel.network_version() as u32,
+    })
```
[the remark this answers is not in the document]
dignifiedquire: I tried to keep the changes to a minimum, no strong preference

shared/src/sys/out.rs
dignifiedquire: @Stebalien wanted the prefix back, so the two of you have to fight that one out

sdk/src/vm.rs
dignifiedquire: fixed

fvm/src/syscalls/vm.rs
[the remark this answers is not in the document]
Stebalien: I would prefer to have them consolidated as well, but let's iterate on master. _That_, at least, is a less-breaking change.

shared/src/sys/out.rs
Stebalien: Ah, sorry, I wanted the `network_` prefix for the _version_ (because we have many versions). The _epoch_ is unambiguous.

raulk (approved):

PR #515, opened 2022-04-27
feat: raw signatures in syscalls

### LINKED ISSUE #511: Avoid CBOR for all "userspace" syscalls
It's fine to use CBOR for "privileged" syscalls, but we should avoid using CBOR for syscalls intended to be called from user code as CBOR decoding is complex and difficult to charge for.

Specifically:

- `verify_signature` should take the signature _type_ and signature as separate fields instead of as one cbor-encoded blobl.
- All of the other verify_* functions should (at least as of M2) only be callable from specific "blessed" actors.

Stebalien:
Don't pass signatures as cbor-encoded objects with signature types. Instead, pass the type and the signature bytes separately.

This change _also_ avoids copying the signature when verifying it.

fixes https://github.com/filecoin-project/ref-fvm/issues/511

### DIFF
--- a/fvm/src/kernel/default.rs
+++ b/fvm/src/kernel/default.rs
@@ -18,6 +18,7 @@ use fvm_shared::commcid::{
     cid_to_data_commitment_v1, cid_to_replica_commitment_v1, data_commitment_v1_to_cid,
 };
 use fvm_shared::consensus::ConsensusFault;
+use fvm_shared::crypto::signature;
 use fvm_shared::econ::TokenAmount;
 use fvm_shared::error::ErrorNumber;
 use fvm_shared::piece::{zero_piece_commitment, PaddedPieceSize};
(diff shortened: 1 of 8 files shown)


fvm/src/kernel/default.rs
```diff
@@ -459,19 +460,40 @@ where
...
-            self.call_manager
-                .price_list()
-                .on_verify_signature(signature.signature_type()),
-        )?;
+        // TODO: remove this after the nv16 upgrade.
+        //
+        // < nv16: charge gas for the key lookup after we charge for the signature
+        // verification.
+        //
+        // <= nv16: charge for gas for the key lookup first (because we need to know the type of
```
Kubuxu, line 472:
```suggestion
        // >= nv16: charge for gas for the key lookup first (because we need to know the type of
```

Kubuxu:
Clippy fails

Kubuxu (commented):
LGTM

PR #516, opened 2022-04-27
feat: parameterize the hash syscall over the hash type

### LINKED ISSUE #270: Hash digest syscall
Currently, we have a single `hash_blake2b` syscall. I'd like to consider a more general `hash` function:

```rust
mod crypto {
    pub fn hash(multihash_code: u64, data: *const u8, data_len: u32, digest: *mut u8, digest_len: u32) -> u32 /* bytes written to digest */;
}
```

This would let us easily support arbitrary multihash codes in the FVM itself. NOTE: it would not write a _multihash_ to `digest`, just the digest part.

Pros:

1. Easier to deploy code that's generic over hash functions without having to upgrade it later.
2. Smaller API surface.
3. Multihash native.

Cons:

1. No static "not supported error". But we're going to run into the same problem with IPLD codecs. IMO, we should have a static "requires" section in the WASM module to export dependencies like this.

Stebalien:
This makes it a bit easier to add support for additional hash functions later. We could just have many syscalls, but parameterizing over a multicodec makes it easier to implement/use multihash libraries in actors.

fixes #270

### DIFF
--- a/fvm/src/kernel/default.rs
+++ b/fvm/src/kernel/default.rs
@@ -474,9 +474,17 @@ where
         Ok(signature.verify(plaintext, &signing_addr).is_ok())
     }
 
-    fn hash_blake2b(&mut self, data: &[u8]) -> Result<[u8; 32]> {
+    fn hash(&mut self, code: u64, data: &[u8]) -> Result<[u8; 32]> {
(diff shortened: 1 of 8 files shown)


raulk (approved):
Need to fix conformance once we release these changes and bubble to built-in actors, and back here.

Issue #504, opened 2022-04-20
Replace multihash::Code with u64 in the Blockstore interfaces


Stebalien:
Unfortunately, depending on "code" here means we need to pull in the hasher implementations themselves. This is a problem because, in the actors, we want to do hashing _outside_.

Stebalien:
A slightly less breaking alternative is to make it generic over `Code` (implementing a `Code` that uses the sdk).

PR #524, opened 2022-05-03
fix: make things independent of the Multihash code table

vmx:
It should now be possible to specify your own Multihash code
table.

@Stebalien this one is for you. Please let me know if this is something we want to tackle and whether it's the right direction. Ideally I'd like to use the default alloc value as little as possible and make it explicit most of the time.

It currently build, but I haven't even tried to compile the tests yet, as I'd like to get basic feedback, before I start making everything pretty.

### DIFF
--- a/ipld/blockstore/src/lib.rs
+++ b/ipld/blockstore/src/lib.rs
@@ -1,7 +1,8 @@
 use std::rc::Rc;
 
 use anyhow::Result;
-use cid::{multihash, Cid};
+use cid::multihash::MultihashDigest;
+use cid::CidGeneric;
 
 pub mod tracking;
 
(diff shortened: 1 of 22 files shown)


ipld/blockstore/src/lib.rs
```diff
@@ -1,40 +1,49 @@
...
+/// The default size for multihashes is currently set to `64` as it's the value the `multihash` is
+/// using is its default code table. Once a custom code table is used, it could be reduced to `32`.
+const DEFAULT_MULTIHASH_ALLOC_SIZE: usize = 64;
 
 /// An IPLD blockstore suitable for injection into the FVM.
 ///
 /// The cgo blockstore adapter implements this trait.
-pub trait Blockstore {
+///
+pub trait Blockstore<const S: usize = DEFAULT_MULTIHASH_ALLOC_SIZE> {
```
Stebalien, line 23: Being generic over the code type makes sense, but being generic over the size really isn't worth it:

1. It will lead to an annoying refactor in the actors (which will have to be backported).
2. It will conflict with all open PRs.

Stebalien:
Note: don't spend too much time on this right now. I'd like to see how much of breaking change this would be if we _only_ make these types generic over the code, but we're trying to wrap up M1 ASAP so bubbling breaking changes that _can_ be punted till M2 isn't a priority.

ipld/blockstore/src/lib.rs
vmx: Agreed, it didn't really felt nice anyway. I'll look if I find a better, less intrusive way.

Stebalien:
This is covered by a few issues (#504, and arguably #1238) but will need a lot of work to revive. Closing but keeping the branch.

Issue #523, opened 2022-05-03
propagate all gas outputs from message execution


raulk:
ApplyRet is missing these attributes:

- base fee burn (FIL)
- over estimation burn
- refund (FIL)
- refund (gas)
- gas burned

They are not propagated from `DefaultExecutor#finish_message()`, nor through the FFI (obviously), which means that clients can't access them.

Several Lotus itests are failing as a result, and there may be JSON-RPCs returning zero data on these fields.

vyzo:
Implementation:
- https://github.com/filecoin-project/ref-fvm/pull/526
- https://github.com/filecoin-project/filecoin-ffi/pull/275
- https://github.com/filecoin-project/lotus/pull/8593

raulk:
All work here is done.

PR #526, opened 2022-05-04
propagate all gas outputs in ApplyRet

vyzo:
See #523.

Also fixes a bug in system actors state, which was broken and revealed by the cbor4ii library (which is stricter than its predecessor).

### DIFF
--- a/fvm/src/executor/mod.rs
+++ b/fvm/src/executor/mod.rs
@@ -5,6 +5,7 @@ use std::fmt::Display;
 pub use default::DefaultExecutor;
 use fvm_ipld_encoding::RawBytes;
 use fvm_shared::bigint::{BigInt, Sign};
+use fvm_shared::econ::TokenAmount;
 use fvm_shared::error::ExitCode;
 use fvm_shared::message::Message;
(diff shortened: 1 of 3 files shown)


Stebalien (approved):

fvm/src/executor/mod.rs
```diff
@@ -70,6 +71,14 @@ pub struct ApplyRet {
     pub penalty: BigInt,
     /// Tip given to miner from message.
     pub miner_tip: BigInt,
+
+    // Gas stuffs
+    pub base_fee_burn: TokenAmount,
+    pub over_estimation_burn: TokenAmount,
+    pub refund: TokenAmount,
```
raulk, line 78: Let's continue using BigInt here. TokenAmount is aliased to BigInt, so not a big deal, but I'd prefer to contain entropy.

raulk (approved):

fvm/src/executor/mod.rs
Stebalien: Hm. We should probably go the other way so we can eventually get rid of that alias, right?

PR #527, opened 2022-05-04
feat: refactor gas charging to remove "borrowing"

Stebalien:
Basically:

1. On entering wasm, we call `update_gas_available` to update the available gas for the actor.
2. On exiting wasm, we call `charge_for_exec` to charge for execution gas.

Unlike the previous code:

1. We can handle this the same way in both the CallManager and the syscalls.
2. We never "borrow" or "return" the gas from/to the gas tracker.

### DIFF
--- a/fvm/src/syscalls/mod.rs
+++ b/fvm/src/syscalls/mod.rs
@@ -1,7 +1,11 @@
+use std::mem;
+
+use anyhow::{anyhow, Context as _};
 use cid::Cid;
-use wasmtime::{Global, Linker};
+use wasmtime::{AsContextMut, Global, Linker, Val};
 
 use crate::call_manager::backtrace;
+use crate::kernel::ExecutionError;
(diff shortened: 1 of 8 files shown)


fvm/src/syscalls/mod.rs
```diff
@@ -30,11 +34,62 @@ pub struct InvocationData<K> {
     /// after receiving this error without calling any other syscalls.
     pub last_error: Option<backtrace::Cause>,
 
-    /// The global containing remaining available gas
+    /// The global containing remaining available gas.
     pub avail_gas_global: Global,
+    /// The last-set milligas limit. When `cahrge_for_exec` is called, we charge for the
```
magik6k, line 39:
```suggestion
    /// The last-set milligas limit. When `charge_for_exec` is called, we charge for the
```

magik6k (approved):

Issue #363, opened 2022-03-07
Implement FIP-0032


jennijuju:
FIP: https://github.com/filecoin-project/FIPs/pull/315

- [x] Acquire data to price execution gas: #333
- [x] #375
- [x] #376
- [x] #377

raulk:
Work breadown created -- @Stebalien off to you.

PR #534, opened 2022-05-06
implement FIP-0032 gas parameters and charges

raulk:
Fully implements https://github.com/filecoin-project/FIPs/pull/368 in ref-fvm.

Also changes the unit in pricelists to milligas, and removes now unnecessary functions that dealt with full-unit gas charges.
Adds i64 Gas and Milligas type aliases for extra clarity (but not safety) in APIs.

Static parameters in pricelists were updated to use a non-saturating macro.

Runtime milligas conversions were needed here:
- message prevalidation: needs to do milligas conversion when evaluating the inclusion cost.
- charge_gas syscall handler: I placed it here instead of making `Kernel#charge_gas` accept full Gas for two reasons, to preserve unit homogeneity in `charge_*` operations throughout the codebase.

## Note

Need to review all `on_*` price list formulae because some aren't using saturation arithmetics (inherited). I don't want to continue growing this patch, so we should follow up with this on master.

---

Closes https://github.com/filecoin-project/ref-fvm/issues/363.

### DIFF
--- a/fvm/src/gas/price_list.rs
+++ b/fvm/src/gas/price_list.rs
@@ -671,7 +671,7 @@ impl PriceList {
         GasCharge::new(
             "OnBlockLink",
             self.block_link_base
-                .saturating_add(2 as i64.saturating_mul(memcpy)),
+                .saturating_add((2 as i64).saturating_mul(memcpy)),
             self.block_link_storage_per_byte_multiplier
                 .saturating_mul(self.storage_gas_multiplier)
                 .saturating_mul(size),
--- a/fvm/src/gas/price_list.rs
+++ b/fvm/src/gas/price_list.rs
@@ -16,13 +16,7 @@ use fvm_wasm_instrument::parity_wasm::elements::Instruction;
 use lazy_static::lazy_static;
 use num_traits::Zero;
 
-use super::GasCharge;
-
-macro_rules! to_milligas {
-    ($ex:expr) => {
-        $ex * crate::gas::MILLIGAS_PRECISION
-    };
-}
+use super::{to_milligas, GasCharge};
 
 lazy_static! {
     static ref OH_SNAP_PRICES: PriceList = PriceList {
--- a/fvm/src/gas/price_list.rs
+++ b/fvm/src/gas/price_list.rs
@@ -441,6 +441,7 @@ impl PriceList {
                 * self.storage_gas_multiplier,
         )
     }
+
     /// Returns the gas required for storing the response of a message in the chain.
     #[inline]
     pub fn on_chain_return_value(&self, data_size: usize) -> GasCharge<'static> {
@@ -450,6 +451,7 @@ impl PriceList {
             data_size as i64 * self.on_chain_return_value_per_byte * self.storage_gas_multiplier,
         )
     }
+
     /// Returns the gas required when invoking a method.
     #[inline]
     pub fn on_method_invocation(
@@ -469,6 +471,7 @@ impl PriceList {
         }
         GasCharge::new("OnMethodInvocation", ret, 0)
     }
+
     /// Returns the gas required for creating an actor.
     #[inline]
     pub fn on_create_actor(&self) -> GasCharge<'static> {
@@ -478,6 +481,7 @@ impl PriceList {
             self.create_actor_storage * self.storage_gas_multiplier,
         )
     }
+
     /// Returns the gas required for deleting an actor.
     #[inline]
     pub fn on_delete_actor(&self) -> GasCharge<'static> {
@@ -487,6 +491,7 @@ impl PriceList {
             self.delete_actor * self.storage_gas_multiplier,
         )
     }
+
     /// Returns gas required for signature verification.
     #[inline]
     pub fn on_verify_signature(&self, sig_type: SignatureType) -> GasCharge<'static> {
@@ -496,11 +501,13 @@ impl PriceList {
         };
         GasCharge::new("OnVerifySignature", val, 0)
     }
+
     /// Returns gas required for hashing data.
     #[inline]
     pub fn on_hashing(&self, _: usize) -> GasCharge<'static> {
         GasCharge::new("OnHashing", self.hashing_base, 0)
     }
+
     /// Returns gas required for computing unsealed sector Cid.
     #[inline]
     pub fn on_compute_unsealed_sector_cid(
@@ -514,6 +521,7 @@ impl PriceList {
             0,
         )
     }
+
     /// Returns gas required for seal verification.
     #[inline]
     pub fn on_verify_seal(&self, _info: &SealVerifyInfo) -> GasCharge<'static> {
@@ -554,11 +562,13 @@ impl PriceList {
             0,
         )
     }
+
     /// Returns gas required for replica verification.
     #[inline]
     pub fn on_verify_replica_update(&self, _replica: &ReplicaUpdateInfo) -> GasCharge<'static> {
         GasCharge::new("OnVerifyReplicaUpdate", self.verify_replica_update, 0)
     }
+
     /// Returns gas required for PoSt verification.
     #[inline]
     pub fn on_verify_post(&self, info: &WindowPoStVerifyInfo) -> GasCharge<'static> {
@@ -577,6 +587,7 @@ impl PriceList {
 
         GasCharge::new("OnVerifyPost", gas_used, 0)
     }
+
     /// Returns gas required for verifying consensus fault.
     #[inline]
     pub fn on_verify_consensus_fault(&self) -> GasCharge<'static> {
@@ -614,6 +625,7 @@ impl PriceList {
             0,
         )
     }
+
     /// Returns the gas required for reading a loaded object.
     #[inline]
     pub fn on_block_read(&self, data_size: usize) -> GasCharge<'static> {
--- a/fvm/src/gas/price_list.rs
+++ b/fvm/src/gas/price_list.rs
@@ -16,44 +16,52 @@ use fvm_wasm_instrument::parity_wasm::elements::Instruction;
 use lazy_static::lazy_static;
 use num_traits::Zero;
 
-use super::{to_milligas, GasCharge};
+use super::GasCharge;
+
+/// Converts a static value to milligas. This operation does not saturate,
+/// and should only be used with constant values in pricelists.
+macro_rules! static_milligas {
+    ($ex:expr) => {
+        $ex * $crate::gas::MILLIGAS_PRECISION
+    };
+}
 
 lazy_static! {
     static ref OH_SNAP_PRICES: PriceList = PriceList {
-        compute_gas_multiplier: to_milligas!(1),
-        storage_gas_multiplier: to_milligas!(1300),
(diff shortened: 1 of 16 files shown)


fvm/src/gas/price_list.rs
```diff
@@ -636,12 +694,30 @@ pub fn price_list_by_network_version(network_version: NetworkVersion) -> &'stati
...
-        Some(self.exec_instruction_cost_milli)
+    fn instruction_cost(&self, instruction: &Instruction) -> Option<u64> {
+        if self.exec_instruction_cost_milli == 0 {
+            return None;
+        }
+
+        // Rules valid for nv16. We will need to be generic over Rules (massive
+        // generics tax), use &dyn Rules (which breaks other things), or pass
+        // in the network version, or rules version, to vary these prices going
+        // forward.
```
raulk, line 705: @Stebalien something to take into account.

fvm/src/gas/price_list.rs
```diff
@@ -18,100 +18,106 @@ use num_traits::Zero;
...
+    ($ex:expr) => {
+        $ex * crate::gas::MILLIGAS_PRECISION
+    };
+}
+
 lazy_static! {
     static ref OH_SNAP_PRICES: PriceList = PriceList {
-        compute_gas_multiplier: 1,
-        storage_gas_multiplier: 1300,
+        compute_gas_multiplier: to_milligas!(1),
```
Kubuxu, line 29: If `compute_gas_multiplier` was used, this would be wrong (you would end up multiplying by 1000 twice) but it looks like `compute_gas_multiplier` is not used.

Kubuxu: We should probably remove it from the price list if it isn't going to be used.

fvm/src/gas/price_list.rs
```diff
@@ -240,31 +253,29 @@ lazy_static! {
...
+
+        block_read_base: 0,
+        block_stat_base: 0,
+
+        syscall_cost: to_milligas!(14000),
+        extern_cost: to_milligas!(21000),
 
         wasm_rules: WasmGasPrices{
-            exec_instruction_cost_milli: 5000,
+            exec_instruction_cost_milli: to_milligas!(5) as u64,
```
Kubuxu, line 277: Gas per exec unit is defined to be 4 in FIP-0032

fvm/src/gas/price_list.rs
```diff
@@ -609,21 +666,22 @@ impl PriceList {
...
-        // TODO: The FIP makes it sound like this would need 2 memcpys, is that what's desired?
+        let size = data_size as i64;
+        let memcpy = self.block_memcpy_per_byte_cost.saturating_mul(size);
         GasCharge::new(
             "OnBlockLink",
-            self.block_link_base,
-            // data_size as i64 * self.block_link_per_byte_cost * self.storage_gas_multiplier,
-            self.block_link_per_byte_cost
+            self.block_link_base
+                .saturating_add((2 as i64).saturating_mul(memcpy)),
```
Kubuxu, line 674: Could we move the hard-coded `2` here to either price list or at least local constant?

fvm/src/kernel/default.rs
```diff
@@ -816,6 +816,10 @@ where
                 .price_list()
                 .on_get_randomness(entropy.len()),
         )?;
+
+        self.call_manager
+            .charge_gas(self.call_manager.price_list().on_extern())?;
+
```
Kubuxu, line 822: Nit: This repeated pattern makes me think that we should have a capability to either charge for both charges at once or the capability to combine charges into one but it isn't something that needs to get done.

It also shouldn't be consensus critical if we want to add this capability in future.

fvm/src/gas/price_list.rs
raulk: Can't do that because this is literally "twice the amount", but I will add a comment to clarify. Good catch.

fvm/src/gas/price_list.rs
raulk: Correct, this was outdated. Fixed.

fvm/src/gas/price_list.rs
raulk: Ack.

fvm/src/kernel/default.rs
raulk: Agree. I started hacking on a const generics version of `charge_gas_n` that took an array and applied all charges sequentially. A few months ago I had also proposed a closure-based version of `charge_gas` that took the price list as an argument. That would also remove some pedanticness. However, we can explore these APIs later; right now it doesn't hurt having two explicit charges.

raulk:
I have no idea why clippy complains here about this and not on master :-/

fvm/src/gas/price_list.rs
raulk: Yes, this, and the storage gas multiplier were wrongly scaled to millis.

fvm/src/syscalls/mod.rs
```diff
@@ -62,7 +61,10 @@ pub fn update_gas_available(
     Ok(())
 }
 
-pub fn charge_for_exec(
+/// Applies gas charges upon passing control to the FVM via a syscall.
+/// It first applies accumulated execution gas, and then applies the syscall
+/// charge for the currently active syscall.
+pub fn apply_charges_on_syscall(
```
magik6k, line 67: Being nitpicky here, but _technically_ this is a charge on exiting wasm, not just syscalls.

fvm/src/gas/price_list.rs
```diff
@@ -300,14 +311,10 @@ impl StepCost {
     }
 }
 
-/// Provides prices for operations in the VM
+/// Provides prices for operations in the VM.
+/// All costs are in milligas.
```
magik6k, line 315: Can we use the `Milligas` type here?

magik6k (approved):
Mostly being nitpicky, this looks good.

fvm/src/gas/price_list.rs
Stebalien: In this case, we can likely do `&'static dyn Rules`. Or, if we can't statically allocate, `Box<dyn Rules>`.

fvm/src/kernel/default.rs
```diff
@@ -296,6 +296,9 @@ where
         self.call_manager
             .charge_gas(self.call_manager.price_list().on_block_open_base())?;
 
+        self.call_manager
+            .charge_gas(self.call_manager.price_list().on_extern())?;
```
Stebalien, line 300: Personally, I'd just charge these together. I.e., make `on_block_open_base` add-in the extern cost. That way we don't accidentially _not_ charge the extern cost in some cases.

fvm/src/kernel/default.rs
```diff
@@ -762,20 +769,14 @@ where
     fn gas_available(&self) -> i64 {
         self.call_manager.gas_tracker().gas_available()
     }
+
     fn milligas_available(&self) -> i64 {
         self.call_manager.gas_tracker().milligas_available()
     }
 
-    fn charge_gas(&mut self, name: &str, compute: i64) -> Result<()> {
+    fn charge_gas(&mut self, name: &str, compute: Milligas) -> Result<()> {
```
Stebalien, line 777:
```suggestion
    fn charge_milligas(&mut self, name: &str, compute: Milligas) -> Result<()> {
```

fvm/src/syscalls/mod.rs
Stebalien: This is actually a bug. We call this when we return from wasm as well.

fvm/src/kernel/default.rs
Stebalien: Otherwise, this is _really_ confusing.

fvm/src/gas/price_list.rs
```diff
@@ -124,85 +131,91 @@ lazy_static! {
         get_randomness_per_byte: 0,
 
         block_memcpy_per_byte_cost: 0,
-        block_io_per_byte_cost: 0,
-        block_link_per_byte_cost: 1,
 
-        block_open_base: 114617,
-        block_read_base: 0,
+        block_open_base: static_milligas!(114617),
+        block_open_memret_per_byte_cost: 0,
```
Stebalien, line 136: What is memret?

raulk: The memory retention cost (see section under https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0032.md#ipld-state-management-gas).

raulk: (Can rename to something else, if it sounds like memory return).

fvm/src/kernel/default.rs
raulk: Works for me.

fvm/src/gas/price_list.rs
```diff
@@ -124,85 +131,91 @@ lazy_static! {
...
 
-        on_chain_message_compute_base: 38863,
-        on_chain_message_storage_base: 36,
-        on_chain_message_storage_per_byte: 1,
+        on_chain_message_compute_base: static_milligas!(38863),
+        on_chain_message_storage_base: static_milligas!(36),
+        on_chain_message_storage_per_byte: static_milligas!(1),
 
-        on_chain_return_value_per_byte: 1,
+        on_chain_return_value_per_byte: static_milligas!(1),
```
Kubuxu, line 162: Personally, I would set the `*_per_byte` storage costs to 1 instead of miligas.
They have an implicit unit of bytes, and then they get multiplied by `storage_gas_multiplier` would be `milltigas/byte`.

fvm/src/kernel/default.rs
raulk: Hm, in fact, I think it might be better that way because when we remove the extern cost (implementing the node-provided ring buffer idea for randomness), we can contain the gas charge change to the PriceList. (Otherwise we would have to do network version detection here to charge conditionally).

fvm/src/syscalls/mod.rs
raulk: Good catch, thanks for that!

fvm/src/gas/price_list.rs
raulk: Hm, it's actually implemented the other way around now. The _multiplier_ is no longer denominated in milligas, but as a raw factor. Are you good with that?

arajasek (approved):

fvm/src/gas/price_list.rs
raulk: Done.

fvm/src/kernel/default.rs
raulk: Done.

fvm/src/kernel/default.rs
raulk: Done.

fvm/src/syscalls/mod.rs
raulk: Will fix this tomorrow.

fvm/src/gas/price_list.rs
Kubuxu: Yeah, I’m fine with that

PR #535, opened 2022-05-06
fix: make sure bitfield is within limits

vmx:
Error if one tries to decode a bitfield that is too large.

I don't know if that's really needed or not. I've added the test case from the Forest Actor Review.


Stebalien (approved):
Nice catch!

Stebalien:
But it looks like there's a compilation error.

vmx:
Rebased, ready to be merged.

Stebalien:
Thanks!

PR #537, opened 2022-05-08
conformance: remove crypto verification stubbing.

raulk:
I suspect this existed in the Go driver to avoid the filecoin-ffi dependency on proofs (and the associated build shenanigans). But we don't have those problems in ref-fvm.

### DIFF
--- a/testing/conformance/src/vm.rs
+++ b/testing/conformance/src/vm.rs
@@ -418,23 +418,16 @@ where
         self.0.verify_signature(signature, signer, plaintext)
     }
 
-    // NOT forwarded
     fn batch_verify_seals(&mut self, vis: &[SealVerifyInfo]) -> Result<Vec<bool>> {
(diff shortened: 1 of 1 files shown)


Stebalien:
Looks like the problem is missing parameters.

raulk (commented):
Correct, that was the motivator:

```
--> caused by: crypto::verify_replica_update -- No cached parameters found for empty-sector-update-merkletree-poseidon_hasher-8-8-0-3b7f44a9362e3985369454947bc94022e118211e49fd672d52bec1cbfd599d18 [failure finding /var/tmp/filecoin-proof-parameters/v28-empty-sector-update-merkletree-poseidon_hasher-8-8-0-3b7f44a9362e3985369454947bc94022e118211e49fd672d52bec1cbfd599d18.params] (1: illegal argument)
```

PR #541, opened 2022-05-09
chore: use a specific Rust nightly version

vmx:
Currently Rust nightly is needed as the currently used `bellperson` version
needs else wouldn't build on ARM64. Using a specific version makes sure
that we don't get arbitrary Clippy failures on CI.

Also fix Clippy warning that comes up with using this nightly version.

### DIFF
--- a/ipld/amt/benches/amt_benchmark.rs
+++ b/ipld/amt/benches/amt_benchmark.rs
@@ -30,13 +30,11 @@ impl Serialize for BenchData {
 }
 
 impl<'de> Deserialize<'de> for BenchData {
-    #[allow(clippy::let_unit_value)]
     fn deserialize<D>(deserializer: D) -> Result<Self, <D as Deserializer<'de>>::Error>
     where
         D: Deserializer<'de>,
     {
-        let _s: () = Deserialize::deserialize(deserializer)?;
-
+        Deserialize::deserialize(deserializer)?;
         Ok(Self::default())
     }
 }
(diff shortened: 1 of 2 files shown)


vmx:
Sorry for all the force pushing. This should now be on most recent master, with a better fix for the Clippy warning.

ipld/amt/benches/amt_benchmark.rs
```diff
@@ -30,13 +30,11 @@ impl Serialize for BenchData {
 }
 
 impl<'de> Deserialize<'de> for BenchData {
-    #[allow(clippy::let_unit_value)]
     fn deserialize<D>(deserializer: D) -> Result<Self, <D as Deserializer<'de>>::Error>
     where
         D: Deserializer<'de>,
     {
-        let _s: () = Deserialize::deserialize(deserializer)?;
-
```
raulk, line 39: I specifically silenced the warning instead of fixing the code because this is a benchmark and I'm pretty sure the motivation for this was to avoid compiler optimizations. May not apply though?

vmx: I think even with this assignment the compiler should be smart enough to optimize it away if it wants to. But I can try to run it locally to see if I see a difference.

vmx: @raulk It's good to go. When I remove this line completely, the benchmark will panic as there will be trailing data that didn't get deserialized => it's not optimized away.

PR #547, opened 2022-05-09
chore: Second pass on wasmtime config

magik6k:


fvm/src/machine/engine.rs
```diff
@@ -105,13 +114,41 @@ pub fn default_wasmtime_config() -> wasmtime::Config {
...
+
+    // Reiterate some defaults
+    c.guard_before_linear_memory(true);
+    c.interruptable(false);
+    c.parallel_compilation(true);
+
+    c.allocation_strategy(InstanceAllocationStrategy::Pooling {
+        strategy: PoolingAllocationStrategy::NextAvailable,
+        instance_limits: InstanceLimits{
+            count: 1000, // todo set this higher than recursive call limit?
```
magik6k, line 140: Only real todo here, I think.

fvm/src/machine/engine.rs
```diff
@@ -105,13 +117,41 @@ pub fn default_wasmtime_config() -> wasmtime::Config {
...
+    // Reiterate some defaults
+    c.guard_before_linear_memory(true);
+    c.interruptable(false);
+    c.parallel_compilation(true);
+
+    c.allocation_strategy(InstanceAllocationStrategy::Pooling {
+        strategy: PoolingAllocationStrategy::NextAvailable,
+        instance_limits: InstanceLimits {
+            count: 1000, // todo set this higher than recursive call limit?
+            size: 1 << 20,
```
Kubuxu, line 144: In combination with the count, as far as I can see this will lead to pre-allocation of 1GiB of memory.
Are we fine with it?

https://docs.rs/wasmtime/latest/wasmtime/struct.InstanceLimits.html#structfield.size

Kubuxu (approved):
Other than the note on introduction of pooled allocator, SGWM

fvm/src/machine/engine.rs
raulk: What happens when an instance is requested and the pool is exhausted? Is one created on demand, or does wasm instantiation fail?

fvm/src/machine/engine.rs
```diff
@@ -105,13 +117,41 @@ pub fn default_wasmtime_config() -> wasmtime::Config {
...
+    c.debug_info(false);
+    c.generate_address_map(false);
+    c.cranelift_debug_verifier(false);
+
+    // Reiterate some defaults
+    c.guard_before_linear_memory(true);
+    c.interruptable(false);
+    c.parallel_compilation(true);
+
+    c.allocation_strategy(InstanceAllocationStrategy::Pooling {
```
raulk, line 140: We seem to be confirming the default values here. However, we have _never_ used the pooling allocator, so we have no idea if these values are good for us?

fvm/src/machine/engine.rs
mriise: `Instance::new()` will return an error

> Err: Module/instance resource limits are exceeded.

https://docs.rs/wasmtime/latest/wasmtime/struct.Instance.html#impl

raulk: Hmmm, I don't think we want this then. We either have to allocate 4096 instances upfront (call stack depth limit), which is wasteful because most messages won't even reach a few instances, or use the on-demand strategy.

@magik6k if you agree, could you please revert this change? I'd be happy with a hybrid model (once we actually characterise its footprint), but that doesn't seem to exist yet.

Kubuxu (approved):

Stebalien (approved):

PR #548, opened 2022-05-11
transmute fatal errors into `SYS_ASSERTION_FAILED` exit code.

### LINKED ISSUE #508: produce a consensus outcome for fatal errors
## Context

Fatal errors are raised when we encounter system-level unexpected conditions during message execution. They are severe and usually indicate a correctness flaw. Either something is found to be broken at runtime (e.g. state tree cannot be decoded, init actor is not found, etc.), or we've hit some kind of programming error.

Fatal errors occur in the FVM itself, _outside actor code_. Panics in actor code are properly handled by emitting exit code `USR_ASSERTION_FAILED`. See the [FVM error spec](https://github.com/filecoin-project/fvm-specs/blob/main/07-errors.md).

Currently, on a fatal error, the `Executor` fails to apply the messge returns the error to the caller (the Filecoin client). However, there is no possible course of action the caller can take.

There are several outcomes here:

- If the failure is caused by a local condition, the node will fork off from the network.
- If the failure is caused by a condition reproduceable in a subset of nodes, the chain will fork.
- If the failure is caused by a generalised error, the network could halt.

## Proposal

The goal is to allow the chain to make progress in the presence of network-wide fatal errors.

1. Convert fatal errors into receipts with a designated `SYS_INTERNAL_ERROR` exit code.
2. Revert all state tree changes.
3. Consume all gas; it doesn't matter at which point the error happened (could've been at different points in different nodes), the result gas consumption will be identical.

This results in deterministic behaviour that the network can agree on in the presence of fatal errors.

## Result

1. If the fatal error affects a single node, that node will produce the above message receipt leading to a consensus fault with the rest of the network, i.e. node strays off just like before.
2. If the fatal error affects the entire network, the network agrees that an internal error happened during the processing of that message, and moves on without halting. (There's some chance that different nodes will observe different internal errors at different points in the execution, yet they will arrive to the same result; this is intended).

## Implementation notes

This change can be entirely self contained inside the `DefaultExecutor`.

This resolves the "Panics during message execution" area of investigation under #428.

raulk:
Closes #508.

This PR transmutes fatal errors into the `SYS_ASSERTION_FAILED`  exit code at the top layer (Executor) to produce a consensus outcome on fatal errors.

It also modifies the Backtrace components to generalise them so they can tackle syscall errors as well as fatal errors. The cause is informed at different times for each, so the APIs have been adapted accordingly:
- For syscall errors, we inform the cause as soon as the syscall fails. Thus we reset the backtrace and accumulate the propagation chain (`Backtrace#begin()`).
- For fatal errors, we inform the cause at the top. Thus we accumulate the propagation chain, _then_ we inform the cause (`Backtrace#set_cause()`).

I think capturing backtraces (both FVM backtraces and Rust backtraces) is crucial for debuggability. Because I had no way of generating fatal errors to verify the behaviour, I added an integration test which performs no assertions, but prints the backtrace on stdout.

Backtraces from syscall-motivated aborts look like this:

```
message failed with backtrace:
00: f010000 (method 1) -- actor aborted with code 1 (9)
--> caused by: ipld::open -- failed to parse cid: CIDv0 requires a DagPB codec (1: illegal argument)
```

Whereas fatal error backtraces display the error chain from anyhow, as well as the backtrace if captured:

```
fatal backtrace: message failed with backtrace:
00: f010001 (method 1) -- fatal error (10)
--> caused by: [FATAL] Error: [from=f1d3nehuc4u3l5mn7hazppnogf3oe6l6ymaicbkhi, to=f010001, seq=1, m=1, h=0]: missing state: baeaikaia, Stacktrace:
   0: std::backtrace_rs::backtrace::libunwind::trace
             at /rustc/cb121987158d69bb894ba1bcc21dc45d1e0a488f/library/std/src/../../backtrace/src/backtrace/libunwind.rs:93:5
   1: std::backtrace_rs::backtrace::trace_unsynchronized
             at /rustc/cb121987158d69bb894ba1bcc21dc45d1e0a488f/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2: std::backtrace::Backtrace::create
             at /rustc/cb121987158d69bb894ba1bcc21dc45d1e0a488f/library/std/src/backtrace.rs:328:13
   3: std::backtrace::Backtrace::capture
             at /rustc/cb121987158d69bb894ba1bcc21dc45d1e0a488f/library/std/src/backtrace.rs:296:9
   4: anyhow::error::<impl anyhow::Error>::msg
             at /Users/raul/.cargo/registry/src/github.com-1ecc6299db9ec823/anyhow-1.0.56/src/error.rs:81:36
   5: <fvm::kernel::default::DefaultKernel<C> as fvm::kernel::BlockOps>::block_open::{{closure}}
             at /Users/raul/W/pl/fvm/fvm/src/kernel/default.rs:304:28
   6: core::option::Option<T>::ok_or_else
             at /rustc/cb121987158d69bb894ba1bcc21dc45d1e0a488f/library/core/src/option.rs:1067:25
   7: <fvm::kernel::default::DefaultKernel<C> as fvm::kernel::BlockOps>::block_open
             at /Users/raul/W/pl/fvm/fvm/src/kernel/default.rs:299:20
   8: fvm::syscalls::ipld::open
             at /Users/raul/W/pl/fvm/fvm/src/syscalls/ipld.rs:10:22
   9: core::ops::function::Fn::call
             at /rustc/cb121987158d69bb894ba1bcc21dc45d1e0a488f/library/core/src/ops/function.rs:77:5
  10: <wasmtime::linker::Linker<fvm::syscalls::InvocationData<K>> as fvm::syscalls::bind::BindSyscall<(A,),Ret,Func>>::bind::{{closure}}
             at /Users/raul/W/pl/fvm/fvm/src/syscalls/bind.rs:171:44
  ...
```

### DIFF
--- a/fvm/src/call_manager/backtrace.rs
+++ b/fvm/src/call_manager/backtrace.rs
@@ -98,72 +98,70 @@ impl Display for Frame {
     }
 }
 
-#[derive(Clone, Debug)]
-pub struct SyscallCause {
-    /// The syscall "module".
-    pub module: &'static str,
-    /// The syscall function name.
-    pub function: &'static str,
-    /// The exact syscall error.
-    pub error: ErrorNumber,
-    /// The informational syscall message.
-    pub message: String,
-}
-
-#[derive(Clone, Debug)]
-pub struct FatalCause {
-    /// The alternate-formatted message from the anyhow error.
-    pub error_msg: String,
-    /// The backtrace, captured if the relevant
-    /// [environment variables](https://doc.rust-lang.org/std/backtrace/index.html#environment-variables) are enabled.
-    pub backtrace: String,
-}
-
 /// The ultimate "cause" of a failed message.
 #[derive(Clone, Debug)]
 pub enum Cause {
     /// The original cause was a syscall error.
-    Syscall(SyscallCause),
+    Syscall {
+        /// The syscall "module".
+        module: &'static str,
+        /// The syscall function name.
+        function: &'static str,
+        /// The exact syscall error.
+        error: ErrorNumber,
+        /// The informational syscall message.
+        message: String,
+    },
     /// The original cause was a fatal error.
-    Fatal(FatalCause),
+    Fatal {
+        /// The alternate-formatted message from the anyhow error.
+        error_msg: String,
+        /// The backtrace, captured if the relevant
+        /// [environment variables](https://doc.rust-lang.org/std/backtrace/index.html#environment-variables) are enabled.
+        backtrace: String,
+    },
 }
 
 impl Cause {
     /// Records a failing syscall as the cause of a backtrace.
     pub fn from_syscall(module: &'static str, function: &'static str, err: SyscallError) -> Self {
-        Self::Syscall(SyscallCause {
+        Self::Syscall {
             module,
             function,
             error: err.1,
             message: err.0,
-        })
+        }
     }
 
     /// Records a fatal error as the cause of a backtrace.
     pub fn from_fatal(err: anyhow::Error) -> Self {
-        Self::Fatal(FatalCause {
+        Self::Fatal {
             error_msg: format!("{:#}", err),
             backtrace: err.backtrace().to_string(),
-        })
+        }
     }
 }
 
 impl Display for Cause {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
         match self {
-            Cause::Syscall(cause) => {
+            Cause::Syscall {
+                module,
+                function,
+                error,
+                message,
+            } => {
                 write!(
                     f,
                     "{}::{} -- {} ({}: {})",
-                    cause.module, cause.function, &cause.message, cause.error as u32, cause.error,
+                    module, function, &message, *error as u32, error,
                 )
             }
-            Cause::Fatal(msg) => {
-                write!(
-                    f,
-                    "[FATAL] Error: {}, Backtrace:\n{}",
-                    msg.error_msg, msg.backtrace
-                )
+            Cause::Fatal {
+                error_msg,
+                backtrace,
+            } => {
+                write!(f, "[FATAL] Error: {}, Backtrace:\n{}", error_msg, backtrace)
             }
         }
     }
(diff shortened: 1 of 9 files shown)


fvm/src/call_manager/backtrace.rs
```diff
@@ -34,22 +36,35 @@ impl Backtrace {
...
 
-    /// Set the backtrace cause. If there is an existing backtrace, this will clear it.
-    pub fn set_cause(&mut self, cause: Cause) {
+    /// Begins a new backtrace. If there is an existing backtrace, this will clear it.
+    ///
+    /// Note: Backtraces are populated _backwards_. That is, a frame is inserted
+    /// every time an actor returns. That's why `begin()` resets any currently
+    /// accumulated state, as once an error occurs, we want to track its
+    /// propagation all the way up.
+    pub fn begin(&mut self, cause: Cause) {
```
Stebalien, line 51: So much nicer.

fvm/src/call_manager/backtrace.rs
```diff
@@ -96,23 +110,61 @@ pub struct Cause {
...
+
+/// The ultimate "cause" of a failed message.
+#[derive(Clone, Debug)]
+pub enum Cause {
+    /// The original cause was a syscall error.
+    Syscall(SyscallCause),
+    /// The original cause was a fatal error.
+    Fatal(FatalCause),
+}
+
```
Stebalien, line 130: We can collapse these into a single enum.

```rust
/// The ultimate "cause" of a failed message.
#[derive(Clone, Debug)]
pub enum Cause {
    /// The original cause was a syscall error.
    Syscall {
        module: &'static str,
        /// The syscall function name.
        function: &'static str,
        /// The exact syscall error.
        error: ErrorNumber,
        /// The informational syscall message.
        message: String,
    },
    /// The original cause was a fatal error.
    Fatal {
        /// The alternate-formatted message from the anyhow error.
        error_msg: String,
        /// The backtrace, captured if the relevant
        /// [environment variables](https://doc.rust-lang.org/std/backtrace/index.html#environment-variables) are enabled.
        backtrace: String,
    },
}
```

Stebalien (approved):
One nit, but feel free to ignore it.

fvm/src/call_manager/backtrace.rs
raulk: I had inline structs in a previous version, but promoted them to types so I could annotate them with `derive` macros. I think I don't need those now, so I can demote them again.

Stebalien (approved):

arajasek (commented):
LGTM

PR #549, opened 2022-05-12
testing/integration: allow manually-defined accounts

hunjixin:
go-fvm-sdk is using this tester for unit testing, but here I hope to be able to add accounts as expected, so that it is convenient to make assertions in unit test.

### DIFF
--- a/testing/integration/src/tester.rs
+++ b/testing/integration/src/tester.rs
@@ -220,7 +220,7 @@ where
         let state_tree = self
             .state_tree
             .as_mut()
-            .ok_or(anyhow!("unable get state tree"))?;
+            .ok_or_else(||anyhow!("unable get state tree"))?;
         let assigned_addr = state_tree.register_new_address(&pub_key_addr).unwrap();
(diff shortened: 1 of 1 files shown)


testing/integration/src/tester.rs
```diff
@@ -194,23 +207,20 @@ where
...
-    use rand::SeedableRng;
-
-    let rng = &mut rand_chacha::ChaCha8Rng::seed_from_u64(8);
-
-    let mut ret: [Account; N] = [(0, Address::default()); N];
-    for account in ret.iter_mut().take(N) {
-        let priv_key = SecretKey::random(rng);
+
+    /// Put account with specified private key and balance
+    pub fn put_secp256k1_accounts(
```
raulk, line 212:
```suggestion
    pub fn make_secp256k1_account(
```

raulk (approved):
LGTM, but could you please change the method name to make it singular (and let's call it make...)?

raulk:
Thanks for the contribution, @hunjixin! 🙌

PR #551, opened 2022-05-12
fix: kernel: ipld::read not using offset correctly.

### LINKED ISSUE #542: ipld::read does not use offset properly
Given a block id, [ipld::read](https://github.com/filecoin-project/ref-fvm/blob/17072f1db3602021a8e66e7081cd51f4734d78a9/sdk/src/sys/ipld.rs#L75) reads a block from the kernel’s BlockRegistry into a provided buffer. ipld::read also accepts an offset into the block where the read should begin.

offset is passed into [Kernel::block_read](https://github.com/filecoin-project/ref-fvm/blob/297a7694ff9a584880632dde8826d9134d976fd9/fvm/src/kernel/default.rs#L365), which first calculates the length of the read to perform before copying block data to the output buffer:

```
let len = if offset as usize >= data.len() {
  0
} else {
  buf.len().min(data.len())
};

// ...

if len != 0 {
  buf.copy_from_slice(&data[offset as usize..][..len]);
}

```
Neither operation correctly handles nonzero offsets:

- The read length should be `buf.len().min(data.len() - offset)`
- The copy operation should be `buf[..len].copy_from_slice(...)`

raulk:
Fixes #542.

### DIFF
--- a/fvm/src/kernel/default.rs
+++ b/fvm/src/kernel/default.rs
@@ -364,18 +364,20 @@ where
 
     fn block_read(&mut self, id: BlockId, offset: u32, buf: &mut [u8]) -> Result<u32> {
         let data = self.blocks.get(id).or_illegal_argument()?.data();
+        let data_len = data.len();
 
-        let len = if offset as usize >= data.len() {
-            0
-        } else {
-            buf.len().min(data.len())
-        };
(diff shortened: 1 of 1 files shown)


raulk:
Superseded by #555.

PR #553, opened 2022-05-12
sdk: reinstate the message package for message accessors.

raulk:
They make more sense here than in the `vm` package.


raulk:
Going to self-merge as this is non-controversial, and we can review ultimately during the merge to master.

PR #554, opened 2022-05-12
Better stack limits

### LINKED ISSUE #552: Determine stack size limit approach
We need to examine what happens as the recursive stack grows. Today, the only case of this is through multisigs, and we have a stack limit of 4096 (+/- 1). We need to check:

- whether that works (can write a test, or do so manually, or both)
- whether we wanna lower that limit
- whether any config params need to be changed to accommodate.

magik6k:
todo:
* [x] Measure max stack wasm use on mainnet
* [x] Set lower recursive send limit
* [x] Set lower wasm stack limit
* [x] (maybe) Make native stacks larger

Fixes #552

### DIFF
--- a/fvm/src/executor/default.rs
+++ b/fvm/src/executor/default.rs
@@ -11,6 +11,7 @@ use fvm_shared::error::{ErrorNumber, ExitCode};
 use fvm_shared::message::Message;
 use fvm_shared::receipt::Receipt;
 use fvm_shared::ActorID;
+use lazy_static::lazy_static;
 use num_traits::Zero;
 
 use super::{ApplyFailure, ApplyKind, ApplyRet, Executor};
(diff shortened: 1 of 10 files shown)


fvm/src/executor/default.rs
```diff
@@ -38,9 +39,24 @@ impl<K: Kernel> DerefMut for DefaultExecutor<K> {
     }
 }
 
+lazy_static! {
+    static ref EXEC_POOL: yastl::Pool = yastl::Pool::with_config(
+        8,
```
magik6k, line 44: This is an arbitrary number; Just for chain-sync we probably don't need more than 1/2, but some users may want this to be higher e.g. to get internal sends in a bunch of historic tipsets.

raulk: How about using NUM CPUs?

fvm/src/executor/default.rs
```diff
@@ -38,9 +39,24 @@ impl<K: Kernel> DerefMut for DefaultExecutor<K> {
...
+        yastl::ThreadConfig::new()
```
raulk, line 45: Were you able to check if a stack overflow panic is lowered to a fatal error, and in turn to a receipt with SYS_ASSERTION_FAILED?

fvm/src/machine/mod.rs
```diff
@@ -124,8 +124,8 @@ impl NetworkConfig {
     pub fn new(network_version: NetworkVersion) -> Self {
         NetworkConfig {
             network_version,
-            max_call_depth: 4096,
-            max_wasm_stack: 64 * 1024,
+            max_call_depth: 1024,
```
raulk, line 127: To be technically correct, we should set 1024 on nv16 and above, only.

testing/integration/tests/lib.rs
```diff
@@ -68,6 +71,53 @@ fn hello_world() {
...
+        ..Message::default()
+    };
+
+    let res = tester
+        .executor
+        .unwrap()
+        .execute_message(message, ApplyKind::Explicit, 100)
+        .unwrap();
+
+    assert_eq!(res.msg_receipt.exit_code.value(), 0)
```
raulk, line 118: Is the exit code you see here 10?

raulk (commented):
Looks good! Just a few questions for now. Review from @Kubuxu @dignifiedquire @Stebalien would be welcome.

fvm/src/executor/default.rs
magik6k: Added a case to the integration test checking that; You'll see illegal_instruction when you run out of stack in wasm, because the stack limiter will call `unreachable` when that happens. If you're asking about some other stack overflow - no other stack overflows can happen with the changes in this PR

testing/integration/tests/lib.rs
magik6k: Added more cases with more readable codes.

fvm/src/machine/mod.rs
Stebalien: So, @magik6k pointed out that we'll hit the rust stack limit at 4096. The only real solution would be to increase the thread size.

But, in practice, it's unclear if one can actually _make_ a message that can get past 1024 calls on mainnet. By "unclear", I mean "totally not close to possible". @magik6k  only got to 1025 by making a custom actor that did nothing but recursively send. On mainnet, one would have to make a recursive call in a multisig actor, where each recursive call would be significantly more expensive (running out of block gas before anything else).

Stebalien: Tradeoffs:

- The current way, it's unlikely that something goes wrong, and the worst-case is a fork (for anyone running with the FVM_EXPERIMENTAL flag).
- By increasing to 4096 (without changing the stack size), it's also unlikely that something will go wrong, but we'll crash (hard) if we do go past 1024.

fvm/src/executor/default.rs
Stebalien: Some people have a _lot_ of CPUs. We're already reserving half a gig for stacks here. We'll actually likely want to shrink this.

Stebalien (approved):

fvm/src/executor/default.rs
Stebalien: I.e., 128 cores = 8GiB of memory just sitting there.

PR #555, opened 2022-05-13
A series of syscall-related cleanups

Stebalien:
This change-set includes 6 patches that should be reviewed (mostly) independently. They're bundled because they build on each other (at least to an extent).

Please see the commit messages for detailed descriptions. I don't feel too strongly about most of these changes, so I'm happy to punt anything we feel we can do later and/or just isn't necessary.


fvm/src/call_manager/backtrace.rs
```diff
@@ -77,8 +76,6 @@ pub struct Frame {
     pub source: ActorID,
     /// The method that was invoked.
     pub method: MethodNum,
-    /// The parameters passed to this method.
-    pub params: RawBytes,
```
Stebalien, line 81: This could have gone in a separate PR, but it makes the send/invoke refactor patch a bit simpler (no need to clone the params).

shared/src/error/mod.rs
```diff
@@ -114,6 +114,7 @@ impl ExitCode {
 /// When a syscall fails, it returns an `ErrorNumber` to indicate why. The syscalls themselves
 /// include documentation on _which_ syscall errors they can be expected to return, and what they
 /// mean in the context of the syscall.
+#[non_exhaustive]
```
Stebalien, line 117: We don't _have_ to do this, but it's easier to make this "extensible" now than it is to do that later.

fvm/src/externs/mod.rs
```diff
@@ -23,7 +21,7 @@ pub trait Rand {
     /// ChainEpoch, Entropy from the ticket chain.
     fn get_chain_randomness(
         &self,
-        pers: DomainSeparationTag,
```
Stebalien, line 26: This is not consensus critical, but it has been on my mind for a while.

1. Removes unnecessary error cases from the kernel.
2. Paves the way for user actors (that won't care about _specific_ tags).

fvm/src/kernel/mod.rs
```diff
@@ -180,7 +180,7 @@ pub trait ActorOps {
     fn create_actor(&mut self, code_cid: Cid, actor_id: ActorID) -> Result<()>;
 
     /// Returns whether the supplied code_cid belongs to a known built-in actor type.
-    fn resolve_builtin_actor_type(&self, code_cid: &Cid) -> Option<actor::builtin::Type>;
+    fn get_builtin_actor_type(&self, code_cid: &Cid) -> Option<actor::builtin::Type>;
```
Stebalien, line 177: I'm happy to revert this. We had two get and two resolve functions in this module, and I was trying to define a pattern.

fvm/src/syscalls/actor.rs
```diff
@@ -10,32 +10,30 @@ pub fn resolve_address(
...
     let addr = context.memory.read_address(addr_off, addr_len)?;
-    let (resolved, value) = match context.kernel.resolve_address(&addr)? {
-        Some(id) => (0, id),
-        None => (-1, 0),
-    };
-    Ok(sys::out::actor::ResolveAddress { resolved, value })
+    let actor_id = context
+        .kernel
+        .resolve_address(&addr)?
+        .ok_or_else(|| syscall_error!(NotFound; "actor not found"))?;
```
Stebalien, line 18: Returns an error on resolution failure, not a -1.

fvm/src/syscalls/context.rs
```diff
@@ -81,6 +83,21 @@ impl Memory {
         .context("failed to parse cid")
     }
 
+    pub fn write_cid(&mut self, k: &Cid, offset: u32, len: u32) -> Result<u32> {
+        let out = self.try_slice_mut(offset, len)?;
+
+        let mut buf = Cursor::new([0u8; MAX_CID_LEN]);
+        // At the moment, all CIDs are gauranteed to fit in 100 bytes (statically) because the max
```
Stebalien, line 90: _Do_ we want to bump that to 256? We can do that in a followup patch.

fvm/src/syscalls/context.rs
```diff
@@ -81,6 +83,21 @@ impl Memory {
         .context("failed to parse cid")
     }
 
+    pub fn write_cid(&mut self, k: &Cid, offset: u32, len: u32) -> Result<u32> {
+        let out = self.try_slice_mut(offset, len)?;
+
+        let mut buf = Cursor::new([0u8; MAX_CID_LEN]);
```
Stebalien, line 89: Writing then copying is likely no slower than pre-computing the size (as we were doing before). However, this is much less likely to have a bug.

fvm/src/syscalls/crypto.rs
```diff
@@ -90,25 +90,21 @@ pub fn compute_unsealed_sector_cid(
     cid_off: u32,
     cid_len: u32,
 ) -> Result<u32> {
+    // Check/read all arguments.
+    let typ = RegisteredSealProof::from(proof_type);
+    if let RegisteredSealProof::Invalid(invalid) = typ {
+        return Err(syscall_error!(IllegalArgument; "invalid proof type {}", invalid).into());
+    }
```
Stebalien, line 97: fixing a todo

fvm/src/kernel/blocks.rs
```diff
@@ -14,6 +17,7 @@ pub(crate) struct BlockRegistry {
 pub type BlockId = u32;
 
 const FIRST_ID: BlockId = 1;
+const MAX_BLOCKS: u32 = i32::MAX as u32; // TODO: Limit
```
Stebalien, line 21: We can set a limit in M2.

fvm/src/kernel/blocks.rs
```diff
@@ -84,41 +111,46 @@ impl BlockRegistry {
...
 impl BlockRegistry {
     /// Adds a new block to the registry, and returns a handle to refer to it.
-    pub fn put(&mut self, block: Block) -> Result<BlockId, BlockError> {
-        // TODO: limit the code types we allow.
-        let mut id: u32 = self
-            .blocks
-            .len()
-            .try_into()
-            .map_err(|_| BlockError::TooManyBlocks)?;
-        id += FIRST_ID;
```
Stebalien, line 94: This could overflow (technically...).

fvm/src/kernel/blocks.rs
```diff
@@ -84,41 +111,46 @@ impl BlockRegistry {
...
+    pub fn put(&mut self, block: Block) -> Result<BlockId, BlockPutError> {
+        if self.is_full() {
+            return Err(BlockPutError::TooManyBlocks);
+        }
+        if block.codec != DAG_CBOR {
```
Stebalien, line 120: Fixing the TODO, we only put cbor.

fvm/src/kernel/blocks.rs
```diff
@@ -61,19 +65,42 @@ impl Block {
 }
 
 #[derive(Error, Debug)]
-pub enum BlockError {
-    #[error("block {0} is unreachable")]
-    Unreachable(Box<Cid>),
+pub enum BlockPutError {
```
Stebalien, line 70: I'm tempted to just use syscall errors, but technically this isn't a part of the syscall interface.

sdk/src/ipld.rs
```diff
@@ -32,24 +32,38 @@ pub fn get(cid: &Cid) -> SyscallResult<Vec<u8>> {
...
-/// Gets the data of the block referenced by BlockId. If the caller knows the
-/// size, this function will avoid statting the block.
-pub fn get_block(id: fvm_shared::sys::BlockId, size: Option<u32>) -> SyscallResult<Vec<u8>> {
-    let size = match size {
-        Some(size) => size,
-        None => unsafe { sys::ipld::stat(id).map(|out| out.size)? },
-    };
-    let mut block = Vec::with_capacity(size as usize);
+/// Gets the data of the block referenced by BlockId. If the caller knows the size, this function
+/// will read the block in syscall. Otherwise, any block over 1KiB will take two syscalls.
```
raulk, line 43:
```suggestion
/// will read the block in a single syscall. Otherwise, any block over 1KiB will take two syscalls.
```

fvm/src/call_manager/default.rs
```diff
@@ -349,7 +353,7 @@ where
                 store.data_mut().memory = memory;
 
                 // Lookup the invoke method.
-                let invoke: wasmtime::TypedFunc<(u32,), u32> = instance
+                let invoke: wasmtime::TypedFunc<(u32, u64, u32), u32> = instance
```
raulk, line 356: Ouch, this is a massive breaking change for the pure sake of optimization. The reason why this is an _optimization_ is that this data can be otherwise obtained through a syscall.

If we are OK augmenting the invoke signature, the method number, caller, receiver, value received, etc. should totally take priority (ignoring momentarily that the method number is part of a _specific_ call convention, so it should technically not be an argument to invoke) over parameter metadata which may even be optional (such as if the method does not require params).

In a nutshell,. I really dislike this, but could stand behind putting these attributes in `InvocationContext`, fetcheable through a syscall that we know every single actor is going to make.

fvm/src/syscalls/mod.rs
```diff
@@ -92,7 +92,7 @@ use self::bind::BindSyscall;
 use self::error::Abort;
 
 /// The maximum supported CID size. (SPEC_AUDIT)
-pub const MAX_CID_LEN: usize = 100;
+pub const MAX_CID_LEN: usize = 256;
```
raulk, line 95: I guess we're doing this now, although we don't have a clear case for it?

raulk (commented):
Main pushback is the change in signature of the invoke entrypoint for actors.

shared/src/error/mod.rs
Stebalien: Removing the `non_exhaustive` later would just mean that users can now exhaustively match, it won't affect existing code.

fvm/src/externs/mod.rs
Stebalien: > This also means that the type (DomainSeparationTag) can move from fvm_shared to fvm_sdk, or directly to the built-in actors runtime (this option is probably better to avoid confusion).

I've removed the type from the FVM entirely.

fvm/src/call_manager/default.rs
Stebalien: Reverted.

fvm/src/externs/mod.rs
raulk: Nah, not a problem.

fvm/src/syscalls/actor.rs
raulk: I mean, just commenting that we're masking a potential NotFound from the kernel, but that it currently can't produce it. Definitely an uber-nit, not important.

fvm/src/kernel/blocks.rs
raulk: Ahh, GitHub diff view...

raulk (approved):
Thanks, this cleanup is great! 💅

fvm/src/kernel/mod.rs
```diff
@@ -166,8 +166,8 @@ pub trait ActorOps {
     /// If the argument is an ID address it is returned directly.
     fn resolve_address(&self, address: &Address) -> Result<Option<ActorID>>;
 
-    /// Look up the code ID at an actor address.
-    fn get_actor_code_cid(&self, addr: &Address) -> Result<Option<Cid>>;
+    /// Look up the code CID of an actor.
+    fn get_actor_code_cid(&self, id: ActorID) -> Result<Option<Cid>>;
```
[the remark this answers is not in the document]
anorth: I've already been getting the built-ins changed to use IDs even tho the syscall was address

PR #533, opened 2022-05-05
merge syscall changes to master

raulk:
Includes:
- https://github.com/filecoin-project/ref-fvm/pull/516
- https://github.com/filecoin-project/ref-fvm/pull/515
- https://github.com/filecoin-project/ref-fvm/pull/509
- https://github.com/filecoin-project/ref-fvm/issues/521 (name consistency only, no name changes)
- https://github.com/filecoin-project/ref-fvm/pull/555

Fixes https://github.com/filecoin-project/ref-fvm/issues/542.
Fixes https://github.com/filecoin-project/ref-fvm/issues/453.
Fixes https://github.com/filecoin-project/fvm-specs/issues/85.
Fixes https://github.com/filecoin-project/ref-fvm/issues/506 (items 1-3, but good enough to close that issue).

PR #556, opened 2022-05-13
remove support for nv14 and actors v6.

raulk:
When this repo was started, the latest network version on mainnet was nv14. In order to run canaries against mainnet (M0 and M0.5 milestones), we added support for nv14. Now that mainnet is on nv15 and actors v7, we can remove support for nv14. Reasons:

- We don't intend to maintain actors v6 under builtin-actors.
- More specifically, actors v6 won't be updated to absorb the syscall changes being introduced in #533.
- We can shed unneeded code by making the deprecation official.


Stebalien (approved):

Stebalien:
Merging before anyone has a chance to tell me why I shouldn't... :heart:.

PR #557, opened 2022-05-13
catch panics inside Wasm and lower to fatal errors.

raulk:
Catches panics when invoking Wasm and transforms them into fatal errors. I'm not sure about the safety properties of this, and if this has the chance of poisoning the machine. It doesn't appear like it will.

This also adds an integration test to induce the behaviour.


Stebalien (approved):
This seems reasonable. We'll probably want to catch panics in other places as well (crypto, decoding, etc.) but this is a good place to start.

Stebalien:
Blocked on resolving https://github.com/filecoin-project/ref-fvm/pull/563.

PR #561, opened 2022-05-15
[wip] feat: Maximum stack size accounting

magik6k:
Depends on https://github.com/filecoin-project/fvm-wasm-instrument/pull/5

TBD if we want to keep this, it's probably useful for debugging; If we want this, we will need to:
* [ ] Put this behind a feature flag
* [ ] Use updated wasm-instrument

### DIFF
--- a/fvm/Cargo.toml
+++ b/fvm/Cargo.toml
@@ -39,7 +39,7 @@ log = "0.4.14"
 byteorder = "1.4.3"
 anymap = "0.12.1"
 blake2b_simd = "1.0.0"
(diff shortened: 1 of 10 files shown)


raulk (commented):
@magik6k If I understand correctly, this is the instrumentation we use to identify the length of the Wasm stacks in existing code, right? If so, I agree that this should be behind a feature flag, and it's not necessary for M1.

Stebalien:
Everything relevant in this PR has been merged.

PR #562, opened 2022-05-15
Strongly typed gas units, and fix two related bugs

Stebalien:
Motivation: we had two bugs in gas units (milligas versus gas):

1. Consensus fault charges from lotus (only applies to nv15).
2. Window post scaling charge.

This fixes those bugs, and adds a new "safe" gas type that lets us automatically avoid:

1. Multiplying gas by gas. Only `Gas * (i32/i64)` is implemented.
2. Adding gas to non-gas. Only `Gas + Gas` is implemented.
3. Swapping gas/milligas.
4. Not using saturating arithmetic. All gas operations use saturating arithmetic internally.

The automatic saturating arithmetic also makes the code significantly cleaner.


magik6k (approved):
Looks good, one nitpick

PR #563, opened 2022-05-15
refactor: split threaded executor into a new struct

Stebalien:
We only _really_ need this executor when the caller doesn't guarantee the stack size.

Unfortunately, it can make testing hard because it means that the `Machine` needs to be `Send`. This breaks #557 which contains `Rc`s and `RefCell`s.

### DIFF
--- a/fvm/src/executor/default.rs
+++ b/fvm/src/executor/default.rs
@@ -11,7 +11,6 @@ use fvm_shared::error::{ErrorNumber, ExitCode};
 use fvm_shared::message::Message;
 use fvm_shared::receipt::Receipt;
 use fvm_shared::ActorID;
-use lazy_static::lazy_static;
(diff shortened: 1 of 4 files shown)


Stebalien:
Alternatives:

- Change #557. But honestly, I'd rather not use the whole threaded executor thing when testing anyways.
- Keep the default executor as "threaded", but split out a base executor.

magik6k (approved):
Looks good; would be good to have a matching PR in filecoin-ffi before merging this so that we don't forget about changing the executor there.

raulk (approved):
nit: Not sure `ThreadedExecutor` is the name I'd choose given that the main purpose of this it's actually to enable deeper stacks, and the threading aspect is just the means to an end. But I can live with this.

PR #574, opened 2022-05-17
catch additional panics

Stebalien:
1. Catch panics when decoding cbor from actors.
2. Catch panics when verifying proofs, etc.

The only strictly necessary change is the window post one, as a panic there could make a proof indisputable. But there's no real _issue_ with being careful and turning all of these panics into "illegal argument" errors as they're all well isolated.

### DIFF
--- a/fvm/src/kernel/default.rs
+++ b/fvm/src/kernel/default.rs
@@ -1,22 +1,16 @@
 use std::collections::BTreeMap;
 use std::convert::{TryFrom, TryInto};
+use std::panic::{self, UnwindSafe};
 
 use anyhow::{anyhow, Context as _};
 use byteorder::{BigEndian, WriteBytesExt};
 use cid::Cid;
-use filecoin_proofs_api::seal::{
-    compute_comm_d, verify_aggregate_seal_commit_proofs, verify_seal as proofs_verify_seal,
-};
-use filecoin_proofs_api::update::verify_empty_sector_update_proof;
-use filecoin_proofs_api::{self as proofs, post, seal, ProverId, PublicReplicaInfo, SectorId};
+use filecoin_proofs_api::{self as proofs, ProverId, PublicReplicaInfo, SectorId};
 use fvm_ipld_blockstore::Blockstore;
 use fvm_ipld_encoding::{bytes_32, from_slice, to_vec};
 use fvm_shared::actor::builtin::Type;
 use fvm_shared::address::Protocol;
 use fvm_shared::bigint::{BigInt, Zero};
-use fvm_shared::commcid::{
-    cid_to_data_commitment_v1, cid_to_replica_commitment_v1, data_commitment_v1_to_cid,
-};
 use fvm_shared::consensus::ConsensusFault;
 use fvm_shared::crypto::signature;
 use fvm_shared::econ::TokenAmount;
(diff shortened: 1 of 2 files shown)


fvm/src/kernel/default.rs
```diff
@@ -1,30 +1,24 @@
...
 
 use anyhow::{anyhow, Context as _};
 use byteorder::{BigEndian, WriteBytesExt};
 use cid::Cid;
-use filecoin_proofs_api::seal::{
-    compute_comm_d, verify_aggregate_seal_commit_proofs, verify_seal as proofs_verify_seal,
-};
-use filecoin_proofs_api::update::verify_empty_sector_update_proof;
-use filecoin_proofs_api::{self as proofs, post, seal, ProverId, PublicReplicaInfo, SectorId};
+use filecoin_proofs_api::{self as proofs, ProverId, PublicReplicaInfo, SectorId};
```
Stebalien, line 8: I started qualifying these calls because distinguishing between "our functions" and "not our functions" was getting difficult.

fvm/src/kernel/default.rs
```diff
@@ -990,3 +869,168 @@ fn verify_seal(vi: &SealVerifyInfo) -> Result<bool> {
     // Worst case, _some_ node falls out of sync. Better than the network halting.
     .context("failed to verify seal proof")
 }
+
+fn verify_post(verify_info: &WindowPoStVerifyInfo) -> Result<bool> {
```
Stebalien, line 873: Copied from above.

fvm/src/kernel/default.rs
```diff
@@ -990,3 +869,168 @@ fn verify_seal(vi: &SealVerifyInfo) -> Result<bool> {
...
+
+    // Generate prover bytes from ID
+    let prover_id = prover_id_from_u64(*prover);
+
+    // Verify Proof
+    proofs::post::verify_window_post(&bytes_32(&randomness), &proofs, &replicas, prover_id)
+        .or_illegal_argument()
+}
+
+fn verify_aggregate_seals(aggregate: &AggregateSealVerifyProofAndInfos) -> Result<bool> {
```
Stebalien, line 904: Copied from above.

fvm/src/kernel/default.rs
```diff
@@ -990,3 +869,168 @@ fn verify_seal(vi: &SealVerifyInfo) -> Result<bool> {
...
+        aggregate.aggregate_proof.try_into().or_illegal_argument()?,
+        aggregate.proof.clone(),
+        &commrs,
+        &seeds,
+        inp,
+    )
+    .or_illegal_argument()
+}
+
+fn verify_replica_update(replica: &ReplicaUpdateInfo) -> Result<bool> {
```
Stebalien, line 969: Copied from above.

fvm/src/syscalls/context.rs
```diff
@@ -105,7 +106,14 @@ impl Memory {
 
     pub fn read_cbor<T: Cbor>(&self, offset: u32, len: u32) -> Result<T> {
         let bytes = self.try_slice(offset, len)?;
-        from_slice(bytes).or_error(ErrorNumber::IllegalArgument)
+        // Catch panics when decoding cbor from actors, _just_ in case.
+        match panic::catch_unwind(|| from_slice(bytes).or_error(ErrorNumber::IllegalArgument)) {
+            Ok(v) => v,
+            Err(e) => {
+                log::error!("panic when decoding cbor from actor: {:?}", e);
+                Err(syscall_error!(IllegalArgument; "panic when decoding cbor from actor").into())
```
raulk, line 114: I would suggest `ErrorNumber::Serialization` here, but `read_cbor` is only used for parsing syscall parameters, so `ErrorNumber::IllegalArgument` is probably a safe assumption to make.

raulk (approved):

PR #572, opened 2022-05-17
chore: update dependencies

Stebalien:

PR #575, opened 2022-05-17
chore: build test dependencies with optimizations enabled

Stebalien:


Stebalien:
Ok, it looks like updating wasmtime actually addresses this. Closing in favor of #572.

PR #576, opened 2022-05-18
chore: audit errors

Stebalien:
1. If we fail to create an account due to a serialization error, treat it as an illegal argument and move on rather than failing.
2. Return NotFound if we fail to find an actor when resolving an address to a key.
3. Return NotFound (as documented) if the beneficiary doesn't exist on delete.
4. Return Forbidden (as documented) if the beneficiary is self on delete.
5. Remove nv14 support from `transfer`, and simplify some error handling.
6. Avoid logging when debugging is enabled.
7. Return Forbidden for various actor creation errors, instead of IllegalArgument.

Unresolved: `resolve_to_key_addr` (called when verifying a signature), returns IllegalArgument if the "signer" address is not an account. The thing is, IllegalArgument usually means that the caller "should have known better" (e.g., they passed a negative number when the docs specified a positive number). In this case, the caller is passing a valid address, it just isn't an address of an account.

### DIFF
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -244,9 +244,15 @@ where
 
         // Now invoke the constructor; first create the parameters, then
         // instantiate a new kernel to invoke the constructor.
-        let params = to_vec(&addr)
-            // TODO(#198) this should be a Sys actor error, but we're copying lotus here.
-            .map_err(|e| syscall_error!(Serialization; "failed to serialize params: {}", e))?;
+        let params = to_vec(&addr).map_err(|e| {
+            // This shouldn't happen, but we treat it as an illegal argument error and move on.
+            // It _likely_ means that the inputs were invalid in some unexpected way.
+            log::error!(
+                "failed to serialize address when creating actor, ignoring: {}",
+                e
+            );
+            syscall_error!(IllegalArgument; "failed to serialize params: {}", e)
+        })?;
 
         self.send_resolved::<K>(
             account_actor::SYSTEM_ACTOR_ID,
@@ -272,7 +278,6 @@ where
         K: Kernel<CallManager = Self>,
     {
         // Get the receiver; this will resolve the address.
-        // TODO: What kind of errors should we be using here?
         let to = match self.state_tree().lookup_id(&to)? {
             Some(addr) => addr,
             None => match to.protocol() {
--- a/fvm/src/kernel/default.rs
+++ b/fvm/src/kernel/default.rs
@@ -114,7 +114,7 @@ where
             .state_tree()
             .get_actor(addr)?
             .context("state tree doesn't contain actor")
-            .or_illegal_argument()?;
+            .or_error(ErrorNumber::NotFound)?;
 
         let is_account = self
             .call_manager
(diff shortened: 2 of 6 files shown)


fvm/src/kernel/default.rs
```diff
@@ -125,6 +125,8 @@ where
             .unwrap_or(false);
 
         if !is_account {
+            // TODO: this is wrong. Maybe some InvalidActor type?
+            // The argument is syntactically correct, but semantically wrong.
```
raulk, line 129: I don't think adding such as specialised error number is useful. It's really just an invalid argument. We might be getting caught up in the nuance between `invalid` and `illegal`. I take this error as meaning "this argument was invalid" without any implication of the cause (invalid syntax, or invalid value). We may want to rename `IllegalArgument` to `InvalidArgument` to open the definition more. (The [spec](https://github.com/filecoin-project/fvm-specs/blob/main/07-errors.md) mixes invalid and illegal).

fvm/src/call_manager/default.rs
```diff
@@ -244,9 +244,15 @@ where
...
-            .map_err(|e| syscall_error!(Serialization; "failed to serialize params: {}", e))?;
+        let params = to_vec(&addr).map_err(|e| {
+            // This shouldn't happen, but we treat it as an illegal argument error and move on.
+            // It _likely_ means that the inputs were invalid in some unexpected way.
+            log::error!(
+                "failed to serialize address when creating actor, ignoring: {}",
+                e
+            );
+            syscall_error!(IllegalArgument; "failed to serialize params: {}", e)
+        })?;
```
raulk, line 255: Can this break nv15 compatibility? I don't follow with exit code we'll abort with.

fvm/src/machine/default.rs
```diff
@@ -193,84 +193,42 @@ where
     }
 
     fn transfer(&mut self, from: ActorID, to: ActorID, value: &TokenAmount) -> Result<()> {
-        if self.context.network_version >= NetworkVersion::V15 {
-            if value.is_negative() {
-                return Err(syscall_error!(IllegalArgument;
+        if value.is_negative() {
```
raulk, line 196: Great catch!

fvm/src/machine/default.rs
```diff
@@ -193,84 +193,42 @@ where
...
-                .state_tree
-                .get_actor_id(to)?
-                .context("cannot transfer to non-existent receiver")
-                .or_error(ErrorNumber::NotFound)?;
-
-            from_actor.deduct_funds(value).map_err(|e| {
-                syscall_error!(InsufficientFunds;
-                           "transfer failed when deducting funds ({}) from balance ({}): {}",
-                           value, &from_actor.balance, e)
-            })?;
```
raulk, line 230: Can we preserve _this_ mesage string inside `deduct_funds`? It's more detailed.

raulk (approved):
A few questions and nits, but approving in anticipation.

fvm/src/kernel/default.rs
Stebalien: I'd like to be able to distinguish between a _bug_ in the actor and a "normal" error. In most cases, `IllegalArgument` means that there's a bug in the actor itself.

But this is hardly the only borderline case. Most of the proof code will shrug and say "eh, illegal argument?".

fvm/src/call_manager/default.rs
Stebalien: This was previously a fatal error that could only be triggered if there was a bug. So, hopefully not?

PR #577, opened 2022-05-18
Unit Tests for DefaultKernel `IpldOps`

mriise:
Extremely rough minimal for now with lots of TODOs, but its compiles :^)


mriise:
Right now this is has some really gross coding, so if you are looking at this commit please brace yourself

fvm/src/machine/mod.rs
```diff
@@ -20,7 +20,7 @@ pub use default::DefaultMachine;
 
 mod engine;
 
-pub use engine::{Engine, MultiEngine};
+pub use engine::{Engine, EngineConfig, MultiEngine};
```
mriise, line 23: This is used [here](https://github.com/filecoin-project/ref-fvm/blob/927394b0be886e4d4685c511fb1d87a96fe35b2f/fvm/src/lib.rs#L255) to give test zero cost wasm executon.

If this was intended to not be public, I can hide this under a cfg for testing only. I would rather have more flexibility with gas prices and usage in tests than using gas prices loaded from the network version.

fvm/src/gas/mod.rs
```diff
@@ -53,37 +53,37 @@ impl Display for Gas {
...
     #[inline]
-    pub fn round_up(&self) -> i64 {
+    pub const fn round_up(&self) -> i64 {
         milligas_to_gas(self.0, true)
     }
 
     /// Returns the gas value as an integer, truncating the fractional part.
     #[inline]
-    pub fn round_down(&self) -> i64 {
+    pub const fn round_down(&self) -> i64 {
```
Stebalien, line 80: :+1:

fvm/src/lib.rs
```diff
@@ -96,7 +101,8 @@ mod test {
             _h2: &[u8],
             _extra: &[u8],
         ) -> anyhow::Result<(Option<fvm_shared::consensus::ConsensusFault>, i64)> {
-            todo!()
+            // consensus is always valid for tests :)
+            anyhow::Result::Ok((None, 0))
```
Stebalien, line 105:
```suggestion
            Ok((None, 0))
```

You only need to use `anyhow::Result::` in a closure where you want rust to guess the return type. Otherwise, you can just use `Ok` and rust will figure out the error type.

fvm/src/machine/mod.rs
Stebalien: It makes sense to make this public.

fvm/tests/default_kernel/dummy.rs
```diff
@@ -0,0 +1,267 @@
...
+use fvm_shared::actor::builtin::Manifest;
+use fvm_shared::address::Address;
+use fvm_shared::state::StateTreeVersion;
+// use fvm_shared::version::NetworkVersion;
+use multihash::Code;
+
+use super::*;
+use crate::DummyExterns;
+
+/// this is essentially identical to DefaultMachine, but has pub fields for reaching inside
```
Stebalien, line 22: If it's nearly identical, this could be a bit annoying to maintain. Any reason not to put these unit tests inside the default kernel module? That'll give them access to private things.

Stebalien: Hm. Although, it doesn't look all that identical.

fvm/tests/default_kernel/dummy.rs
```diff
@@ -0,0 +1,267 @@
...
+    #[deref_mut]
+    pub machine: DummyMachine,
+    pub gas_tracker: GasTracker,
+    pub charge_gas_calls: usize,
+    pub origin: Address,
+    pub nonce: u64,
+}
+/// a wrapper to let us inspect values during testing, all borrows are done with .borrow() or .borrow_mut(), so this should be used in single thereaded tests only
+/// TODO this introduces some rough edges that might need to be cleaned up
+pub struct DummyCallManager(Arc<InnerDummyCallManager>);
```
Stebalien, line 146: Any reason not to use an `Arc<Mutex>` (or even an `Rc<RefCell>`)? That should avoid having to even think about safety, right?

Stebalien (commented):
Some initial comments.

fvm/tests/default_kernel/dummy.rs
mriise: Not really, I was stumbling over `Rc`'s api for too long and decided to go to something I've at least tried once before, I think moving to `Rc<RefCell>` would make most sense here

fvm/src/lib.rs
mriise: Rust at one point was having issues figuring the type out, though it isn't necessary anymore.

fvm/tests/default_kernel/dummy.rs
mriise: it is outside of not using a buffered blockstore, non-parameterized types, non complete transfer fn and create actor fn

mriise: I would argue that `DummyX` structs should stay independent as its supposed to be a testing backboard that (hopefully) unexpected changes in *just the kernel* get caught by tests which means having minimal dependencies on other parts of code. Expected changes should need refactoring, but keeping that as explicit should be correct.

I could move unit tests inside of the default kernel module and make fields `pub(crate)` so I could access values directly, though it would be a fairly hefty file after adding of all the tests for it.

mriise: i could do 
```
kernel
- tests 
  - mod.rs
  - ops.rs
  - x.rs
- mod.rs
- default.rs
```
 
```rust
// in mod.rs
#[cfg(test)]
mod tests;
``` 
```rust
// in default.rs
struct DefaultKernel {
  pub(crate): ...
}
```
which would let me avoid all the dealing with external call manager shenanigans that i'm doing right now, though theoretically with the current setup it would be able to test different kernel implementations with few(er) modifications.

Stebalien: Eh, we're probably going to need to instrument the test machine quite a bit... so we might as well keep it separate. We're not testing it.

fvm/tests/default_kernel/dummy.rs
Stebalien: Ah. Ok, so, what if you _don't_ share the entire call manager. I'm pretty sure you don't need to do that.

Instead, store an `Rc<RefCell<TestData>>` inside the `CallManager`, and keep a reference inside the test itself.

1. The call manager will dereference/borrow this to record test operations, check to see if it needs to do something special at some point, etc.
2. The test-case can safely read/write to this shared object.

This should remove a lot of the complexity.

mriise: well not quite what I am doing currently, but the problem remains where the trait expects a normal reference and `RefCell` will only give out a `Ref`, another option I explored was to keep `Rc` inside of the test and just give the kernel `RefMut<Inner>` though Call manager expects a `'static` lifetime

mriise: Honestly at this point I think I will avoid doing any fancy instrumentation, and use `into_inner()` after doing some operations, this will get a fair amount of coverage by itself and we can add more complex things later.

Stebalien: I'm suggesting:

```rust
struct DummyCallManager {
    field1: Foo,
    field2: Bar,
    the_mutable_stuff: Rc<RefCell<TestData>>,
}
```

You can continue to borrow the outer fields normally, and share/mutate the data in `the_mutable_stuff`.

mriise: ok, so instead of just keeping each field as shared data, let them be normal but just have another field that stores extra data that can be accessed from the outside. That way the trait can get what it expects and we can do whatever inner mutation in addition to be shared externally

Stebalien: Yep. It let's you split out the shared mutable state from the private state.

mriise:
uhhh... that test should be passing (at least it passing on my machine with that command)

Stebalien:
@mriise you need to rebase on master for the latest bug fixes.

mriise:
oh interesting it does a test merge before running things, good CI :)

mriise:
I made the mistake of not using -ff-only the first time around :\

mriise:
@Stebalien should this be left as a valid call?
```rust
let buf = &mut [0u8; 3];
// `id` points to block: "foo".as_bytes()
let diff = kern.block_read(id, 0xFF, buf)?; // offset is larger than total bytes in block
assert_eq!(diff, -255);
```

Stebalien:
Yes, that's correct. _That_ case is a bit strange..., but the expected use is:

```rust
let buf = &mut [0u8; 258];
// `id` points to block: "foo".as_bytes()
let diff = kern.block_read(id, 255, buf)?; // offset is larger than total bytes in block
assert_eq!(diff, -255);
buf[buf.len() + diff]
```

Which would make more sense.

mriise:
alright, then that test would still fail
```rust
let buf = &mut [0u8; 0xFF +  // offset is larger than total bytes in block
        let diff = kern.block_read(id, 0xFF, buf)?; 
        assert_eq!(diff, -0xFF); // failed: got -510

File: fvm/tests/default_kernel/ops.rs [hunk 6]
```diff
@@ -0,0 +1,516 @@
...
+    use fvm::kernel::IpldBlockOps;
+    use fvm::machine::Machine;
+    use fvm_ipld_blockstore::Blockstore;
+    use fvm_ipld_encoding::DAG_CBOR;
+    use fvm_shared::error::ErrorNumber;
+    use multihash::MultihashDigest;
+
+    use super::*;
+
+    // TODO tests for block_open
```
mriise, line 14: I will leave `block_open` tests for another PR since this is already huge

fvm/tests/default_kernel/ops.rs
```diff
@@ -0,0 +1,516 @@
...
+        // make a block
+        let id = kern.block_create(DAG_CBOR, block)?;
+        let id1 = kern1.block_create(DAG_CBOR, "bar".as_bytes())?;
+
+        // TODO are these assumption correct? other ID values could be used although it would be weird
+        assert_eq!(id, 1, "first block id should be 1");
+        assert_eq!(
+            id, id1,
+            "two blocks of the different content but same order should have the same block id"
+        );
```
mriise, line 86: opinions?

Stebalien:
> alright, then that test would still fail

Ah, sorry, you're right. It should have been:


```rust
let buf = &mut [0u8; 258];
// `id` points to block: "foo".as_bytes()
let diff = kern.block_read(id, 255, *buf[255..])?; // offset is larger than total bytes in block
assert_eq!(diff, -255);
buf[buf.len() + diff]
```

fvm/tests/default_kernel/ops.rs
```diff
@@ -0,0 +1,516 @@
...
+                call_manager.gas_tracker.gas_used(),
+                expected_create_price + expected_stat_price,
+                "gas use of creating and 'stat'ing a block does not match price list"
+            )
+        }
+        Ok(())
+    }
+
+    #[test]
+    #[ignore = "TODO should gas be charged for failed stat operations?"]
```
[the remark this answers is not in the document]
mriise: how gas is charged will be changed for M2 and will need to be redone anyway, so gas specific asserts will be removed for now.

fvm/tests/default_kernel/ops.rs
```diff
@@ -0,0 +1,516 @@
...
+        }
+
+        assert_eq!(
+            test_data.borrow().charge_gas_calls,
+            0,
+            "operation failed but charge_gas was called!"
+        );
+
+        // TODO should this be allowed?
+        let _ = kern.block_create(DAG_CBOR, &[])?;
```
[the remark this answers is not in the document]
mriise: while Gas operation is being kicked down the road for M2, whether creating an empty block should be valid is still a question

fvm/tests/default_kernel/dummy.rs
```diff
@@ -0,0 +1,250 @@
+use std::borrow::Borrow;
```
Stebalien, line 1: Can we move this into the root "dummy" module? It seems like it may be useful in other tests?

fvm/tests/default_kernel/dummy.rs
```diff
@@ -0,0 +1,250 @@
...
+    ) -> kernel::Result<InvocationResult> {
+        // Ok(InvocationResult::Return(None));
+        todo!()
+    }
+
+    fn with_transaction(
+        &mut self,
+        _f: impl FnOnce(&mut Self) -> kernel::Result<InvocationResult>,
+    ) -> kernel::Result<InvocationResult> {
+        Ok(InvocationResult::Return(None))
```
Stebalien, line 199: `todo!()`? Unless we expect this to be called.

fvm/tests/fvm.rs
```diff
@@ -0,0 +1,22 @@
+mod default_kernel;
+mod dummy;
+
+use dummy::DummyExterns;
+
+// TODO
```
Stebalien, line 6: remove me?

Stebalien (commented):
Looking good. I have a few questions, but I think we can go ahead and merge this without much discussion and iterate in future PRs.

fvm/tests/default_kernel/dummy.rs
```diff
@@ -0,0 +1,250 @@
...
+
+/// this is essentially identical to DefaultMachine, but has pub fields for reaching inside
+pub struct DummyMachine {
+    pub engine: Engine,
+    pub state_tree: StateTree<MemoryBlockstore>,
+    pub ctx: MachineContext,
+    pub builtin_actors: Manifest,
+}
+
+// hardcoded elsewhere till relavant TODOs are solved
```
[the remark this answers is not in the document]
mriise: line 20 `// TODO should this stay as V15?` which is answered below.

fvm/src/lib.rs
```diff
@@ -49,6 +49,11 @@ lazy_static::lazy_static! {
     };
 }
 
+/// For internal testing, ensures WASM gas charges aren't being added on top of unit test operations.
+pub const __TESTING_FREE_WASM_PRICE: &gas::WasmGasPrices = &gas::WasmGasPrices {
+    exec_instruction_cost: gas::Gas::new(0),
+};
```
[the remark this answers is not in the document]
mriise: field is pub to crate, and I think it should probably stay like that
```rust
pub struct WasmGasPrices {
    pub(crate) exec_instruction_cost: Gas,
}
```

fvm/tests/default_kernel/dummy.rs
```diff
@@ -0,0 +1,250 @@
...
+        // TODO should this stay as V15?
+        let ctx = NetworkConfig::new(fvm_shared::version::NetworkVersion::V15)
+            .override_actors(actors_cid)
+            .for_epoch(0, root);
+
+        Ok(Self {
+            ctx,
+            engine: Engine::new_default(EngineConfig {
+                max_wasm_stack: 1024,
+                wasm_prices: fvm::__TESTING_FREE_WASM_PRICE,
```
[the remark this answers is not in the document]
mriise: Wanted to be sure wasm gas prices were set to 0 for tests, though this doesn't impact things now, I thought it might in the future.

fvm/tests/default_kernel/util.rs
```diff
@@ -0,0 +1,16 @@
+use std::rc::Rc;
```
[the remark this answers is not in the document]
mriise: I will be adding more util functions (and the macro for checking errors) later so I would like to leave it, but it does seem like overkill at the moment

fvm/tests/default_kernel/mod.rs
```diff
@@ -0,0 +1,18 @@
...
+use fvm::kernel::default::DefaultKernel;
+use fvm::kernel::{Block, Kernel};
+use fvm::machine::Machine;
+use multihash::Code;
+use util::*;
+
+type TestingKernel = DefaultKernel<dummy::DummyCallManager>;
+
+// TODO gas functions assert calls are being charged properly
+// TODO maybe make more util functions
```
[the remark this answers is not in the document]
mriise: not for this PR anymore

fvm/tests/default_kernel/mod.rs
```diff
@@ -0,0 +1,18 @@
+mod dummy;
+mod ops;
+mod util;
+
+use std::cell::RefCell;
+
+use dummy::*;
+// test target
```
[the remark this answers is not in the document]
mriise: cheeky comment saying that the below import is the test target for this module :)

fvm/tests/default_kernel/ops.rs
```diff
@@ -0,0 +1,516 @@
...
+            buf,
+            &[0, 0, 0],
+            "offeset went over total length so no data should be read"
+        );
+        // assert!(way_over + 0xFFFF == block.len() as i32);
+        // assert_eq!(
+        //     test_data.borrow().charge_gas_calls,
+        //     0,
+        //     "operation failed but charge_gas was called!"
+        // );
```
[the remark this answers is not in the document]
mriise: Should I add that in documentation in this PR or save it for later?

fvm/tests/default_kernel/ops.rs
```diff
@@ -0,0 +1,516 @@
...
+        let cid1 = kern1.block_link(id1, Code::Blake2b256.into(), 32)?;
+
+        // link a block of different data into kern1
+        let other_id = kern1.block_create(DAG_CBOR, other_block)?;
+        let other_cid = kern1.block_link(other_id, Code::Blake2b256.into(), 32)?;
+
+        let (call_manager, _) = kern.into_inner();
+
+        // CIDs match CIDs generated manually from CID crate
+        let expected_cid = cid::Cid::new_v1(DAG_CBOR, Code::Blake2b256.digest(block).truncate(32));
```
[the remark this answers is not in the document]
mriise: yeah, was just copying the same sequence of operations, will remove.

fvm/tests/default_kernel/ops.rs
```diff
@@ -0,0 +1,516 @@
...
+        let err = kern
+            .block_create(0xFF, block)
+            .expect_err("Returned Ok though invalid codec (0xFF) was used");
+        // TODO should tests be made for returning correct errors like below?
+        match err {
+            fvm::kernel::ExecutionError::Syscall(e) => {
+                assert!(e.1 as u32 == ErrorNumber::IllegalCodec as u32)
+            }
+            _ => panic!("expected a syscall error"),
+        }
```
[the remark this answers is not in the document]
mriise: Kicking down the line for the next PR

mriise:
@Stebalien ready to merge :)

fvm/tests/default_kernel/ops.rs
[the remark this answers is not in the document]
Stebalien: Ah, document `block_read`? We document it in the syscall SDK, but yeah, we should definitely document it on the kernel interface as well.

But yeah, we can just punt that to a new PR.

fvm/src/lib.rs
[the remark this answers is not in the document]
Stebalien: Hm, I'm not sure about that. I expect that users are going to want to configure this (and gas as well).

What if we:

1. Derived `Default` for `PriceList` and `WasmGasPrices`.
2. Marked them as `#[non_exhaustive]`.

This should probably go in a separate PR (make a new PR, we can merge that, then you can rebase and merge this PR), but I really don't want to export a `__TESTING_FREE_WASM_PRICES` object in at the root of the crate _just_ to handle this case.

fvm/tests/default_kernel/dummy.rs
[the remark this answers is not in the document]
Stebalien: It's probably fine if they're not set to 0. We can just provide infinite (effectively) gas.

fvm/tests/default_kernel/util.rs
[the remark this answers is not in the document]
Stebalien: Up to you. My general policy is to _not_ plan for the future like that, because the future rarely turns out as expected.

Also note: `util` crates/modules tend to cause more problems than they solve in my experience.

- At first, put helpers next to the code in question.
- Later, move helpers to some more general location when it becomes clear where else you'll need them. Usually, at that point, it'll be pretty clear.

If you just stick them in a util module, the util module will just grow and grow and grow with random helpers.

fvm/src/lib.rs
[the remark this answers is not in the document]
Stebalien: _Or_ we can just use the defaults.

Stebalien:
The only blocking conversation is https://github.com/filecoin-project/ref-fvm/pull/577#discussion_r883083614 as I'd like to avoid exporting an underscore-underscore-only-for-testing variable in production code.

Stebalien (approved):

Stebalien:
Tests! We have tests!

PR #579, opened 2022-05-19
revert MAX_CID_LEN to 100.

raulk:
See [thread](https://github.com/filecoin-project/ref-fvm/pull/555/commits/ec3086d38601426e9cb219ad4301ac3b6cb6248d#r872509685).


raulk:
Merging because we had already agreed that this change needs to be reverted in the linked thread; it just fell through the cracks.

Stebalien:
Hm. This is a bit concerning because I thought I pushed a commit to revert that...

PR #583, opened 2022-05-19
chore: cleanup todos, warnings, imports, etc.

Stebalien:
Mostly just a "did we forget something" pass.

~Also hides the gas/state_tree modules unless we're "testing" (fixing a todo).~ *edit: forest is using these, and they seem reasonable to export*

Also removes a wasmtime feature we're not using.

### DIFF
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -165,10 +165,10 @@ where
     }
 
     fn finish(mut self) -> (FinishRet, Self::Machine) {
+        // TODO: Having to check against zero here is fishy, but this is what lotus does.
         let gas_used = self.gas_tracker.gas_used().max(Gas::zero()).round_up();
 
         let inner = self.0.take().expect("call manager is poisoned");
-        // TODO: Having to check against zero here is fishy, but this is what lotus does.
         (
             FinishRet {
                 gas_used,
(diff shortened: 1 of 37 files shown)


testing/conformance/src/vm.rs
```diff
@@ -79,6 +79,12 @@ impl TestMachine<Box<DefaultMachine<MemoryBlockstore, TestExterns>>> {
 
         let machine = DefaultMachine::new(&engine, &mc, blockstore, externs).unwrap();
 
+        // Preload the actors. We don't usually preload actors when testing, so we're going to do
```
Stebalien, line 82: Due to the way feature unification works, preloading may or may not be enabled by default (because the testing feature may get enabled by some other crate in the build).

So I'm just explicitly enabling the testing feature for conformance tests (also lets us hide types like the StateTree and Gas).

fvm/src/call_manager/default.rs
```diff
@@ -165,10 +165,10 @@ where
     }
 
     fn finish(mut self) -> (FinishRet, Self::Machine) {
+        // TODO: Having to check against zero here is fishy, but this is what lotus does.
```
arajasek, line 168: I'd be pretty happy to see this turned into an error and fail the message, but...TODO for now...

fvm/src/kernel/mod.rs
```diff
@@ -31,7 +31,13 @@ pub enum SendResult {
     Abort(ExitCode),
 }
 
-/// The "kernel" implements
+/// The "kernel" implements the FVM interface as presented to the actors. It:
+///
+/// - Manage's the Actor's state.
```
arajasek, line 36:
```suggestion
/// - Manages the Actor's state.
```

testing/conformance/src/vm.rs
arajasek: Makes sense to me.

arajasek (approved):

PR #584, opened 2022-05-19
fix: fix read syscall and sdk method

Stebalien:
1. Return the correct offset (relative to the end of the logical buffer, not relative to the end of the bytes read).
2. Don't cast "remaining" to a usize in read_block as it can be negative.
3. Add an exhaustive test.

### DIFF
--- a/sdk/src/ipld.rs
+++ b/sdk/src/ipld.rs
@@ -62,7 +62,13 @@ pub fn get_block(id: fvm_shared::sys::BlockId, size_hint: Option<u32>) -> Syscal
             )?;
             debug_assert!(remaining <= 0, "should have read whole block");
         }
-        buf.set_len(buf.capacity() + (remaining as usize));
+        let size = (buf.capacity() as i64) + (remaining as i64);
+        debug_assert!(size >= 0, "size can't be negative");
+        debug_assert!(
+            size <= buf.capacity() as i64,
+            "size shouldn't exceed capacity"
+        );
+        buf.set_len(size as usize);
     }
     Ok(buf)
 }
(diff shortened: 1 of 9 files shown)


testing/integration/tests/fil-ipld-actor/src/lib.rs
```diff
@@ -0,0 +1,127 @@
+use fvm_ipld_encoding::{to_vec, BytesSer, DAG_CBOR};
+use fvm_sdk as sdk;
+use fvm_shared::error::ExitCode;
+
+#[no_mangle]
+pub fn invoke(_: u32) -> u32 {
+    std::panic::set_hook(Box::new(|info| {
```
Stebalien, line 7: This just makes debugging a _lot_ easier.

raulk:
Will review this tomorrow, but just noting that @mriise is working on #577 separately.

Stebalien:
Yeah, that'll give us quite a bit of coverage. But we still need integration tests like this to get syscall (mostly concerning illegal arguments, etc.) and sdk coverage.

sdk/src/ipld.rs
```diff
@@ -62,7 +62,13 @@ pub fn get_block(id: fvm_shared::sys::BlockId, size_hint: Option<u32>) -> Syscal
...
             debug_assert!(remaining <= 0, "should have read whole block");
         }
-        buf.set_len(buf.capacity() + (remaining as usize));
+        let size = (buf.capacity() as i64) + (remaining as i64);
+        debug_assert!(size >= 0, "size can't be negative");
+        debug_assert!(
+            size <= buf.capacity() as i64,
+            "size shouldn't exceed capacity"
+        );
+        buf.set_len(size as usize);
```
wadeAlexC, line 71: Could you change these to harder assertions?

Stebalien: Yeah, fair enough. Changed.

vyzo (approved):

PR #586, opened 2022-05-20
fix: make flush match lotus as much as possible

Stebalien:
1. Reject anything we don't expect.
2. Properly handle identity cids.

We're _also_ checking the hash function (only allowing blake2b). I could drop that, but it seems like a reasonable sanity check.

### DIFF
--- a/fvm/src/blockstore/buffered.rs
+++ b/fvm/src/blockstore/buffered.rs
@@ -5,6 +5,7 @@ use anyhow::{anyhow, Result};
 use byteorder::{BigEndian, ByteOrder, ReadBytesExt};
 use cid::Cid;
 use fvm_ipld_blockstore::{Blockstore, Buffered};
+use fvm_shared::commcid::{FIL_COMMITMENT_SEALED, FIL_COMMITMENT_UNSEALED};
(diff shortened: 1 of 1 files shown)


fvm/src/blockstore/buffered.rs
```diff
@@ -171,31 +172,59 @@ fn copy_rec<'a>(
...
-    scan_for_links(&mut Cursor::new(block), |link| {
-        if link.codec() != DAG_CBOR {
-            return Ok(());
+    const DAG_RAW: u64 = 0x55;
+    const BLAKE2B_256: u64 = 0xb220;
+    const BLAKE2B_LEN: u8 = 32;
+    const IDENTITY: u64 = 0x0;
+
+    // Differences from lotus (vm.Copy):
+    // 1. We assume that if we don't have a blcok in our buffer, it must already be in lotus. We
```
arajasek, line 181:
```suggestion
    // 1. We assume that if we don't have a block in our buffer, it must already be in the client.
```

Can we rewrite this to not be Lotus-specific? Lotus is not the only client integrating the FVM.

PR #588, opened 2022-05-23
AsRef/AsMut for `ipld_create/read`

mriise:
A little nicety I thought of while building kernel tests, not sure how it would relate to morphization bloat mentioned in #373

### DIFF
--- a/fvm/src/kernel/default.rs
+++ b/fvm/src/kernel/default.rs
@@ -275,7 +275,7 @@ where
         Ok((id, stat))
     }
 
-    fn block_create(&mut self, codec: u64, data: impl AsRef<[u8]>) -> Result<BlockId> {
+    fn block_create(&mut self, codec: u64, data: &(impl AsRef<[u8]> + ?Sized)) -> Result<BlockId> {
         self.call_manager
             .charge_gas(self.call_manager.price_list().on_block_create( core::mem::size_of_val(data.as_ref())))?;
 
@@ -314,7 +314,7 @@ where
         Ok(k)
     }
 
-    fn block_read(&mut self, id: BlockId, offset: u32, mut buf: impl AsMut<[u8]>) -> Result<i32> {
+    fn block_read(&mut self, id: BlockId, offset: u32, buf: &mut (impl AsMut<[u8]> + ?Sized)) -> Result<i32> {
         let block = self.blocks.get(id)?;
         let data = block.data();
 
--- a/fvm/src/kernel/default.rs
+++ b/fvm/src/kernel/default.rs
@@ -275,11 +275,11 @@ where
         Ok((id, stat))
     }
 
-    fn block_create(&mut self, codec: u64, data: &[u8]) -> Result<BlockId> {
+    fn block_create(&mut self, codec: u64, data: impl AsRef<[u8]>) -> Result<BlockId> {
         self.call_manager
-            .charge_gas(self.call_manager.price_list().on_block_create(data.len()))?;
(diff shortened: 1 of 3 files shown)


Stebalien:
So, there's a trade-off here:

- Complexity of implementation.
- Complexity of use.

In general:

- For public interfaces, common concrete types, collections, etc .where I'm going to implement them once and use them frequently, I go full generic so that the caller can just pass in whatever they want.
- For common interfaces, or interfaces that only have a few users (not including tests), I optimize to make them easy to understand/read/implement, even if I need a little extra boilerplate to _use_ them.

In this case, given that the extra boilerplate to use them is a simple `&`, I'd leave the code as-is.

The _other_ consideration here is object safety, but I don't think that's _that_ much of an issue right now.


> would relate to morphization bloat mentioned in https://github.com/filecoin-project/ref-fvm/issues/373

That issue is specifically concerned with monomorphizeation in actors. The kernel lives in the FVM, so it's not an issue.

fvm/src/kernel/default.rs
```diff
@@ -275,11 +275,11 @@ where
         Ok((id, stat))
     }
 
-    fn block_create(&mut self, codec: u64, data: &[u8]) -> Result<BlockId> {
+    fn block_create(&mut self, codec: u64, data: &(impl AsRef<[u8]> + ?Sized)) -> Result<BlockId> {
         self.call_manager
```
Stebalien, line 279: note: one would usually call `let data = data.as_ref()` at the top.

fvm/src/kernel/default.rs
```diff
@@ -275,11 +275,11 @@ where
...
-            .charge_gas(self.call_manager.price_list().on_block_create(data.len()))?;
+            .charge_gas(self.call_manager.price_list().on_block_create( core::mem::size_of_val(data.as_ref())))?;
```
Stebalien, line 280: `len` is still the correct thing to call here. _Technically_, `size_of_val` will work, but it has different semantics (byte-size of an arbitrary dynamically sized type, versus the length of a byte buffer).

> core

Given that we depend on `std`, just use `std`. Use `core` in `no_std` crates.

fvm/src/kernel/default.rs
```diff
@@ -314,19 +314,19 @@ where
         Ok(k)
     }
 
-    fn block_read(&mut self, id: BlockId, offset: u32, buf: &mut [u8]) -> Result<i32> {
+    fn block_read(&mut self, id: BlockId, offset: u32, buf: &mut (impl AsMut<[u8]> + ?Sized)) -> Result<i32> {
```
Stebalien, line 317: One would usually take `impl AsRef<[u8]>`:

- This makes it possible to pass a buffer by-value (if desired).
- `AsRef<U>` is implemented over `&T` where `T: AsRef<U> +?Sized`.
- This lets you drop the `?Sized`.

Stebalien (commented):
(comments just to teach a bit of rust)

fvm/src/kernel/default.rs
Stebalien: Why:

- Can leave the generics out of the rest of the function.
- Maybe the compiler will get smart and split the function into a monomorphized portion, and a shared portion.

fvm/src/kernel/default.rs
Stebalien: I.e.

```rust
     fn block_read(&mut self, id: BlockId, offset: u32, buf: impl AsMut<[u8]>) -> Result<i32> { 
```

mriise:
> The other consideration here is object safety

good to know :+1:

mriise:
closing since this won't be merged

PR #591, opened 2022-05-24
fix: we're not lotus specific

Stebalien:


fvm/src/blockstore/buffered.rs
```diff
@@ -175,10 +175,10 @@ fn copy_rec<'a>(
     const IDENTITY: u64 = 0x0;
 
     // Differences from lotus (vm.Copy):
-    // 1. We assume that if we don't have a block in our buffer, it must already be in the client.
-    //    don't check. This should only happen if the lotus node is missing state.
-    // 2. We always write-back new blocks, even if lotus already has them. We haven't noticed a perf
-    //    impact.
+    // 1. We assume that if we don't have a block in our buffer, it must already be in the client
+    //    and don't check. This should only happen if the lotus node is missing state.
```
arajasek, line 179:
```suggestion
    //    and don't check. This should only happen if the client node is missing state.
```

arajasek (changes requested):
Please fix.

arajasek (approved):

PR #596, opened 2022-05-30
fix: AMT: correctly return whether found in batch_delete

arajasek:
Thanks to @WadeAlexC for the catch!


dignifiedquire:
what about adding a test?

raulk (approved):
While the short-circuiting behaviour does not change, this fixes the return value, so it looks correct. However, a test would be nice (and simple to implement) as @dignifiedquire says.

Stebalien (approved):

PR #598, opened 2022-06-01
Unit Tests for Default kernel `GasOps`

mriise:
Also adds `pretty_assertions` as a dev dependency since trying to find what is wrong with large structs using the default is near impossible.


mriise:
The tests feel a bit sparse to me, but they are relatively simple operations so it could just be a feeling. If CI passes and there aren't any tests that I i missed it will be ready for review @Stebalien

Stebalien (approved):
LGTM as basic gas tracker tests. But we should probably have more tests of the actual `Gas` primitive (test saturation, rounding, etc.).

Issue #602, opened 2022-06-02
Fix instrumentation when the bulk memory is used


Stebalien:
See https://github.com/paritytech/parity-wasm/pull/331. The _rust_ compiler doesn't appear to output this section. But other compilers, and optimizers including binaryen, do.

Stebalien:
Fixed.

PR #603, opened 2022-06-02
fix: fixup wasm sections after instrumentation

Stebalien:
This works around #602, until patches get merged upstream.


Stebalien:
This appears to fix the issue.

magik6k (approved):

PR #605, opened 2022-06-02
debug execution: implement actor redirects in engine

vyzo:
This seems like the simplest way to do it, without getting into a mess trying to implement intercepting or generic engines.

### DIFF
--- a/fvm/src/machine/engine.rs
+++ b/fvm/src/machine/engine.rs
@@ -201,7 +201,7 @@ impl Engine {
             module_cache: Default::default(),
             instance_cache: Mutex::new(anymap::Map::new()),
             config: ec,
-            actor_redirect: actor_redirect,
+            actor_redirect,
         })))
     }
 }


fvm/src/machine/engine.rs
```diff
@@ -182,13 +186,19 @@ impl Engine {
         let dummy_memory = Memory::new(&mut dummy_store, MemoryType::new(0, Some(0)))
             .expect("failed to create dummy memory");
 
+        let actor_redirect = ec
+            .actor_redirect
+            .as_ref()
+            .map(|v| v.iter().cloned().collect());
```
Stebalien, line 192: If we're not overriding anything, we might as well just have an empty hashmap.

```suggestion
            .as_ref()
            .unwrap_or_default().iter().cloned().collect();
```

vyzo: sure, do you want to drop the option from the data structure?
I thought it would be faster to `None` check instead of a get in an empty map, but if there is no difference we might as well.

raulk: An option prevents us from hashing on access when there are no redirects set.

raulk (approved):

vyzo:
rebased on master for the logging fix.

fvm/src/machine/mod.rs
```diff
@@ -116,6 +116,9 @@ pub struct NetworkConfig {
     ///
     /// DEFAULT: The price-list for the current network version.
     pub price_list: &'static PriceList,
+
+    /// Actor redirects for debug execution
+    pub actor_redirect: Option<Vec<(Cid, Cid)>>,
```
Stebalien, line 121: No reason to make this an option either. An empty vector is free.

PR #606, opened 2022-06-03
fix: machine config: disable async even with feature enabled

magik6k:
Might fix the `wasm stack size cannot exceed the async stack size` issue @karim-agha is seeing


Stebalien (approved):
So, this may help, but it won't always help. Unfortunately, due to the way features work, if some _other_ crate enables the async feature, this cfg won't catch taht.

Issue #608, opened 2022-06-06
Generate Code Coverage Reports


mriise:
With Tests (https://github.com/filecoin-project/ref-fvm/issues/595, https://github.com/filecoin-project/ref-fvm/issues/585) being added to FVM, a good metric to start tracking is code coverage. 

Note: this is separate from https://github.com/filecoin-project/ref-fvm/issues/314 that is for code coverage inside actors.

mriise:
@dtynn pinging since this is (somewhat) related to the work you have already done

dtynn:
Hi @mriise 
I did some work in #314, but as you know, it focus on generating & collection coverage instruments inside **generated wasm bytecode**..

For this issue and its related work, I do think filecoin-project/builtin-actors#22 & filecoin-project/builtin-actors#120 may be more helpful.

mriise:
@dtynn yes, code coverage for most things can be generated, but for integraion tests we try to build WASM modules (mainly fil_ipld_actor is important here) that get then get ran buy tests which would be nice to count towards coverage. We _could_ just ignore the coverage that calls inside of the actors give and just run a custom setup for integration tests that prevents llvm-cov from trying to add instrumentation to the actors and just instruments the runtime.
Either way I would be happy to take up the work you have already done + add it to CI and get things merged.

dtynn:
ah, it would be nice to see some pieces of the work are useful to you.,
however, there are a few tricky steps before we can make everything work.
let me know if you need any help

Stebalien:
@mriise what's the current status of this? (i.e., what's done, what's left, etc.)

mriise:
Current coverage (at the time of writing) is ~50% https://app.codecov.io/gh/filecoin-project/ref-fvm
Done: https://github.com/filecoin-project/ref-fvm/pull/616, https://github.com/filecoin-project/ref-fvm/pull/612

easiest way to improve the % IMO would be to write future syscall tests in actors to excersize the sdk (currently at near 0%), syscalls, and everything else in the path. Most of the lines outside of syscall linking helping things that we should have covered are in DefaultKernel and the SDK.

Stebalien:
Ah, ok, so it sounds like we're including all the tests we care about in "code coverage".

PR #612, opened 2022-06-08
Rust Code Coverage Reporting

mriise:
This Does not include code coverage for WASM modules or pulling anything from reports done by conformance and integration tests.


.github/workflows/coverage.yml
```diff
@@ -0,0 +1,22 @@
...
+
+jobs:
+  coverage:
+    runs-on: ubuntu-latest
+    env:
+      CARGO_TERM_COLOR: always
+    steps:
+      - uses: actions/checkout@v3
+      - name: Install Rust
+        run: rustup toolchain install nightly-2022-05-09 --component llvm-tools-preview
```
Stebalien, line 13: https://github.com/filecoin-project/ref-fvm/blob/0b2a24fdff8de8dac6aa0d262be719942ad9a2d8/.github/workflows/ci.yml#L18-L23

Stebalien: (but with the right component.

.github/workflows/coverage.yml
```diff
@@ -0,0 +1,22 @@
...
+      - name: Install cargo-llvm-cov
+        uses: taiki-e/install-action@cargo-llvm-cov
+      - name: Generate code coverage
+        run: cargo llvm-cov --no-default-features --features debug --workspace --exclude "*tests*" --exclude "fil*" --lcov --output-path lcov.info
```
Stebalien, line 17: This will re-run all the tests, right? We proabably just want to modify the existing workflows to generate code coverage.

Stebalien (commented):
(we should also remove `lcov.info` from git)

.github/workflows/coverage.yml
mriise: ok

mriise:
coverage percentage should go down as it is no longer including anything outside of the FVM package. 

Current setup is brittle till I learn how to actually write CI.
Pulling in `llvm-cov` should be done with a SHA at some point.
In order to have more than just a single report made we will likely need to use `grcov` to aggregate reports, tho this should probably happen after I figure out CI so it is done properly.

Stebalien:
> In order to have more than just a single report made we will likely need to use grcov to aggregate reports, tho this should probably happen after I figure out CI so it is done properly.

The codecov script will automatically pick up all code coverage outputs and aggregate them.

Stebalien:
> The codecov script will automatically pick up all code coverage outputs and aggregate them.

IIRC, codecov will also merge uploads from multiple jobs.

mriise:
> > The codecov script will automatically pick up all code coverage outputs and aggregate them.
> 
> IIRC, codecov will also merge uploads from multiple jobs.

Ooo I'll double check but that's super convenient

mriise:
linking related issue #608

mriise:
I needed to change the name of the jobs back to just `test` since it requires a job named that to complete. Also setting integration and test-fvm as required jobs is a setting in the repo that I don't have access to.

.github/workflows/ci.yml
```diff
@@ -37,11 +37,18 @@ jobs:
...
+            components: llvm-tools-preview
           - name: test
+            covname: lcov.info
+            command: llvm-cov
+            args: --all --exclude fvm --exclude fvm_conformance_tests --exclude fvm_integration_tests --exclude "*actor" --lcov --output-path lcov.info
+            components: llvm-tools-preview
+          - name: integration
             command: test
-            args: --all --exclude fvm --exclude fvm_conformance_tests
+            args: --package fvm_integration_tests --package "*actor"
```
Stebalien, line 51: - Any reason we're excluding the actor packages?
- Any reason not to count the integration tests towards test coverage?

mriise: - actor tests try to build to wasm via `wasm-builder` and breaks the coverage tool so getting coverage reporting them are more complex (will likely need to edit some cargo.toml) than I wanted to have for this PR
- integration tests build actors

These should still be added, but I wanted to push this PR through first so we can start getting metrics. If you feel like it should all happen in one PR then I can do that instead.

Stebalien: Makes sense.

Stebalien (approved):

Stebalien:
Thanks! Now we just need to make that chart a bit more... green.

Issue #609, opened 2022-06-07
ci: set up platform/OS matrix of builds


raulk:
ARM64 architecture wasn't the problem in https://github.com/filecoin-project/ref-fvm/issues/599, the ancient Linux kernel was. However, we should run unit, integration and conformance tests (as well as the build, obviously) on CI on the platforms and OS we support. Currently this is limited to the following by Wasmtime:

* Linux x86_64 (already tested)
* Linux aarch64 (no coverage)
* macOS x86_64 (no coverage)
* Windows x86_64 (no coverage)

While Windows is not supported by Lotus: https://lotus.filecoin.io/lotus/install/prerequisites/ (and it's safe to assume that other Filecoin clients like Forest, Venus and Fuhon don't aim to support it either), developers building on the FVM _will_ run Windows. Therefore, we can't discard this OS.

raulk:
@galargh has agreed to help out in setting up Linux on aarch64 via self-hosted runners. From a conversation with him:

> I’m able to bring up graviton machines as gh self-hosted runners now.
> Here’s where this setup lives: https://github.com/pl-strflt/tf-aws-gh-runner
> These are the instructions on how to start using runners defined in that repo in other repositories: https://github.com/pl-strflt/tf-aws-gh-runner#how-to-use-an-existing-self-hosted-runner-type-in-your-repository (please see the note about security of using self-hosted runners too).
> The runner type you’re after is linux-arm64-default. Here’s how it can be used in a workflow: https://github.com/pl-strflt/tf-aws-gh-runner/blob/7f94e9c19d62104de01a2f0df915d72d9d71164a/.github/workflows/playground.yml#L9
> And here’s an example workflow run that confirms the setup is OK: https://github.com/pl-strflt/tf-aws-gh-runner/runs/6753443481?check_suite_focus=true

On setting it up on ref-fvm:

> See the installation instructions here https://github.com/pl-strflt/tf-aws-gh-runner#how-to-use-an-existing-self-hosted-runner-type-in-your-repository
> TLDR, it requires
> installing a GitHub App for the org
> changing org settings to allow self-hosted runners usage in public repos
> changing repo settings to require approval for ALL workflow runs from outside contributors or changing org settings to allow self-hosted runners usage only in specific workflows (the latter cannot be applied to pull_request workflows; doing both is a valid choice too)

raulk:
(Oops, fat fingers made me accidentally close this)

raulk:
[...]

PR #616, opened 2022-06-11
Debug artifacts syscall + integration test coverage

mriise:
adds a new syscall under `debug` to save artifacts to a folder ~~in current dir `artifacts/` which is somewhat horrible and should probably be something else, but it works and adding the ability to modify that sort of thing is easy enough, just need to get a consensus on how it should be done.~~ defined by env var `FVM_STORE_ARTIFACT_DIR`

ATM this doesn't do any CI as I think CI should be heavily refactored anyways and adding this coverage can be added during then (ref https://github.com/filecoin-project/ref-fvm/issues/609)

### DIFF
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -45,7 +45,7 @@ pub struct InnerDefaultCallManager<M> {
     backtrace: Backtrace,
     /// The current execution trace.
     exec_trace: ExecutionTrace,
-    /// Current invocation count
+    /// Number of actors that have been invoked in this message execution.
     invocation_count: u64,
 }
 
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -45,6 +45,8 @@ pub struct InnerDefaultCallManager<M> {
     backtrace: Backtrace,
     /// The current execution trace.
     exec_trace: ExecutionTrace,
+    /// Current invocation count
+    invocation_count: u64,
 }
 
 #[doc(hidden)]
@@ -79,6 +81,7 @@ where
             call_stack_depth: 0,
             backtrace: Backtrace::default(),
             exec_trace: vec![],
+            invocation_count: 0,
         })))
     }
 
@@ -217,6 +220,10 @@ where
         self.num_actors_created += 1;
         ret
     }
+
+    fn invocation_count(&self) -> u64 {
+        self.invocation_count
+    }
 }
 
 impl<M> DefaultCallManager<M>
@@ -336,6 +343,9 @@ where
         } else {
             NO_DATA_BLOCK_ID
         };
+
+        // Increment invocation count
+        self.invocation_count += 1;
 
         // This is a cheap operation as it doesn't actually clone the struct,
         // it returns a referenced copy.
--- a/fvm/src/call_manager/mod.rs
+++ b/fvm/src/call_manager/mod.rs
@@ -84,7 +84,7 @@ pub trait CallManager: 'static {
     fn next_actor_idx(&mut self) -> u64;
 
     // TODO: should this be ActorID instead?
-    /// Gets the call-stack actor creation index (aka current total created actors)
+    /// Gets the total invocations done on this call stack.
     fn invocation_count(&self) -> u64;
 
     /// Returns the current price list.
--- a/fvm/src/call_manager/mod.rs
+++ b/fvm/src/call_manager/mod.rs
@@ -83,6 +83,10 @@ pub trait CallManager: 'static {
     /// Gets and increment the call-stack actor creation index.
     fn next_actor_idx(&mut self) -> u64;
 
+    // TODO: should this be ActorID instead?
+    /// Gets the call-stack actor creation index (aka current total created actors)
+    fn invocation_count(&self) -> u64;
+
     /// Returns the current price list.
     fn price_list(&self) -> &PriceList {
         self.machine().context().price_list
--- a/fvm/src/call_manager/mod.rs
+++ b/fvm/src/call_manager/mod.rs
@@ -83,7 +83,6 @@ pub trait CallManager: 'static {
     /// Gets and increment the call-stack actor creation index.
     fn next_actor_idx(&mut self) -> u64;
 
-    // TODO: should this be ActorID instead?
     /// Gets the total invocations done on this call stack.
     fn invocation_count(&self) -> u64;
 
--- a/fvm/src/kernel/default.rs
+++ b/fvm/src/kernel/default.rs
@@ -793,7 +793,7 @@ where
         self.call_manager.context().actor_debugging
     }
 
-    fn store_artifact(&self, name: &str, data: &[u8]) -> Result<()>{
+    fn store_artifact(&self, name: &str, data: &[u8]) -> Result<()> {
         // Ensure well formed artifact name
         {
             if name.len() > 256 {
@@ -818,12 +818,14 @@ where
         if let Ok(dir) = std::env::var(ENV_ARTIFACT_DIR) {
             let dir: PathBuf = [
                 dir,
-                self.call_manager.machine().machine_id().to_string(),
+                self.call_manager.machine().machine_id(),
                 self.call_manager.origin().to_string(),
                 self.call_manager.nonce().to_string(),
                 self.actor_id.to_string(),
                 self.call_manager.invocation_count().to_string(),
-            ].iter().collect();
+            ]
+            .iter()
+            .collect();
 
             if let Err(e) = std::fs::create_dir_all(dir.clone()) {
                 log::error!("failed to make directory to store debug artifacts {}", e);
--- a/fvm/src/kernel/default.rs
+++ b/fvm/src/kernel/default.rs
@@ -39,6 +39,7 @@ lazy_static! {
 
 const BLAKE2B_256: u64 = 0xb220;
 const ENV_ARTIFACT_DIR: &str = "FVM_STORE_ARTIFACT_DIR";
+const MAX_ARTIFACT_NAME_LEN: usize = 256;
 
 /// The "default" [`Kernel`] implementation.
 pub struct DefaultKernel<C> {
@@ -796,7 +797,7 @@ where
     fn store_artifact(&self, name: &str, data: &[u8]) -> Result<()> {
         // Ensure well formed artifact name
         {
-            if name.len() > 256 {
+            if name.len() > MAX_ARTIFACT_NAME_LEN {
                 Err("debug artifact name should not exceed 256 bytes")
             } else if name.chars().any(std::path::is_separator) {
                 Err("debug artifact name should not include any path separators")
@@ -815,14 +816,14 @@ where
         .or_error(fvm_shared::error::ErrorNumber::IllegalArgument)?;
 
         // Write to disk
-        if let Ok(dir) = std::env::var(ENV_ARTIFACT_DIR) {
+        if let Ok(dir) = std::env::var(ENV_ARTIFACT_DIR).as_deref() {
             let dir: PathBuf = [
                 dir,
                 self.call_manager.machine().machine_id(),
-                self.call_manager.origin().to_string(),
-                self.call_manager.nonce().to_string(),
-                self.actor_id.to_string(),
-                self.call_manager.invocation_count().to_string(),
+                &self.call_manager.origin().to_string(),
+                &self.call_manager.nonce().to_string(),
+                &self.actor_id.to_string(),
+                &self.call_manager.invocation_count().to_string(),
             ]
             .iter()
             .collect();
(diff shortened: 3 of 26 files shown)


mriise:
@dtynn It wont let me add you as a reviewer, though I would love to hear your thoughts on this

sdk/src/sys/debug.rs
```diff
@@ -9,4 +9,8 @@ super::fvm_syscalls! {
 
     /// Logs a message on the node.
     pub fn log(message: *const u8, message_len: u32) -> Result<()>;
+
+    /// TODO: this technically lets anyone store whatever data they want on the node's filesystem, this should *not* be enabled on an actual network until proper gaurds are in place
+    /// Save data as a debug artifact on the node.
```
dtynn, line 14: maybe we could put all these things into a special `feature`, and make sure that we can only touch them in tests & benches

fvm/src/syscalls/debug.rs
```diff
@@ -21,3 +23,36 @@ pub fn enabled(context: Context<'_, impl Kernel>) -> Result<i32> {
...
+    }
+    let src_dir = std::env::current_dir()
+        .or_fatal()?
+        .canonicalize()
+        .or_fatal()?
+        .join("artifacts");
+    DirBuilder::new()
+        .recursive(true)
+        .create(src_dir.clone())
+        .or_fatal()?;
```
Stebalien, line 47: This should probably be configured via an environment variable. That way we don't save artifacts when debug is enabled, unless the user explicitly configures where to put them.

fvm/src/syscalls/debug.rs
```diff
@@ -21,3 +23,36 @@ pub fn enabled(context: Context<'_, impl Kernel>) -> Result<i32> {
...
+
+    let data = context.memory.try_slice(data_off, data_len)?;
+    let name = context.memory.try_slice(name_off, name_len)?;
+    let name = String::from_utf8(name.to_owned())
+        .or_error(fvm_shared::error::ErrorNumber::IllegalArgument)?;
```
Stebalien, line 52: 1. No need for `to_owned`.
2. We should enforce some _basic_ naming rules:
    1. Less than 256 bytes.
    2. Single path component (make sure `path::is_separator` is false for all bytes).
    3. Doesn't start with a `.`.

fvm/src/syscalls/mod.rs
```diff
@@ -169,6 +169,7 @@ pub fn bind_syscalls(
 
     linker.bind("debug", "log", debug::log)?;
     linker.bind("debug", "enabled", debug::enabled)?;
+    linker.bind("debug", "capture_artifact", debug::capture_artifact)?;
```
Stebalien, line 172: nit: `save_artifact` or `write_artifact` or `store_artifact` etc. The artifact has already been created, so we're not really "capturing" anything.

sdk/src/debug.rs
```diff
@@ -38,6 +38,21 @@ mod inner {
         }
     }
 
+    /// Saves an artifact to the host env. New artifacts with the same name will overwrite old ones
+    pub fn capture_artifact(name: impl AsRef<str>, data: Vec<u8>) {
```
Stebalien, line 42:
```suggestion
    pub fn capture_artifact(name: impl AsRef<str>, data: impl AsRef<[u8]>) {
```

Or just `&[u8]` is probably sufficient. But we definitely don't need ownership.

sdk/src/sys/debug.rs
Stebalien: That would make it difficult to debug from lotus. IMO, an environment variable (artifacts are only saved when set) and some path checks _should_ be sufficient?

Stebalien: In the future, we could expand that env var to a set of rules like `FVM_STORE_ARTIFACTS=code:SomeCid=/path/to/dir;addr:f01=/path/to/other-dir`.

fvm/src/syscalls/debug.rs
```diff
@@ -21,3 +23,36 @@ pub fn enabled(context: Context<'_, impl Kernel>) -> Result<i32> {
...
+
+    println!("writing artifact: {} to {:?}", name, src_dir);
+    std::fs::write(src_dir.join(name), data).or_fatal()?;
```
Stebalien, line 55: We should probably write this to a subdirectory. I'm thinking `$epoch-$randomness/$msg_from_addr/$msg_nonce/$actor_addr/$invocation_count`. Where:

- `$epoch` is the current epoch.
- `$randomness` is derived from tipset randomness (the `get_randomness_from_tickets` syscall but without charging for gas). We should probably invoke this _once_ when constructing the machine to generate the top-level directory. Otherwise, chain reorgs (which replay epochs with different messages) would cause problems.
- `$msg_from_addr` is the message's "origin" address (stored as `origin` in the call manager.
- `$msg_nonce` is the nonce of that message.
- `$actor_addr` is the ID address of the actor saving the artifact.
- `$invocation_count` is an incrementing invocation ID stored in the call manager. Every time we invoke a new actor in the call manager, we increment this ID and store that ID in the kernel. That way, each invocation of an actor in a call stack gets a different directory.

This ensures actor invocations all get their own scoped directories. This isn't really about _security_, it's mostly just to reduce the amount of actor-side logic one might need to avoid clobbering artifacts.

fvm/src/syscalls/debug.rs
mriise: iv. is a least one char

fvm/src/syscalls/debug.rs
mriise: Should that be a fatal error or fail silently?

fvm/src/call_manager/mod.rs
```diff
@@ -83,6 +83,10 @@ pub trait CallManager: 'static {
     /// Gets and increment the call-stack actor creation index.
     fn next_actor_idx(&mut self) -> u64;
 
+    // TODO: should this be ActorID instead?
+    /// Gets the total invocations done on this call stack.
```
Stebalien, line 87: Nah, ActorID is the ID address of an actor. This is just a counter.

fvm/src/machine/default.rs
```diff
@@ -37,6 +37,9 @@ pub struct DefaultMachine<B, E> {
     state_tree: StateTree<BufferedBlockstore<B>>,
     /// Mapping of CIDs to builtin actor types.
     builtin_actors: Manifest,
+    /// Somewhat unique ID of the machine consisting of (epoch, randomness)
+    /// randomness is generated with `initial_state_root`
+    id: MachineId,
```
Stebalien, line 42: I'd just use a string. The ID is opaque.

fvm/src/syscalls/debug.rs
```diff
@@ -21,3 +21,45 @@ pub fn enabled(context: Context<'_, impl Kernel>) -> Result<i32> {
...
+            .ok_or("debug artifact name should be at least one character")
+            .or_error(fvm_shared::error::ErrorNumber::IllegalArgument)?
+            == '.'
+        {
+            Err("debug artifact name should not start with a decimal '.'")
+        } else {
+            Ok(())
+        }
+    }
+    .or_error(fvm_shared::error::ErrorNumber::IllegalArgument)?;
```
Stebalien, line 60: Let's move this into the kernel itself. Ideally, the syscalls would do the minimum possible parsing.

testing/integration/README.md
```diff
@@ -19,3 +19,11 @@ The following flow has been defined as a default usage:
 1. Wasm bytecode is now expected to be received through a binary type (`&[u8]`). This be upgraded to work Rust module compiled
 at test time.
 2. Some testing and examples should be added to demonstrate how the framework works.
+
+TODO: (hack to get coverage reports from actors + integration tests) 
+```shell
+cargo build -p "*actor"
+export SKIP_WASM_BUILD true
+export FVM_STORE_ARTIFACT_DIR ../../target/llvm-cov-target/
+cargo llvm-cov -p fvm_integration_tests --lcov
```
Stebalien, line 28: Let's stick with bash.

fvm/src/call_manager/default.rs
```diff
@@ -45,6 +45,8 @@ pub struct InnerDefaultCallManager<M> {
     backtrace: Backtrace,
     /// The current execution trace.
     exec_trace: ExecutionTrace,
+    /// Current invocation count
```
Stebalien, line 48: The number of actors that have been invoked in this message execution.

fvm/src/kernel/default.rs
```diff
@@ -790,6 +792,30 @@ where
...
+
+    fn store_artifact(&self, name: &str, data: &[u8]) {
+        if let Ok(dir) = std::env::var(ENV_ARTIFACT_DIR) {
+            let mut dir = PathBuf::from(dir);
+
+            dir.push(self.call_manager.machine().machine_id().to_string());
+            dir.push(self.call_manager.origin().to_string());
+            dir.push(self.call_manager.nonce().to_string());
+            dir.push(self.actor_id.to_string());
+            dir.push(self.call_manager.invocation_count().to_string());
```
Stebalien, line 804: nit: I think you can write something like.

```suggestion
            let dir: PathBuf = [
                self.call_manager.machine().machine_id().to_string(),
                self.call_manager.origin().to_string(),
                self.call_manager.nonce().to_string(),
                self.actor_id.to_string(),
                self.call_manager.invocation_count().to_string(),
            ].collect()
```

Stebalien (commented):
LGTM with some nits.

testing/integration/README.md
Stebalien: I meant `export FOO bar` isn't valid bash. `export FOO=bar` is.

fvm/src/machine/default.rs
mriise: so `id: String` and `Self { id: machine_id.to_string(), .. }` ?

testing/integration/README.md
mriise: ah, well who uses bash anyways, fish is better :^)

fvm/src/kernel/default.rs
```diff
@@ -790,6 +792,55 @@ where
     fn debug_enabled(&self) -> bool {
         self.call_manager.context().actor_debugging
     }
+
+    fn store_artifact(&self, name: &str, data: &[u8]) -> Result<()> {
+        // Ensure well formed artifact name
+        {
+            if name.len() > 256 {
```
Stebalien, line 799: nit: make this a constant.

fvm/src/machine/default.rs
```diff
@@ -117,12 +120,24 @@ where
...
             engine: engine.clone(),
             externs,
             state_tree,
             builtin_actors,
+            id: format_args!(
+                "{}-{}",
+                context.epoch,
+                cid::multibase::encode(cid::multibase::Base::Base58Btc, &randomness[..16])
+            )
+            .to_string(),
```
Stebalien, line 140:
```suggestion
            id: format!(
                "{}-{}",
                context.epoch,
                cid::multibase::encode(cid::multibase::Base::Base58Btc, &randomness[..16])
            ),
```

fvm/src/machine/default.rs
```diff
@@ -117,12 +120,24 @@ where
...
         Ok(DefaultMachine {
             context: context.clone(),
             engine: engine.clone(),
             externs,
             state_tree,
             builtin_actors,
+            id: format_args!(
+                "{}-{}",
+                context.epoch,
+                cid::multibase::encode(cid::multibase::Base::Base58Btc, &randomness[..16])
```
Stebalien, line 138: Base58 doesn't work on windows (case sensitive). We don't _really_ support it, but I'd just use base32 unless we have a strong reason to use base58.

fvm/src/machine/default.rs
```diff
@@ -234,4 +249,8 @@ where
     fn into_store(self) -> Self::Blockstore {
         self.state_tree.into_store()
     }
+
+    fn machine_id(&self) -> String {
+        self.id.clone()
+    }
```
Stebalien, line 255:
```suggestion
    fn machine_id(&self) -> &str {
        &self.id
    }
```

testing/integration/src/tester.rs
```diff
@@ -178,6 +178,8 @@ where
 
         let mut nc = NetworkConfig::new(self.nv);
         nc.override_actors(self.builtin_actors);
+        // TODO: maybe cfg flag this?
+        nc.enable_actor_debugging();
```
Stebalien, line 182: Eh, it's reasonable to just always enable this in tests.

Stebalien (approved):

fvm/src/machine/default.rs
```diff
@@ -117,12 +120,23 @@ where
         #[cfg(not(any(test, feature = "testing")))]
         engine.preload(state_tree.store(), builtin_actors.left_values())?;
 
+        let randomness = externs.get_chain_randomness(
+            0,
+            context.epoch,
```
arajasek, line 125: It is unsupported behaviour to call `get_chain_randomness` with an epoch greater than or equal to the current epoch (the `context.epoch` in this case).

If it's not too much work, I would suggest looking for a simpler source of randomness here.

PR #627, opened 2022-06-19
Recover pubkey syscall

karim-agha:
This change
- Implements `recover_public_key` syscall in FVM for getting sender's public key from signature
- Fixes a bug in the call manager that fails to load uncached actor code.
- Updates the builtin actors bundle to V8


fvm/src/machine/engine.rs
```diff
@@ -200,6 +202,32 @@ struct Cache<K> {
...
+    /// Implicitly caches instantiated modules.
+    pub fn load_code_cid<BS: Blockstore>(
+        &self,
+        code_cid: &Cid,
+        blockstore: BS,
+    ) -> anyhow::Result<Option<wasmtime::Module>> {
+        {
+            let cache = self.0.module_cache.lock().expect("module_cache poisoned");
+
+            if cache.contains_key(code_cid) {
```
karim-agha, line 218: `if let Some(...) = cache.get(..)` feels more elegant. todo

fvm/src/machine/engine.rs
```diff
@@ -118,6 +118,8 @@ pub fn default_wasmtime_config() -> wasmtime::Config {
     // wasmtime default: 512KiB
     // Set to something much higher than the instrumented limiter.
     // Note: This is in bytes, while the instrumented limit is in stack elements
+    #[cfg(feature = "testing")]
+    c.async_stack_size(4 << 20).unwrap();
```
raulk, line 122: What's going on here?

fvm/src/machine/engine.rs
```diff
@@ -200,6 +202,31 @@ struct Cache<K> {
 }
 
 impl Engine {
+    /// Attempts to load Actor Code CID from the blockstore and
+    /// instantiate a wasmtime Module from the WASM bytecode if the
+    /// given code CID is present, otherwise returns None.
+    ///
+    /// Implicitly caches instantiated modules.
+    pub fn load_code_cid<BS: Blockstore>(
```
raulk, line 210: Any reason why not to enhance get_module instead?

shared/src/crypto/signature.rs
```diff
@@ -21,6 +21,8 @@ pub const BLS_PUB_LEN: usize = 48;
 pub const SECP_SIG_LEN: usize = 65;
 /// Secp256k1 Public key length in bytes.
 pub const SECP_PUB_LEN: usize = 65;
+/// Length of the signature input message hash in bytes (32).
+pub const MESSAGE_SIZE: usize = 32;
```
raulk, line 25: This name is ambiguous. Can we consider more specific names?

testing/integration/Cargo.toml
```diff
@@ -8,13 +8,13 @@ repository = "https://github.com/filecoin-project/ref-fvm"
...
-fvm_ipld_hamt = { version = "0.5.1", path = "../../ipld/hamt"}
-fvm_ipld_amt = { version = "0.4.2", path = "../../ipld/amt"}
+fvm_ipld_hamt = { version = "0.5.1", path = "../../ipld/hamt" }
+fvm_ipld_amt = { version = "0.4.2", path = "../../ipld/amt" }
 fvm_ipld_car = { version = "0.4.1", path = "../../ipld/car" }
 fvm_ipld_blockstore = { version = "0.1.1", path = "../../ipld/blockstore" }
 fvm_ipld_encoding = { version = "0.2.2", path = "../../ipld/encoding" }
 
-actors-v7 = { version = "~7.5", package = "fil_builtin_actors_bundle" }
+actors-v8 = { package = "fil_builtin_actors_bundle", git = "https://github.com/filecoin-project/builtin-actors.git", branch = "experimental/fvm-m2" }
```
raulk, line 17: This is tricky. In theory v8 should point to a concrete version now, and we should start a v9 actors on master where we merge our changes to the init actor. However, in reality they will likely be rolled out in v10.

raulk (commented):
A few comments. The only blocking one is figuring out the actors versioning. Other than that, LGTM.

fvm/src/machine/engine.rs
karim-agha: When the integration test code is running, the `async` feature is on in wasmtime and it checks if the `async_stack_size` is at least as big as the stack size that is set in the following line, if it is smaller it panics.

fvm/src/machine/engine.rs
karim-agha: yeah, that was my first attempt, but here we need access to the `blockstore`, which is accessible from the kernel. The kernel gets moved and consumed by value when crating a wasmer store. Then the store is needed to construct an `Instance`. This dependency cycle has few solutions, the least elegant and invasive is this thing that I have here.

Once I have a complete WASM prototype, I thought about tacking the lifetimes usage between `Kernel`, `CallManager` and `Machine`. I have few ideas around it but those changes would be quite big and touch many files.

testing/integration/Cargo.toml
karim-agha: https://crates.io/crates/fil_builtin_actors_bundle is a `7.5.1` and master is at `8.0.0-alpha.1`. Should I branch off `experimental/fvm-m2` and bump versions to 9?

fvm/src/kernel/default.rs
```diff
@@ -460,6 +460,21 @@ where
         })
     }
 
+    fn recover_public_key(
```
Stebalien, line 463:
```suggestion
    fn recover_secp_public_key(
```

fvm/src/kernel/default.rs
```diff
@@ -460,6 +460,21 @@ where
...
+        &mut self,
+        hash: &[u8; SIG_MESSAGE_HASH_SIZE],
```
Stebalien, line 465: This is more "flexible", but a bit less convenient. Are there any use-cases where we wouldn't want to take the _message_ itself, and hash it internally?

fvm/src/machine/engine.rs
Stebalien: This issue has been fixed on master, you shouldn't need to change this.

fvm/src/syscalls/crypto.rs
```diff
@@ -43,6 +43,28 @@ pub fn verify_signature(
         .map(|v| if v { 0 } else { -1 })
 }
 
+pub fn recover_public_key(
+    context: Context<'_, impl Kernel>,
+    hash_off: u32,
+    hash_len: u32,
+    sig_off: u32,
+    sig_len: u32,
+) -> Result<[u8; SECP_PUB_LEN]> {
```
Stebalien, line 52: If these are actually fixed-length, we can just pass the offset and not the length.

sdk/src/actor.rs
```diff
@@ -62,6 +62,12 @@ pub fn create_actor(actor_id: ActorID, code_cid: &Cid) -> SyscallResult<()> {
     unsafe { sys::actor::create_actor(actor_id, cid.as_ptr()) }
 }
 
+/// Installs or ensures an actor code CID is valid and loaded.
+pub fn install_actor(code_cid: &Cid) -> SyscallResult<()> {
+    let cid = code_cid.to_bytes();
+    unsafe { sys::actor::install_actor(cid.as_ptr()) }
+}
+
```
raulk, line 70: GitHub does not render the diff cleanly after switching the base branch to `master`. The `install_actor` syscall is already present in master. The branch history with multiple merges from ancilliary branches is confusing GH. I will have to close this PR, create a new branch with a clean diff, and open a new PR. I will preserve authorship in the commit history.

PR #632, opened 2022-06-23
Release FVM 1.1

arajasek:
Release FVM 1.1, with dual execution support.

### DIFF
--- a/fvm/CHANGELOG.md
+++ b/fvm/CHANGELOG.md
@@ -6,6 +6,10 @@ Changes to the reference FVM implementation.
 
 ...
 
+## 1.1.0 [2022-06-27]
+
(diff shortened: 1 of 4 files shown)


Stebalien (approved):

arajasek:
@Stebalien Can I get a second approval on this? I mostly wanna make sure you are okay releasing a _minor_ version of the FVM for something that is NOT a breaking change. The only reason we are doing that is so that any hotfixes on 1.0.0 don't have to go on top of this (not strictly necessary) change.

Is that an okay thing to do?

Stebalien:
@arajasek that's fine. In general, miner releases _are_ for non-breaking changes.

Stebalien (approved):

PR #635, opened 2022-06-23
Release V1

Stebalien:


arajasek:
@Stebalien Should the version be something like 2.0.0-pre or some such?

Besides that, this definitely works for what we currently need, since all the changes are wanted in master. There is a hypothetical future where we don't want to be doing this, because changes in version `n` are actually not wanted in `master`, which has moved onto `n+1`. We can custom pick the commits in such a case, I just wanna flag that this may not always be the workflow we want.

Stebalien:
> Should the version be something like 2.0.0-pre or some such?

Not yet, we still need to merge the release branch into master. I don't want to bump master to v2 before it even hits v1...

Stebalien:
Well, I guess that wouldn't be the end of the world, but still.

PR #652, opened 2022-06-28
Trim Ci cache before upload, rollback wasmtime to v0.37

mriise:
Wasmtime v0.38 introduced a regression, causing integration test `native_overflow_somethingsomething` to run for waay to long.

Rolling back the dep brought back some CI issues I saw earlier where the cache would keep growing after each build, so now it is trimmed before upload (as well as printing out sccache and cargo cache size)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -121,9 +121,11 @@ jobs:
       if: ${{ matrix.push }}
       uses: actions-rs/cargo@844f36862e911db73fe0815f00a4a2602c279505 # v1.0.3
       with:
-        command: install cargo-cache
+        command: install
+        args: cargo-cache
     - name: Cleaning Cargo cache
       if: ${{ matrix.push }}
       uses: actions-rs/cargo@844f36862e911db73fe0815f00a4a2602c279505 # v1.0.3
       with:
-        command: cargo cache -e -i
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -117,3 +117,21 @@ jobs:
...
+      uses: actions-rs/cargo@844f36862e911db73fe0815f00a4a2602c279505 # v1.0.3
+      with:
+        command: cache
+        args: -e -i
+    - name: Getting sccache size
+      if: ${{ matrix.push }}
+      uses: actions-rs/cargo@844f36862e911db73fe0815f00a4a2602c279505 # v1.0.3
+      with:
+        command: cache
+        args: sccache
```
Stebalien, line 137: Do we actually need sscache?

mriise: not really? but it's not what is causing the cache to grow, its `/.cargo/registry`

mriise: it does improve compile times by a bit and stays around 500MB, so I wouldn't remove it just because

mriise:
success! https://github.com/filecoin-project/ref-fvm/runs/7101326285?check_suite_focus=true

Stebalien (approved):

PR #654, opened 2022-06-29
Actor upgrade workflow: the become syscall

vyzo:
Heavy WIP -- doesn't compile yet due to XXXs.

### DIFF
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -153,20 +153,18 @@ where
                     ExecutionEvent::CallError(SyscallError::new(ErrorNumber::Forbidden, "fatal"))
                 }
                 Err(ExecutionError::Syscall(s)) => ExecutionEvent::CallError(s.clone()),
-                Err(ExecutionError::Abort(Abort::Return)) =>
-                    ExecutionEvent::CallReturn(RawBytes::default()),
-                Err(ExecutionError::Abort(_)) => ExecutionEvent::CallError(SyscallError::new(
-                    ErrorNumber::Forbidden,
-                    "aborted",
-                )),
+                Err(ExecutionError::Abort(Abort::Return)) => {
+                    ExecutionEvent::CallReturn(RawBytes::default())
+                }
+                Err(ExecutionError::Abort(_)) => {
+                    ExecutionEvent::CallError(SyscallError::new(ErrorNumber::Forbidden, "aborted"))
+                }
             });
         }
 
         match result {
-            Err(ExecutionError::Abort(Abort::Return)) => {
-                Ok(InvocationResult::Return(None))
-            }
-            e => e
+            Err(ExecutionError::Abort(Abort::Return)) => Ok(InvocationResult::Return(None)),
+            e => e,
         }
     }
 
@@ -187,12 +185,10 @@ where
             .get_actor_id(who)?
             .ok_or_else(|| syscall_error!(NotFound; "actor not found: {}", who))?;
 
-
-        state_tree
-            .set_actor(
-                &origin,
-                ActorState::new(new_code_cid, state.state, state.balance, state.sequence),
-            )?;
+        state_tree.set_actor(
+            &origin,
+            ActorState::new(new_code_cid, state.state, state.balance, state.sequence),
+        )?;
 
         // abortive return
         // at this point, we are inside an invoke panic handler, from the message where the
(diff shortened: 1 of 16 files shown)


mriise:
so the syscall itself (from wasm perspective) would look like `fn become_actor(from: (CID)) -> !` where `!` is important inside WASM to trigger the trap code in wasmtime

from the FVM side of things, it would look more like `fn become_actor(context: Context<'_, impl Kernel>, from: (CID)) -> Result<Never, Abort>`

mriise:
@vyzo  why is the tuple important here?

vyzo:
extensibility and API consistency with the other entry points.

fvm/src/syscalls/actor.rs
```diff
@@ -109,3 +112,19 @@ pub fn get_code_cid_for_type(
...
+    // TODO different exit code?
+    let cid = context
+        .memory
+        .read_cid(code_cid_off)
+        .map_err(|e| Abort::from_error(ExitCode::USR_ILLEGAL_ARGUMENT, e))?;
+    context.kernel.become_actor(&cid)?; // ExitCode::SYS_INVALID_RECEIVER
+    Err(Abort::Exit(
+        ExitCode::OK,
+        "became a new actor something something".to_string(),
+    ))
```
Stebalien, line 129: Hm. So, the tricky part will be returning a value.

Stebalien: We probably need a new variant in `Abort::` along the lines of `Abort::Return`. (and maybe we should rename `Abort` to `EarlyExit`, or something like that...

sdk/src/sys/actor.rs
```diff
@@ -95,4 +95,8 @@ super::fvm_syscalls! {
     /// **Privledged:** May only be called by the init actor.
     #[doc(hidden)]
     pub fn create_actor(actor_id: u64, typ_off: *const u8) -> Result<()>;
+
+    /// TODO
+    #[doc(hidden)]
+    pub fn become_actor(cid_off: *const u8) -> !;
```
Stebalien, line 101: This is `Result<!>` because it _can_ fail.

fvm/src/kernel/mod.rs
```diff
@@ -184,6 +185,10 @@ pub trait ActorOps {
 
     /// Returns the CodeCID for the supplied built-in actor type.
     fn get_code_cid_for_type(&self, typ: actor::builtin::Type) -> Result<Cid>;
+
+    /// Become changes the code of the running actor using the specified code cid;
+    /// it does not return unless there is an error
+    fn become_actor(&self, code_cid: &Cid) -> std::result::Result<(), Abort>;
```
Stebalien, line 191: This should probably return `Result<BlockId>`.

- If it succeeds, we abort with the `BlockId` as the result.
- If it fails, we return a syscall error to the user.

sdk/src/sys/actor.rs
mriise: Result<!> requires nightly feature flag

vyzo: I have already changed this, pls stop.

mriise: Result<Infallible> should work the same though

Stebalien: Yep, either way.

fvm/src/syscalls/actor.rs
mriise: I like `Terminate`

fvm/src/call_manager/default.rs
```diff
@@ -148,10 +150,148 @@ where
...
+                // TODO charge gas
+                // Charge for any remaining uncharged execution gas, returning an error if we run
+                // out.
+                // charge_for_exec(&mut store)?;
+
+                // If the invocation failed due to running out of exec_units, we have already
+                // detected it and returned OutOfGas above. Any other invocation failure is returned
+                // here as an Abort
+                Ok(())
+            })();
```
mriise, line 284: can use `let res: Result<(), Abort> = { <code> Ok(()) }` instead of closure

mriise: nvm, we are using `?` opeator

fvm/src/call_manager/default.rs
```diff
@@ -430,6 +570,13 @@ where
                             "fatal error".to_owned(),
                             Err(ExecutionError::Fatal(err)),
                         ),
+                        // XXX this really is not possible dear compiler, Abort::Return
```
mriise, line 573: if something really isn't reachable, you can tell the compiler with `unreachable!()`
```rust 
match result {
    Ok(_) => unreachable!(),
    Err(e) => e,
}

PR #659, opened 2022-07-08
fix three bugs in our car implementation

Stebalien:
These aren't really critical as we only use CARs for distributing bundles, but they're still good fixes to have.

### DIFF
--- a/ipld/car/src/util.rs
+++ b/ipld/car/src/util.rs
@@ -11,6 +11,7 @@ pub(crate) async fn ld_read<R>(mut reader: &mut R) -> Result<Option<Vec<u8>>, Er
 where
     R: AsyncRead + Send + Unpin,
 {
+    const MAX_ALLOC: usize = 1 << 20;
     let l: usize = match VarIntAsyncReader::read_varint_async(&mut reader).await {
         Ok(len) => len,
         Err(e) => {
@@ -20,7 +21,7 @@ where
             return Err(Error::Other(e.to_string()));
         }
     };
-    let mut buf = Vec::with_capacity(l as usize);
+    let mut buf = Vec::with_capacity(std::cmp::min(l as usize, MAX_ALLOC));
     reader
         .take(l as u64)
         .read_to_end(&mut buf)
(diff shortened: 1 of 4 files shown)


Stebalien:
Review commit by commit.

vyzo (approved):
Not particulalry well versed in this, but it looks fine to me.

ipld/car/src/util.rs
```diff
@@ -20,7 +21,7 @@ where
             return Err(Error::Other(e.to_string()));
         }
     };
-    let mut buf = Vec::with_capacity(l as usize);
+    let mut buf = Vec::with_capacity(std::cmp::min(l as usize, MAX_ALLOC));
```
dignifiedquire, line 24: I think `read_to_end` will still allocate, so you should limit `l` as well.

Stebalien: Hm. I'm less concerned about that unless we're reading from a compressed archive, but fair point.

Stebalien: Hm. No, there's no real correct solution here. I think it's safe to assume that the caller will limit the input. Ideally, the caller would be able to pass in a maximum block size, but that's a more invasive change and isn't necessary for our purposes.

dignifiedquire: I solved it like this 
- check `l` if it is in range, if not err out
- if buffer length is smaller than `l`, grow 
- use `.read_exact(&mut buf[..l])` to do the actual reading (this nicely deals with EOFs as well)

Stebalien: `l` is in range of _what_? There is no defined maximum size for blocks in CAR files.

dignifiedquire: I went with 4MiB for now, matching block size limits, but it is not ideal. I will probably go see how to avoid having to read it entirely at all, and just stream anything larger through to remove the limit, without increasing memory usage.

Stebalien: Yeah, I think this is good enough for us for now, but we likely do need real block limits later (or something like that). But that's a more general problem where we just assume that all blocks can fit in memory.

PR #661, opened 2022-07-12
chore: remove go interop feature from AMT

Stebalien:
I'm not entirely sure why we needed it in the first place, but it looked related to gas usage. At this point, we can make a breaking change and simplify the code.

This will get bundled into the next actors release.


Stebalien:
@ZenGround0 I'd appreciate a review here as this will affect actors, and I'm still not entirely sure why this feature was necessary in the first place.

Stebalien:
Hm. Ok, I think I see what's going on here and I'm not sure if this is correct.

PR #664, opened 2022-07-19
Let integration tests generic over externs

mriise:
@tchataigner

### DIFF
--- a/testing/integration/examples/integration.rs
+++ b/testing/integration/examples/integration.rs
@@ -1,4 +1,5 @@
 use fvm::executor::{ApplyKind, Executor};
+use fvm_integration_tests::dummy::DummyExterns;
 use fvm_integration_tests::tester::{Account, Tester};
 use fvm_ipld_blockstore::MemoryBlockstore;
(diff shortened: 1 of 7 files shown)


tchataigner:
Thank you for that it's awesome ! Sorry for the delay :( Opened a [PR](https://github.com/mriise/ref-fvm/pull/2) on yours to add mock behavior to those dummies 👍

mriise:
thanks! @tchataigner sorry for the delay as well :\

testing/integration/examples/integration.rs
```diff
@@ -50,7 +51,7 @@ pub fn main() {
         .unwrap();
 
     // Instantiate machine
-    tester.instantiate_machine().unwrap();
+    tester.instantiate_machine(DummyExterns).unwrap();
```
raulk, line 54: Maybe we could do some magic here with a default implementation for the Externs type param, but this is good enough ;-)

raulk (approved):

tchataigner (approved):
I think it does the job ! Thank you for handling it !

PR #666, opened 2022-07-25
Account Abstraction Prototyping

mriise:


fvm/src/call_manager/default.rs
```diff
@@ -471,6 +477,153 @@ where
...
 
+    fn tst_validate(&mut self, msg: Message, sig: Signature) -> Result<()> {
+        // Lookup the actor.
+        let state = self
+            .state_tree()
+            .get_actor_id(msg.from.id().unwrap())?
+            .ok_or_else(|| syscall_error!(NotFound; "actor does not exist: {}", msg.from))?;
+
+        // Builtin or external?
+        // TODO
```
raulk, line 488: We will need some form of caching to memoize if a given sender code CID exposes a validate entrypoint or not. That way, we don't have to bootstrap an entire Wasm instance only to find out that the actor couldn't act as a sender. Could you open an issue to track this, and maybe jot down some ideas?

fvm/src/call_manager/default.rs
```diff
@@ -471,6 +477,153 @@ where
...
+
+        // Increment invocation count
+        self.invocation_count += 1;
+
+        // This is a cheap operation as it doesn't actually clone the struct,
+        // it returns a referenced copy.
+        let engine = self.engine().clone();
+
+        log::trace!("validating message {} -> {}::{}", msg.to, msg.from, sig.sig_type);
+        self.map_mut(|cm| {
```
raulk, line 505: I assume there's a bunch of logic we can deduplicate here? Do you want to venture and do it?

fvm/src/call_manager/default.rs
```diff
@@ -471,6 +477,153 @@ where
...
+            let result = result.and_then(|ret_id| {
+                Ok(if ret_id == NO_DATA_BLOCK_ID {
+                    None
+                } else {
+                    Some(block_registry.get(ret_id).map_err(|_| {
+                        Abort::Exit(
+                            ExitCode::SYS_MISSING_RETURN,
+                            String::from("returned block does not exist"),
+                        )
+                    })?)
```
raulk, line 570: Could you create an issue as a holding pen for the error conditions we need to think through and model for?

fvm/src/call_manager/default.rs
```diff
@@ -471,6 +477,153 @@ where
...
+                })
+            });
+
+            // Process the result, updating the backtrace if necessary.
+            let ret = match result {
```
raulk, line 575: We might need a special entry in the backtrace to represent the validation.

fvm/src/call_manager/mod.rs
```diff
@@ -121,6 +123,9 @@ pub trait CallManager: 'static {
         self.gas_tracker_mut().apply_charge(charge)?;
         Ok(())
     }
+
+    /// TODO
+    fn validate(&mut self, msg: Message, sig: Signature);
```
raulk, line 128: Why do we have a signature here as a parameter? The whole point of account abstraction is that the signature is opaquely inside the message (if the validation logic is signature driven)

fvm/src/kernel/default.rs
```diff
@@ -376,6 +376,12 @@ where
     fn msg_value_received(&self) -> TokenAmount {
         self.value_received.clone()
     }
+
+    fn msg_validate(&self, msg: Message, sig: Signature) -> Result<GasSpec> {
+        self.msg_caller();
+
+        todo!()
+    }
```
raulk, line 384: I don't understand what this method would be for. Maybe add some docs? What are you thinking?

shared/src/crypto/signature.rs
```diff
@@ -30,6 +30,8 @@ pub const SECP_PUB_LEN: usize = 65;
 pub enum SignatureType {
     Secp256k1 = 1,
     BLS = 2,
+    // TODO, should this be a different number?
+    External = 256,
```
raulk, line 34: This is not correct. I think this type is used only in the `validate_signature` syscall. And there is no such thing as validating an external signature. Instead, the signed chain message needs to specify "Delegated" as the signature type at the client level:

- https://github.com/filecoin-project/lotus/blob/aa0aaccc243774763a27db0a694f8014f1e3ea53/chain/types/signedmessage.go#L45-L48
- https://github.com/filecoin-project/go-state-types/blob/master/crypto/signature.go#L14-L19

This would prompt the FVM to delegate the signature to the validate entrypoint in the sender actor.

shared/src/message.rs
```diff
@@ -101,3 +101,9 @@ impl<'de> Deserialize<'de> for Message {
         })
     }
 }
+
+pub struct GasSpec {
+    gas_limit: i64,
+	gas_fee_cap: TokenAmount,
+	gas_premium: TokenAmount,
+}
```
raulk, line 109: The FVM needs to apply this gas spec.

fvm/src/call_manager/default.rs
mriise: I think an open issue for this, though it will probably end up being a new field added to `ActorState` to be checked when calling `.get_actor()` though this may be a good place to cache calls into `upgrade` as well.

fvm/src/call_manager/default.rs
mriise: there is. I don't want to say yes just yet since while they do have lots of similar code, their flows could end up being different (including the upgrade call in this as well)

fvm/src/call_manager/default.rs
mriise: yes this was copy-pasta'd from `send_resolved`, this will be removed for the time being though.

fvm/src/call_manager/mod.rs
mriise: `Message` represents and unsigned message, so the signature is deserialized elsewhere (in user code i believe since `Signature` is only used for the SDK validate_signature) and passed in. Would the signature be directly deserialized from `params` then? from what I can tell, the FVM does nothing automatically validate a signature, which would mean mostly just adding some internal functions to kick off a WASM module when it gets a signature type `External`.

fvm/src/kernel/default.rs
mriise: leftover from some thoughts that I abandoned yesterday.

shared/src/crypto/signature.rs
mriise: `Delegated` is a much better signature type name. There was some confusion about signature validation method calls (deal signatures, which is validated by `validate_signature`  _is not_ message validation that happens in lotus etc.)

PR #669, opened 2022-08-02
add thread random generation of machine IDs 

mriise:
Getting chain randomness of current epoch is not allowed. Adds `rand` dep for now till better solutions are made.


raulk (approved):
Ok temporarily, but we need an issue to track the real solution.

mriise:
> Ok temporarily, but we need an issue to track the real solution.

 #670

Stebalien (approved):
This works. I believe the "fix" would actually be to draw randomness from the parent tipset (that's the tipset that includes the messages being executed by this machine), but using a generated randomness (or counter, etc.) may be a better solution anyways.

PR #702, opened 2022-08-10
chore: switch to rust stable

Stebalien:


rust-toolchain
```diff
@@ -1 +1 @@
-nightly-2022-05-09
+stable
```
vmx, line 1: I'd pin it to a specific stable version. I usually use the least possible version, which is a good indication for others which version they need. Another reason is that it's good to compile without warnings and between table releases the Clippy warnings often change. So the warnings will depend on whatever stable version someone has currently installed.

mriise: MSRV can be specified with `rust = "1.30"` in cargo.toml, I think this should be `stable` so we are _also_ saying we expect this to compile with stable (which i feel is reasonable), though keeping clippy consistent does make sense.

PR #703, opened 2022-08-10
Add new accepted hash functions to the `hash` syscall (sha256, keccak256, ripemd160)

mriise:
- [x] sha256
- [x] sha3FIPS256 (Keccak256)
- [x] [ripemd160](https://github.com/multiformats/rust-multihash/pull/226) ~~(waiting on new release of rust-multihash)~~
- [x] new integration test for the syscall


fvm/src/kernel/default.rs
```diff
@@ -463,24 +464,21 @@ where
...
-        Ok(digest)
+        // Supported hash functions are in shared::FvmHashCode
+        let code = FvmHashCode::try_from(code).map_err(|e| {
+            if let multihash::Error::UnsupportedCode(code) = e {
+                syscall_error!(IllegalArgument; "unsupported hash code {}", code)
+            } else {
+                // will only error unsupported code, since that is all we are checking
+                // https://github.com/multiformats/rust-multihash/blob/f805963970aa509ab4e0175d296eaf5f4d8dde92/derive/src/multihash.rs#L255
+                unreachable!()
+            }
```
raulk, line 479: This should cleanly return an `AssertionFailed` error number. That error number is designed precisely for situations where assertions fail.

fvm/src/kernel/mod.rs
```diff
@@ -245,7 +246,7 @@ pub trait CryptoOps {
     /// `digest_out`, returning the size of the digest written to `digest_out`. If `digest_out` is
     /// to small to fit the entire digest, it will be truncated. If too large, the leftover space
     /// will not be overwritten.
-    fn hash(&mut self, code: u64, data: &[u8]) -> Result<[u8; 32]>;
+    fn hash(&mut self, code: u64, data: &[u8]) -> Result<MultihashGeneric<32>>;
```
raulk, line 249: Note: this should not return a multihash, but rather the digest itself, i.e. no self-describing prefix. This should probably be a Vec<u8> to account for future digest functions returning more than 32 bytes.

shared/src/crypto/hash.rs
```diff
@@ -0,0 +1,14 @@
+use multihash::derive::Multihash;
+use multihash::{Blake2b256, Keccak256, Sha2_256};
+
+#[derive(Clone, Copy, Debug, Eq, Multihash, PartialEq)]
+#[mh(alloc_size = 32)]
+pub enum FvmHashCode {
```
raulk, line 6: Let's call this enum `SupportedHashes` -- that's the role it plays in the system. `FvmHashCode` is a poor name because it carries little information besides what can be inferred from the context:

1. We know we're talking about the fvm because this is inside the `fvm_shared` crate, so `Fvm` is superfluous.
2. We can easily infer this is a code for a hash function from everything else.

shared/Cargo.toml
```diff
@@ -43,7 +43,7 @@ byteorder = "1.4.3"
 rand = "0.8"
 rand_chacha = "0.3"
 serde_json = "1.0.56"
-multihash = { version = "0.16.1", default-features = false, features = ["blake2b", "multihash-impl"] }
+multihash = { version = "0.16.1", default-features = false, features = ["blake2b", "multihash-impl", "sha2", "sha3"] }
```
raulk, line 46: We will need to require the `ripemd` feature here.

fvm/src/kernel/default.rs
```diff
@@ -463,24 +464,21 @@ where
...
+        })?;
+        Ok(code.digest(data))
```
raulk, line 481: Beware that you're changing the contract of this syscall! It used to return the digest, now it's returning a multihash. Is this what you intended?

raulk (changes requested):

raulk (commented):
@mriise @vmx just released multihash 0.16.3 for us: https://github.com/multiformats/rust-multihash/releases/tag/v0.16.3

Thanks, @vmx!

fvm/src/kernel/default.rs
mriise: I would disagree, this doesn't change the actual syscall, but it does change how we are handling it internally. really all we care about is having owned bytes and a length, which `MultihashGeneric` provides without allocating

fvm/src/kernel/mod.rs
mriise: This can be bumped to 64 bytes as well, is there an expectation that we will be adding larger hashes than that in the future that we would need to allocate for? We don't care for the prefix, true, but we do care about owned bytes and a length (can also pass in mutable bytes). Using an existing struct for that is easy and we dont have to resort to allocating.

Stebalien: I would bump to 64 bytes (512 hashes aren't uncommon) and leave it at that.

fvm/src/kernel/default.rs
Stebalien: Yeah, this changes the kernel, not the syscall.

fvm/src/kernel/mod.rs
raulk: Makes sense.

fvm/src/kernel/default.rs
raulk: Oh yeah, I missed the patch on the syscall handler that copies only the digest over, apologies. Works for me.

Stebalien:
Can we get a new integration test?

raulk (approved):
Thanks for the integration test and the follow-up work!

mriise:
note: it is very odd that clippy was silent about this locally, maybe it was removed in latest toolchain update?

Kubuxu:
For context ripemd160 was added as it is used by Bitcoin for address generation (for some random reason).

PR #707, opened 2022-08-12
introduce a crypto::recover_secp_public_key syscall.

raulk:
Supersedes https://github.com/filecoin-project/ref-fvm/pull/627 with a clean diff, refresh from master, and some small changes (check commit log).

### DIFF
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -352,6 +352,8 @@ where
         let engine = self.engine().clone();
 
         // prepare actor code
+        // TODO remove once the EVM smart contract actor is integrated in the builtin-actors bundle
+        //  https://github.com/filecoin-project/ref-fvm/issues/693
         self.engine()
             .load_code_cid(&state.code, self.blockstore())
             .map_err(
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -351,6 +351,13 @@ where
         // it returns a referenced copy.
         let engine = self.engine().clone();
 
+        // prepare actor code
+        self.engine()
+            .load_code_cid(&state.code, self.blockstore())
+            .map_err(
+                |_| syscall_error!(NotFound; "actor code cid does not exist {}", &state.code),
+            )?;
+
         log::trace!("calling {} -> {}::{}", from, to, method);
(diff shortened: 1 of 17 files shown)


fvm/src/call_manager/default.rs
```diff
@@ -351,6 +351,15 @@ where
...
 
+        // prepare actor code
+        // TODO remove once the EVM smart contract actor is integrated in the builtin-actors bundle
+        //  https://github.com/filecoin-project/ref-fvm/issues/693
+        self.engine()
+            .load_code_cid(&state.code, self.blockstore())
+            .map_err(
+                |_| syscall_error!(NotFound; "actor code cid does not exist {}", &state.code),
+            )?;
+
```
raulk, line 362: Tracked in #708.

Stebalien: Isn't this part of a separate patch?

- Introduce a secp recover syscall.
- Code loading?

fvm/src/machine/engine.rs
```diff
@@ -208,6 +208,33 @@ struct Cache<K> {
 }
 
 impl Engine {
+    /// Attempts to load Actor Code CID from the blockstore and
+    /// instantiate a wasmtime Module from the WASM bytecode if the
+    /// given code CID is present, otherwise returns None.
+    ///
+    /// Implicitly caches instantiated modules.
+    pub fn load_code_cid<BS: Blockstore>(
```
Stebalien, line 216: `install_code`? We don't actually need to return it.

Stebalien (approved):
Logic LGTM. I'd do this in two PRs as one is a new syscall, and one supports EVM code loading. But I don't feel too strongly about that.

I'd also change `load_code` so we don't have to clone the module on every call (we don't use the return value).

mriise (approved):
LGTM, +1 on splitting this PR though (not the end of the world if it isnt though).

fvm/src/machine/engine.rs
raulk: SGTM. I'll move this part to a different patch and we can address there.

raulk:
The extra feature has been moved to #711 and this PR is now strictly about the recover_secp_public_key syscall.

PR #709, opened 2022-08-12
feat: add gas charges to the execution trace

Stebalien:
BREAKING:

- This adds a new execution event type.
- The GasCharge object is now `'static`, but internally uses a `Cow` for the name.

### DIFF
--- a/fvm/src/gas/charge.rs
+++ b/fvm/src/gas/charge.rs
@@ -1,20 +1,24 @@
 // Copyright 2019-2022 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
 
+use std::borrow::Cow;
+
 use super::Gas;
 
 /// Single gas charge in the VM. Contains information about what gas was for, as well
 /// as the amount of gas needed for computation and storage respectively.
-pub struct GasCharge<'a> {
-    pub name: &'a str,
+#[derive(Clone, Debug)]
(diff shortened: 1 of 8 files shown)


Stebalien:
Forward-port of #675.

fvm/src/gas/charge.rs
```diff
@@ -1,20 +1,24 @@
...
+
 use super::Gas;
 
 /// Single gas charge in the VM. Contains information about what gas was for, as well
 /// as the amount of gas needed for computation and storage respectively.
-pub struct GasCharge<'a> {
-    pub name: &'a str,
+#[derive(Clone, Debug)]
+pub struct GasCharge {
+    pub name: Cow<'static, str>,
```
mriise, line 12: Why not a String? are we trying to enforce that price names must be known at compiletime?

mriise (approved):
LGTM, will be nice not having a static lifetime on the whole struct

fvm/src/gas/charge.rs
Stebalien: I'm trying to avoid allocating per gas-charge.

For some history:

- The string used to be `'static`.
- Then we made it take a lifetime, but that meant I couldn't store it.
- Now I'm making it a `Cow` so we can store it and _usually_ not have to allocate (unless the name comes from an actor itself).

PR #711, opened 2022-08-14
m2 native: ensure that code for an actor is ready for execution.

raulk:
When the m2-native feature is enabled, the call manager now prepares the code for execution by calling the engine.

### DIFF
--- a/fvm/src/machine/engine.rs
+++ b/fvm/src/machine/engine.rs
@@ -208,6 +208,35 @@ struct Cache<K> {
 }
 
 impl Engine {
+    /// Loads an actor's Wasm code from the blockstore by CID, and prepares
+    /// it for execution by instantiating and caching the Wasm module. This
+    /// method errors if the code CID is not found in the store.
+    pub fn prepare_actor_code<BS: Blockstore>(
+        &self,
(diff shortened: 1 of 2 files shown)


raulk:
This is the extra feature from #707 that we wanted to review in a separate PR. I refactored the initial contribution to name things more accurately and to make the code easier to follow.

fvm/src/machine/engine.rs
```diff
@@ -242,8 +263,8 @@ impl Engine {
         }
     }
 
-    /// Load some wasm code into the engine.
-    pub fn load_bytecode(&self, k: &Cid, wasm: &[u8]) -> anyhow::Result<Module> {
+    /// Loads some Wasm code into the engine and prepares it for execution.
+    pub fn prepare_wasm_bytecode(&self, k: &Cid, wasm: &[u8]) -> anyhow::Result<Module> {
```
raulk, line 267: I found `load_bytecode` ambiguous; it sounds like we're loading code from the engine, not _into_ the engine. So I proposed a more unambiguous name here.

vyzo (approved):

mriise (approved):

PR #713, opened 2022-08-15
Move the builtin Type enum out of the SDK

### LINKED ISSUE #712: Remove built-in Type enum from SDK
The built-in actor `Type` enum needs to live in the built-in actors repo (https://github.com/filecoin-project/builtin-actors/issues/520), though a copy is necessary in the FVM too (to avoid importing it from actors).

After https://github.com/filecoin-project/builtin-actors/pull/547, the actors will have no reference to the FVM version of this enum. Remove it from the exported SDK to avoid confusion.

As part of this, change the return/parameter types of these syscall wrappers to pass through the bare i32, rather than wrapping in a `Type`.
- `fvm::actor::get_builtin_actor_type`
- `fvm::actor::get_code_cid_for_type`

Note that in the long run we'll want to get rid of this type concept and have the built-in actors rely on addresses and/or code CIDs to confirm actor identity and access. The syscalls should then be removed (and should never be exposed to non-built-in actors).

anorth:
The built-in `Type` enum has been copied into the built-in actors repo, so no longer needs to be part of fvm_shared.

- [x] Wait for https://github.com/filecoin-project/builtin-actors/pull/547 to be approved and merged first

Closes #712

### DIFF
--- a/fvm/Cargo.toml
+++ b/fvm/Cargo.toml
@@ -12,36 +12,38 @@ keywords = ["filecoin", "web3", "wasm"]
 crate-type = ["lib"]
 
 [dependencies]
-anyhow = "1.0.47"
-thiserror = "1.0.30"
-num-traits = "0.2"
-derive_builder = "0.11.2"
-num-derive = "0.3.3"
-cid = { version = "0.8.5", default-features = false, features = ["serde-codec"] }
-multihash = { version = "0.16.1", default-features = false }
 fvm_shared = { version = "0.9.0", path = "../shared", features = ["crypto"] }
 fvm_ipld_hamt = { version = "0.5.1", path = "../ipld/hamt"}
 fvm_ipld_amt = { version = "0.4.2", path = "../ipld/amt"}
 fvm_ipld_blockstore = { version = "0.1.1", path = "../ipld/blockstore" }
 fvm_ipld_encoding = { version = "0.2.2", path = "../ipld/encoding" }
-serde = { version = "1.0", features = ["derive"] }
-serde_tuple = "0.5"
-serde_repr = "0.1"
-lazy_static = "1.4.0"
-derive-getters = "0.2.0"
-derive_more = "0.99.17"
-replace_with = "0.1.7"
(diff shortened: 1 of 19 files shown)


Stebalien (approved):

raulk (commented):
Wait. I'm onboard with delinking builtin-actors from this type so we can evolve builtin-actors independently of FVM, but are you sure this is _exactly_ what we want?

1. Developers using the SDK will no longer have access to the Type, and will need to recreate it _yet again_.
2. We can't just remove it from `shared` because it's also used in https://github.com/filecoin-project/builtin-actors-bundler: https://github.com/filecoin-project/builtin-actors-bundler/pull/7. We can detangle that first.
3. Won't you _anyway_ need to make here so that your new actors can be recognised by the FVM and handled in the corresponding syscalls?

Stebalien:
So, I expect that _most_ developers won't actually need this, right? We weren't planning on making these methods public in M2.2.

anorth:
1. As Steb notes, I think we'd rather work towards no-one ever needing this. But I guess we're not convinced we'll be able to do that.
2. I didn't know about that. But should the bundler use the version in the builtin-actors repo? That's really the canonical one, I think
3. Yes, we'll still need to keep the FVM one up to date. But we can't support a dependency from the built-ins onto the constant in FVM. If you can depend on it from the VM, that would be sweet but I understand why you might not want to.

If we're not confident in 1 I can abandon this change. There's probably not too much risk of builtin-actor authors depending on the wrong one.

Stebalien:
> I didn't know about that. But should the bundler use the version in the builtin-actors repo? Didn't the bundler used to be in there?

Hm. I didn't think about that.

We moved the bundler out because it's released independently of the actors. Having it in the same repo was kind of annoying (especially because we were actively maintaining multiple actor versions).

Making the bundler depend on the actors themselves isn't great. But we can probably just remove `Type` from the bundler entirely and instead pass a string and an integer.

Stebalien:
Replaced with #812 (removed it entirely).

PR #714, opened 2022-08-15
fix:set sys code cid

hunjixin:


Stebalien:
Context?

hunjixin:
in sdk, use this tester to run testing(include actor install and create), for install and create operation, this tester need to find init_code_cid actor cache. this cache was build when instantiate_machine from this code cid. so when running install actor,  the message would got panic ( actor code not found). 

only init_code_cid can also work, sys_code_cid,accounts_code_cid is just added by hand

PR #715, opened 2022-08-16
shared: add EVM built-in actor type.

raulk:
Note: there's an active proposal to remove the type token entirely in #712. However, for the reasons outlined in the discussion, we need to retain it for now, unless we find a better indirection for actor typing (or we extinguish the need for it altogether).

This is needed to add the EVM actor (https://github.com/filecoin-project/builtin-actors/pull/517) to the built-in actors bundle.


vyzo (approved):
maybe `evmcontract`? But honestly, just evm is fine.

Stebalien (approved):

PR #716, opened 2022-08-18
Change TokenAmount from a type alias to a struct type wrapping BigInt

### LINKED ISSUE #704: Change TokenAmount type to a new-type
The TokenAmount type is currently just an alias for BigInt. This means values can usually be transparently interpreted, compared etc between TokenAmount and other types, despite TokenAmount having a logical interpretation as a decimal. This exposes significant potential for error, e.g. as I discovered in in https://github.com/filecoin-project/builtin-actors/pull/514

If we make it a struct type and then implement pass-through methods for the common operations, and explicit conversion to strings, big decimals etc, then downstream users (which includes all actors) will have to work hard to mess it up. As a bonus, we can implement serialisation so we don't have to wrap in `BigIntDe` everywhere it's used in a HAMT, etc.

I'm willing to attempt this if FVM people concur.

(If you don't, in built-in actors I'll probably do it anyway and hide the FVM's version of the type in our runtime layer).

anorth:
This came out pretty cleanly, without much churn in code elsewhere. Please review carefully. Suggestions for better method names are welcome.

I have implemented display in human units (i.e. with decimal point). I've not implemented parsing from string but that would make a nice follow-up for someone.

Closes #704

### DIFF
--- a/fvm/src/gas/outputs.rs
+++ b/fvm/src/gas/outputs.rs
@@ -48,7 +48,7 @@ impl GasOutputs {
 
         if out.gas_burned != 0 {
             out.over_estimation_burn = base_fee_to_pay * out.gas_burned;
-            out.miner_penalty = out.miner_penalty + (base_fee - base_fee_to_pay) * out.gas_burned;
+            out.miner_penalty += (base_fee - base_fee_to_pay) * out.gas_burned;
         }
         let required_funds = fee_cap * gas_limit;
         let refund =
--- a/fvm/src/gas/outputs.rs
+++ b/fvm/src/gas/outputs.rs
@@ -48,7 +48,7 @@ impl GasOutputs {
 
         if out.gas_burned != 0 {
             out.over_estimation_burn = base_fee_to_pay * out.gas_burned;
-            out.miner_penalty += (base_fee - base_fee_to_pay) * out.gas_burned;
+            out.miner_penalty = out.miner_penalty + (base_fee - base_fee_to_pay) * out.gas_burned;
         }
         let required_funds = fee_cap * gas_limit;
         let refund =
--- a/fvm/src/state_tree.rs
+++ b/fvm/src/state_tree.rs
@@ -541,13 +541,13 @@ impl ActorState {
                 syscall_error!(InsufficientFunds; "when deducting funds ({}) from balance ({})", amt, self.balance).into(),
             );
         }
-        self.balance = &self.balance - amt;
+        self.balance -= amt;
 
         Ok(())
     }
     /// Deposits funds to an Actor
     pub fn deposit_funds(&mut self, amt: &TokenAmount) {
-        self.balance = &self.balance + amt;
+        self.balance += amt;
     }
 }
 
--- a/fvm/src/state_tree.rs
+++ b/fvm/src/state_tree.rs
@@ -11,7 +11,6 @@ use fvm_ipld_encoding::tuple::*;
 use fvm_ipld_encoding::CborStore;
 use fvm_ipld_hamt::Hamt;
 use fvm_shared::address::{Address, Payload};
-use fvm_shared::bigint::bigint_ser;
 use fvm_shared::econ::TokenAmount;
 use fvm_shared::state::{StateInfo0, StateRoot, StateTreeVersion};
 use fvm_shared::{ActorID, HAMT_BIT_WIDTH};
@@ -522,7 +521,6 @@ pub struct ActorState {
     /// Sequence of the actor.
     pub sequence: u64,
     /// Tokens available to the actor.
-    #[serde(with = "bigint_ser")]
     pub balance: TokenAmount,
 }
 
@@ -543,13 +541,13 @@ impl ActorState {
                 syscall_error!(InsufficientFunds; "when deducting funds ({}) from balance ({})", amt, self.balance).into(),
             );
         }
-        self.balance -= amt;
+        self.balance = &self.balance - amt;
 
         Ok(())
     }
     /// Deposits funds to an Actor
     pub fn deposit_funds(&mut self, amt: &TokenAmount) {
-        self.balance += amt;
+        self.balance = &self.balance + amt;
     }
 }
 
--- a/fvm/src/state_tree.rs
+++ b/fvm/src/state_tree.rs
@@ -638,7 +638,7 @@ mod tests {
     use fvm_ipld_blockstore::MemoryBlockstore;
     use fvm_ipld_encoding::{CborStore, DAG_CBOR};
     use fvm_shared::address::{Address, SECP_PUB_LEN};
-    use fvm_shared::bigint::BigInt;
+    use fvm_shared::econ::TokenAmount;
     use fvm_shared::state::StateTreeVersion;
     use fvm_shared::{IDENTITY_HASH, IPLD_RAW};
     use lazy_static::lazy_static;
@@ -753,7 +753,7 @@ mod tests {
             ActorState::new(
                 *DUMMY_ACCOUNT_ACTOR_CODE_ID,
                 *DUMMY_ACCOUNT_ACTOR_CODE_ID,
-                BigInt::from(55),
+                TokenAmount::from_atto(55),
                 1,
             ),
         )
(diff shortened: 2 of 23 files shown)


fvm/src/gas/outputs.rs
```diff
@@ -48,7 +48,7 @@ impl GasOutputs {
 
         if out.gas_burned != 0 {
             out.over_estimation_burn = base_fee_to_pay * out.gas_burned;
-            out.miner_penalty += (base_fee - base_fee_to_pay) * out.gas_burned;
```
Stebalien, line 51: Implement https://doc.rust-lang.org/std/ops/trait.AddAssign.html

shared/src/econ/mod.rs
```diff
@@ -1,7 +1,345 @@
...
+            atto: BigInt::from(atto),
+        }
+    }
+
+    /// Creates a token amount from a quantity of indivisible units (10^-18 whole units).
+    pub fn from_atto_u64(atto: u64) -> Self {
+        Self {
+            atto: BigInt::from(atto),
+        }
+    }
```
Stebalien, line 48:
```suggestion
    /// Creates a token amount from a quantity of indivisible units (10^-18 whole units).
    pub fn from_atto(atto: impl Into<BigInt>) -> Self {
        Self {
            atto: atto.into(),
        }
    }
```

shared/src/econ/mod.rs
```diff
@@ -1,7 +1,345 @@
...
+
+    /// Creates a token amount from a quantity of whole units (10^18 indivisible units).
+    pub fn from_whole(tokens: i64) -> Self {
+        Self {
+            atto: BigInt::from(tokens) * Self::PRECISION,
+        }
```
Stebalien, line 54:
```suggestion
        Self::from_atto((tokens as i128) * (Self::PRECISION as i128))
```

Slightly faster, and it re-uses other methods.

shared/src/econ/mod.rs
```diff
@@ -1,7 +1,345 @@
...
+    }
+
+    pub fn zero() -> Self {
+        Self {
+            atto: BigInt::zero(),
+        }
+    }
```
Stebalien, line 61: Implement `num_traits::Zero` instead.

shared/src/econ/mod.rs
```diff
@@ -1,7 +1,345 @@
...
+impl Default for TokenAmount {
+    #[inline]
+    fn default() -> TokenAmount {
+        TokenAmount::zero()
+    }
+}
+
+impl fmt::Debug for TokenAmount {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        fmt::Display::fmt(self, f)
```
Stebalien, line 104: It might be helpful to keep the newtype here (pretty common in debug impls.

```suggestion
        write!(f, "TokenAmount({})", self)
```

But I don't feel strongly about this.

shared/src/econ/mod.rs
Stebalien: (It's nice to be able to just write `Zero::zero()` in places and not care about the specific type).

fvm/src/gas/outputs.rs
anorth: I know I could have, I chose not to for simplicity. Will you require it?

shared/src/econ/mod.rs
anorth: It's a bit of pain to have to import `num_traits::Zero` separately just to use the method, but ok done.

fvm/src/gas/outputs.rs
Stebalien: It's not _necessary_, but:

1. People expect it.
2. It can help with performance by reusing the allocation.

But we always add it in later.

shared/src/econ/mod.rs
Stebalien: Yeah, I agree. But it's what people are used to.

fvm/src/state_tree.rs
```diff
@@ -543,13 +541,13 @@ impl ActorState {
                 syscall_error!(InsufficientFunds; "when deducting funds ({}) from balance ({})", amt, self.balance).into(),
             );
         }
-        self.balance -= amt;
+        self.balance = &self.balance - amt;
```
Stebalien, line 544: E.g., this is a case where we now have to allocate a new "balance". `self.balance -= amt` wouldn't need that.

shared/src/sys/mod.rs
```diff
@@ -20,24 +22,24 @@ pub struct TokenAmount {
...
+        crate::econ::TokenAmount::from_atto((v.hi as u128) << 64 | (v.lo as u128))
     }
 }
 
 impl TryFrom<crate::econ::TokenAmount> for TokenAmount {
-    type Error = <crate::econ::TokenAmount as TryInto<u128>>::Error;
+    type Error = TryFromBigIntError<()>;
     fn try_from(v: crate::econ::TokenAmount) -> Result<Self, Self::Error> {
-        v.try_into().map(|v: u128| Self {
+        v.atto().try_into().map(|v: u128| Self {
```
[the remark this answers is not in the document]
anorth: Thanks, it was bugging me too. I was never going to figure out that the associated error type was the cause.

shared/src/lib.rs
```diff
@@ -83,13 +84,8 @@ pub trait NetworkParams {
     const MINING_REWARD_TOTAL: i64;
 
     /// Initial reward actor balance. This function is only called in genesis setting up state.
-    fn initial_reward_balance() -> BigInt {
-        BigInt::from(Self::MINING_REWARD_TOTAL) * Self::TOTAL_FILECOIN
```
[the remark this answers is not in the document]
Stebalien: Hm. We should investigate and maybe delete this file (it doesn't look used?)

shared/src/econ/mod.rs
```diff
@@ -1,7 +1,384 @@
...
+
+// Only a single div/rem method is implemented, rather than the full Div and Rem traits.
+// Division isn't a common operation with money-like units, and deserves to be treated carefully.
+impl TokenAmount {
+    #[inline]
+    pub fn div_rem(&self, other: impl Into<BigInt>) -> (TokenAmount, TokenAmount) {
+        let (q, r) = self.atto.div_rem(&other.into());
+        (TokenAmount { atto: q }, TokenAmount { atto: r })
+    }
+}
```
Stebalien, line 279: I have to wonder if we even _need_ this, or if users should be forced to call `.atto()` and do their own thing.

shared/src/econ/mod.rs
```diff
@@ -1,7 +1,384 @@
...
+        self.atto *= &other.atto;
+    }
+}
+
+impl<'a> MulAssign<&'a TokenAmount> for TokenAmount {
+    #[inline]
+    fn mul_assign(&mut self, other: &TokenAmount) {
+        self.atto *= &other.atto;
+    }
+}
```
Stebalien, line 269: The units here are nonsensical: `FIL * FIL = FIL^2`.

`MulAssign<number_like>` (u32, BigInt, etc.) would make sense.

Stebalien (changes requested):
LGTM except the new MulAssign implementations.

shared/src/econ/mod.rs
anorth: I think division is just likely enough to be relevant to some actors that it was worth adding this one method for it.

shared/src/econ/mod.rs
anorth: Thanks, that had been my intention, lost in some macro fiddling that I discarded. Done now for an `Into<BigInt>`.

shared/src/econ/mod.rs
```diff
@@ -1,7 +1,377 @@
...
+    impl<> Mul<i64> for TokenAmount;
+    impl<'a> Mul<i64> for &'a TokenAmount;
+}
+
+impl<T: Into<BigInt>> MulAssign<T> for TokenAmount {
+    #[inline]
+    fn mul_assign(&mut self, other: T) {
+        self.atto *= &other.into();
+    }
+}
```
Stebalien, line 262:
```suggestion
impl<T> MulAssign<T> for TokenAmount
where
    BigInt: MulAssign<T>,
{
    #[inline]
    fn mul_assign(&mut self, other: T) {
        self.atto *= other;
    }
}
```

shared/src/econ/mod.rs
Stebalien: Fair enough.

shared/src/econ/mod.rs
Stebalien: Will work for all relevant types, and avoid an allocation.

Stebalien (approved):
One small nit, otherwise LGTM.

PR #793, opened 2022-08-18
Return message origin in `vm::context()` syscall

### LINKED ISSUE #300: Add a syscall to obtain the originator of a call
The `msg_caller` syscall returns the immediate caller, i.e. the actor that called the current actor. This is all that built-in actors have needed up until now.

However, it's useful for expose the originator of the call (i.e. the account actor whose transaction initiated the call stack) for many use cases, e.g. escrow, loans, DeFi-like cases, staking, pledging, etc.

As a sidenote: Ethereum has two distinct opcodes for this: [`CALLER`](https://www.ethervm.io/#33) and [`ORIGIN`](https://www.ethervm.io/#32), so at minimum we will need this capability for EVM compatibility.

vyzo:
Closes #300

### DIFF
--- a/fvm/src/kernel/default.rs
+++ b/fvm/src/kernel/default.rs
@@ -366,6 +366,10 @@ where
         self.caller
     }
 
+    fn msg_origin(&self) -> Result<Option<ActorID>> {
+        self.resolve_address(&self.call_manager.origin())
+    }
+
     fn msg_receiver(&self) -> ActorID {
         self.actor_id
     }
--- a/fvm/src/kernel/default.rs
+++ b/fvm/src/kernel/default.rs
@@ -366,8 +366,8 @@ where
         self.caller
     }
 
-    fn msg_origin(&self) -> Result<Option<ActorID>> {
-        self.resolve_address(&self.call_manager.origin())
+    fn msg_origin(&self) -> ActorID {
+        self.call_manager.origin()
     }
 
     fn msg_receiver(&self) -> ActorID {
@@ -733,7 +733,7 @@ where
     // TODO(M2) merge new_actor_address and create_actor into a single syscall.
     fn new_actor_address(&mut self) -> Result<Address> {
         let oa = self
-            .resolve_to_key_addr(&self.call_manager.origin(), false)
+            .resolve_to_key_addr(&Address::new_id(self.call_manager.origin()), false)
             // This is already an execution error, but we're _making_ it fatal.
             .or_fatal()?;
 
--- a/fvm/src/kernel/default.rs
+++ b/fvm/src/kernel/default.rs
@@ -366,7 +366,7 @@ where
         self.caller
     }
 
-    fn msg_origin(&self) -> ActorID {
+    fn msg_origin(&self) -> (ActorID, &Address) {
         self.call_manager.origin()
     }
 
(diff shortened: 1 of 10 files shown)


fvm/src/syscalls/vm.rs
```diff
@@ -53,6 +53,12 @@ pub fn context(context: Context<'_, impl Kernel>) -> crate::kernel::Result<Invoc
 
     Ok(InvocationContext {
         caller: context.kernel.msg_caller(),
+        origin: context
+            .kernel
+            .msg_origin()
+            .or_fatal()?
```
raulk, line 59: Needs context

raulk (approved):
Lgtm, module failing tests. Also deserves a note on why we need the Result on the Kernel method despite a call like that being infallible in practice.

fvm/src/kernel/default.rs
```diff
@@ -366,6 +366,10 @@ where
         self.caller
     }
 
+    fn msg_origin(&self) -> Result<Option<ActorID>> {
+        self.resolve_address(&self.call_manager.origin())
```
Stebalien, line 370: This method charges gas. We should pre-resolve origin when we construct the call manager. Really, we just need to pass `sender_id` instead of `msg.from` in https://github.com/filecoin-project/ref-fvm/blob/ef808624f20611fb0e6079d9c8d24193d4f4d440/fvm/src/executor/default.rs#L69.

vyzo: ack, will do.

raulk: Hm, it doesn't? There is no syscall gas (we already charged it), and the kernel method doesn't charge explicit gas. However, I'm also fine with pre-resolving when constructing the call manager (would also involve less work when origin is called more than once in the call stack).

raulk: Regardless: I agree with @Stebalien. There's no reason for the origin in the CallManager to be an address. It can just be an ActorID and problem solved. Implemented here: https://github.com/filecoin-project/ref-fvm/commit/bfb70ef6e2f1f4a5199d070c61f2eac5a89a2e7f

fvm/src/kernel/default.rs
```diff
@@ -729,7 +733,7 @@ where
     // TODO(M2) merge new_actor_address and create_actor into a single syscall.
     fn new_actor_address(&mut self) -> Result<Address> {
         let oa = self
-            .resolve_to_key_addr(&self.call_manager.origin(), false)
```
raulk, line 732: Conformance tests are failing here because before we _already_ had a key address here, so we didn't query the state tree to resolve it from the ID address. Now, we are forced to query the state tree, which is also a small performance penalty. Let's track both the original sender address and the sender ID in the Call Manager.

The more correct solution would be to introduce an `ActorPointer` type that could be initialised from an ID or from another addresses, and would transparently handle and memoise resolutions.

raulk: That would also unify the mess of Actor ID, ID addresses, and non-ID addresses.

raulk (approved):
Moving forward with this.

fvm/src/kernel/default.rs
Stebalien: > Now, we are forced to query the state tree, and that part is not available in the test vector.

I'm confused. We resolve the origin in the Executor, so that _shouldn't_ be causing this issue.

Stebalien: Oh, I see. It's the other way around.

PR #799, opened 2022-08-23
Move multihash table from shared to fvm and replace with simple enum

mriise:
Avoid pulling in hasher implementations into actors

Also adds blake2b512 to integration tests


raulk (approved):
LGTM, removes a dependency that we added inadvertently.

PR #806, opened 2022-08-25
feat: release fvm_shared v2 alpha

Stebalien:
This will let us release a v1 release/v1 for nv17.


Stebalien:
This will hopefully unbreak master.

Stebalien:
No, this won't fix it. We need to merge https://github.com/filecoin-project/builtin-actors/pull/583, then port that to next, then update that here.

Stebalien:
I reverted this on master because we have changes that shouldn't be included in this release.

PR #808, opened 2022-08-26
change randomness type to fix length byte array

lyswifter:
this is a necessary underlying changing for [builtin actor pull request 560](https://github.com/filecoin-project/builtin-actors/pull/560), to give an fix length byte array for `Randomness` instead of a `u8` vector


Stebalien (approved):
LGTM. This changes the SDK, but doesn't affect the syscall itself.

PR #812, opened 2022-08-27
feat: remove actor "type" enum

Stebalien:
Instead:

- Determine type IDs from the manifest itself.
- Built out a manifest datastructure for the actor types we care about.

That way, we can just add more actor types to the builtin actors without having to change the FVM.


fvm/src/machine/manifest.rs
```diff
@@ -0,0 +1,129 @@
...
+        Manifest::new(vec)
+    }
+
+    /// Construct a new manifest from actor name/cid tuples.
+    pub fn new(iter: impl IntoIterator<Item = (String, Cid)>) -> anyhow::Result<Self> {
+        let mut by_name = HashMap::new();
+        let mut by_id = HashMap::new();
+        let mut by_code = HashMap::new();
+        for ((name, code_cid), id) in iter.into_iter().zip(1u32..) {
+            by_id.insert(id, code_cid);
```
vyzo, line 94: shouldnt we sort those?

vyzo (commented):
generally an improvement i think, but shouldnt we sort the namez to ensure consistency?

vyzo:
actually nevermind, it wont help when we add a new one, so better to take them in manifest order.

vyzo (approved):

fvm/src/machine/manifest.rs
Stebalien: We need to maintain ordering as they appear in the manifest. Basically, the actor bundle is the ultimate source of truth here.

fvm/Cargo.toml
```diff
@@ -45,6 +45,7 @@ rand = "0.8.5"
 
 [dev-dependencies]
 pretty_assertions = "1.2.1"
+fvm = { path = ".", features = ["testing"] }
```
Stebalien, line 48: The version-2 feature resolver will prevent this feature from infecting non-dev builds.

fvm/src/machine/manifest.rs
```diff
@@ -0,0 +1,163 @@
...
+            let name = name.into();
+            by_id.insert(id, code_cid);
+            by_code.insert(code_cid, id);
+            by_name.insert(name, code_cid);
+        }
+        let singletons = SINGLETON_ACTOR_NAMES
+            .iter()
+            .flat_map(|&k| by_name.get(k))
+            .copied()
+            .collect();
```
Stebalien, line 101: @vyzo I needed to make some changes here to get the tests to pass (needed to allow "partial" manifests for testing). Mind giving it another pass? (I'm sorry for force pushing, I didn't realize you had already reviewed).

vyzo: sure

fvm/src/machine/manifest.rs
```diff
@@ -0,0 +1,163 @@
...
+        };
+        Manifest::new(vec)
+    }
+
+    /// Construct a new manifest from actor name/cid tuples.
+    pub fn new(iter: impl IntoIterator<Item = (impl Into<String>, Cid)>) -> anyhow::Result<Self> {
+        let mut by_name = HashMap::new();
+        let mut by_id = HashMap::new();
+        let mut by_code = HashMap::new();
+        for ((name, code_cid), id) in iter.into_iter().zip(1u32..) {
```
vyzo, line 91: can we add a comment that this needs to start at 1 or disaster with the 0 is not found semantics?

sdk/src/actor.rs
```diff
@@ -72,25 +70,28 @@ pub fn install_actor(code_cid: &Cid) -> SyscallResult<()> {
 
 /// Determines whether the supplied CodeCID belongs to a built-in actor type,
 /// and to which.
-pub fn get_builtin_actor_type(code_cid: &Cid) -> Option<actor::builtin::Type> {
+pub fn get_builtin_actor_type(code_cid: &Cid) -> Option<i32> {
```
vyzo, line 73: i32?

vyzo (approved):
lg, just a couple of comments.

fvm/src/machine/manifest.rs
```diff
@@ -0,0 +1,163 @@
...
+#[cfg(any(feature = "testing", test))]
+const fn id_cid(name: &[u8]) -> Cid {
+    use std::mem;
+
+    use fvm_shared::{IDENTITY_HASH, IPLD_RAW};
+    use multihash::Multihash;
+
+    // This code is ugly because const fns are a bit ugly right now:
+    //
+    // 1. We can't drop the result, we need to forget it.
```
mriise, line 43:
```suggestion
    // 1. Compiler can't drop the result, we need to forget it manually. https://doc.rust-lang.org/error-index.html#E0493
```

fvm/src/machine/manifest.rs
mriise, line 91:
```suggestion
        for ((name, code_cid), id) in iter.into_iter().zip(1u32..) { // actor 0 is invalid, first actor builtin actor is 1
```

sdk/src/actor.rs
mriise: yes, after a certain point i32 were being converted into the type anyways.

mriise (approved):
just some commentary nits

sdk/src/actor.rs
Stebalien: u32 would be better, but I'm trying to avoid breaking the API too much. Once we've backported this to v2, we can change v3.

fvm/src/kernel/mod.rs
```diff
@@ -191,11 +191,11 @@ pub trait ActorOps {
     #[cfg(feature = "m2-native")]
     fn install_actor(&mut self, code_cid: Cid) -> Result<()>;
 
-    /// Returns whether the supplied code_cid belongs to a known built-in actor type.
-    fn get_builtin_actor_type(&self, code_cid: &Cid) -> Option<actor::builtin::Type>;
+    /// Returns the actor's "type" (if builitin) or 0 (if not).
```
raulk, line 194:
```suggestion
    /// Returns the actor's "type" (if built-in) or 0 (if not).
```

raulk: Would be useful to link to the enumeration to provide guidance on how to interpret this number.

raulk (approved):

anorth:
Thanks!

PR #816, opened 2022-08-29
Merge release/v2 into master

Stebalien:


Stebalien:
So, plan:

- Merge this locally with `--strategy=ours`.
- In the future, continuously merge release/v2 into master.

This will let us make PRs against v2 and forward port them.

- **Pro**: Easier to maintain.
- **Con**: Duplicate commits due to the merge.

Thoughts?

Stebalien:
Note: The _ideal_ solution would have been to put v3 changes in a feature branch... But that ship has sailed.

Stebalien:
This just isn't worth it.

jennijuju:
What’s happening here… anything I should be aware of?

Stebalien:
Nothing. This was a no-op PR to merge release/v2 into master to make future ports easier (with a merge, rather than a cherry-pick). But it would have lead to messy history.

For now, I'm just going to cherry-pick.

PR #827, opened 2022-09-01
fix: downgrade actors to v9 (in the fvm v2 integration tests)

Stebalien:


Stebalien:
Hm. Ok, this is going to be annoying. I _think_ I need to depend on the version on crates.io. But I'm also pretty sure that we can't publish the bundle to crates.io anymore (due to some feature shenanigans that I really don't want to deal with).

- We could fix that and start publishing to crates.io again.
- We could vendor a bundle as we do in lotus.

Stebalien:
Ah, so it looks like the fix is pretty easy. We just needed to _remove_ the `version` specifier... 🙄 .

jennijuju:
Lolol

PR #828, opened 2022-09-01
fix: avoid plucking integration test actors from target

Stebalien:
Instead, use the WASM_BINARY const exported by wasm-builder.


testing/integration/tests/fil-integer-overflow-actor/src/actor/mod.rs
```diff
@@ -0,0 +1,127 @@
+use cid::multihash::Code;
```
Stebalien, line 1: moved with no changes.

testing/integration/tests/fil-malformed-syscall-actor/src/actor.rs
```diff
@@ -0,0 +1,15 @@
+use fvm_sdk::sys::crypto::compute_unsealed_sector_cid;
```
Stebalien, line 1: moved with no changes

mriise (approved):

PR #832, opened 2022-09-01
feat: avoid importing the bundle into the integration tests

Stebalien:
This change:

- Moves the actor "bundle" dependency to the dev dependencies.
- Makes the `Tester` take a "bundle root cid".
- Exports a helper function for importing a bundle into a blockstore.

That way:

1. Users can instantiate integration tests with their _own_ bundle.
2. We can publish the integration testing _framework_ while testing against a dev bundle locally.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -7,7 +7,7 @@ members = [
     "testing/conformance",
     "testing/integration",
     "ipld/*",
-    "testing/integration/tests/*"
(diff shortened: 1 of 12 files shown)


anorth (approved):
Great, thanks. I was pretty worried about seeing actors v10 in the dependency tree when I was trying to use this yesterday.

PR #834, opened 2022-09-02
add hash_arr fn to SDK

mriise:
uses `([u8; 64], usize)` instead of a vec

### DIFF
--- a/sdk/src/crypto.rs
+++ b/sdk/src/crypto.rs
@@ -67,7 +67,6 @@ pub fn hash_blake2b(data: &[u8]) -> [u8; 32] {
 
 /// Hashes input data using one of the supported functions.
 pub fn hash(hasher: SupportedHashes, data: &[u8]) -> Vec<u8> {
-    // TODO, either this or ([u8; 64], usize)
     let mut ret = Vec::with_capacity(64);
 
     unsafe {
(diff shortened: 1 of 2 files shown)


Stebalien:
Any reason not to take a mutable slice? That would let us have one (versatile) function.

```rust
fn hash(code: SupportedHashes, data: &[u8], digest: &mut [u8]) -> usize;
```

mriise:
> Any reason not to take a mutable slice?

Not really, having a const sized array is nice for some things though.

sdk/src/crypto.rs
```diff
@@ -66,26 +66,28 @@ pub fn hash_blake2b(data: &[u8]) -> [u8; 32] {
...
+    let mut digest = Vec::with_capacity(64);
+    let written = hash_into(hasher, data, &mut digest);
 
+    assert!(written <= digest.capacity());
+    unsafe { digest.set_len(written) }
+    digest
+}
+
+/// Hashes input data using one of the supported functions into a buffer.
+pub fn hash_into(hasher: SupportedHashes, data: &[u8], digest: &mut [u8]) -> usize {
```
Stebalien, line 80: feel free to rename this back to hash

Stebalien (approved):

Issue #694, opened 2022-08-08
Account Abstraction Tracking Issue


mriise:
- [ ]  Endpoint for FFI to call into FVM to start up a validator
- [ ]  Syscall access constraining 
    - [ ]  simple gaurds added on each syscall
    - [ ] add flags where needed to indicate current constraints on the runtime
- [ ]  Add new syscalls
    - [ ]  validate_context()
    - [ ]  (and/or) origin() getter
- [ ]  Gas benchmarking of reasonable gas limit to prevent DOSing
- [ ]  Pass message payload's parameters into the actor
- [ ] Add utility "is this an account actor" field in kernel

TODO: add tracking issue
TODO: add component specific issues

karim-agha:
Accounts should be

```rust
struct Account {
   pubkey: [u8; 32],
   representations: Vec<Multihash>
}
```

Stebalien:
## Lotus Support

Walk through of the 4 places account abstraction will touch in lotus.

- [ ] Message Validation (pubsub)
    - [ ] Execute validate before forwarding the message.
        - Don't care about "value spent" here, we check that on block creation.
        - https://github.com/filecoin-project/lotus/blob/7663ec2bde2c4074c5a7cbdd8ce2ac4939cb3252/chain/messagepool/messagepool.go#L739-L759
        - https://github.com/filecoin-project/lotus/blob/7663ec2bde2c4074c5a7cbdd8ce2ac4939cb3252/chain/sub/incoming.go#L327-L377
    - [ ] Make validation async
        - May require a new message propegation topic? ideally not...
        - https://github.com/filecoin-project/lotus/blob/7663ec2bde2c4074c5a7cbdd8ce2ac4939cb3252/node/modules/services.go#L172-L199
    - [ ] Prioritize non-abstract account messages over abstract account messages.
        - A new topic would definitely make this easier.
    - [ ] Limit the number of messages in a chain that we validate.
        - Current proposal says one, we really want more than that.
        - But we still need *a* limit.
- [ ] Block Creation
    - Unsolved:
        - Requires a cached "gas limit" for the message unless we change the current account abstraction proposal to not return this from validate.
            - https://github.com/filecoin-project/FIPs/discussions/388#discussioncomment-3808952
        - Unclear what to do about checking "value spent" as value can also be spent by internal sends.
[...]

PR #848, opened 2022-09-08
Account Abstraction `validate` method

mriise:
attempt no. 3 

tracking #694 

- [x] ~~ffi entrypoint (TODO link PR)~~ later
- [x] validate method on ~~kernel~~ executor
- [x] linker trap on illegal calls (now a manual error asserted inside of `Kernel`)
- [x] caching wasm instance (things may just work:tm: here)
- [x] execution type value
- [x] validate params (XXX: I expect this to change a lot, for now whole FIL message)
- [x] fil message with evm payload format (spec work)
  - Honestly I would love to just have the whole message embedded in params, even if we end up duplicating values


fvm/src/call_manager/default.rs
```diff
@@ -179,6 +186,55 @@ where
...
+            // Do the actual validate.
+            self.validate_unchecked::<K>(from, params)
+        };
+
+        if self.machine.context().tracing {
+            self.trace(match &result {
+                Ok(InvocationResult::Return(v)) => ExecutionEvent::CallReturn(
+                    v.as_ref()
+                        .map(|blk| RawBytes::from(blk.data().to_vec()))
+                        .unwrap_or_default(),
```
Stebalien, line 220: Ah, so, is this the logic that would go away if we stopped returning the gas spec?

fvm/src/call_manager/default.rs
```diff
@@ -510,6 +574,153 @@ where
...
+        //     )?;
+
+        log::trace!("calling validate from {}", from);
+        self.map_mut(|cm| {
+            // Make the kernel.
+            let kernel = K::new(
+                cm,
+                block_registry,
+                from,
+                from,
```
Stebalien, line 620: This should either be "from" this actor and "to" the target actor, or "from" the system actor and "to" this actor. Honestly, I think the second may make more sense. I.e., the _system_ is calling us.

fvm/src/call_manager/default.rs
```diff
@@ -510,6 +574,153 @@ where
         })
     }
 
+    fn validate_unchecked<K>(&mut self, from: ActorID, params: Block) -> Result<InvocationResult>
```
Stebalien, line 577: This is a lot of duplicate code. Can we put most of it into some form of helper?

fvm/src/executor/mod.rs
```diff
@@ -43,6 +45,13 @@ pub trait Executor {
     fn flush(&mut self) -> anyhow::Result<Cid>;
 }
 
+pub trait ValidateExecutor: Executor {
+    type Validator: Kernel;
+
+    /// This is the entrypoint to validate a message from an abstract account.
+    fn validate_message(&mut self, msg: Message, sig: Vec<u8>) -> anyhow::Result<ValidateRet>;
+}
```
Stebalien, line 53: We can just add this as a method to `Executor`, unless I'm missing something.

fvm/src/executor/validator.rs
```diff
@@ -0,0 +1,132 @@
...
+use fvm_shared::message::Message;
+
+use super::{ApplyKind, ApplyRet, DefaultExecutor, Executor, ValidateExecutor, ValidateRet};
+use crate::call_manager::{CallManager, ExecutionType, InvocationResult};
+use crate::kernel::{Block, Context, ExecutionError};
+use crate::machine::Machine;
+use crate::Kernel;
+
+/// TODO try not to be stuck with Default, but it has methods methods i want for validate.
+pub struct DefaultValidateExecutor<K: Kernel>(pub DefaultExecutor<K>);
```
Stebalien, line 15: Yeah, let's just stick to a single type unless there's a strong reason to wrap this.

fvm/src/kernel/default.rs
```diff
@@ -200,18 +205,24 @@ where
     }
 
     fn set_root(&mut self, new: Cid) -> Result<()> {
+        crate::assert_validator!(self, "Validator can't set root.");
```
Stebalien, line 208: Shouldn't this be a local macro?

fvm/src/kernel/mod.rs
```diff
@@ -88,6 +92,11 @@ pub trait Kernel:
     fn machine(&self) -> &<Self::CallManager as CallManager>::Machine;
 }
 
+/// TODO
+pub trait Validator {
```
Stebalien, line 96: `ValidateOps`? Or, well, I'm not sure what we're using this for.

fvm/src/syscalls/vm.rs
```diff
@@ -49,27 +49,5 @@ pub fn abort(
 }
 
 pub fn context(context: Context<'_, impl Kernel>) -> crate::kernel::Result<InvocationContext> {
-    use anyhow::Context as _;
-
-    Ok(InvocationContext {
-        caller: context.kernel.msg_caller(),
```
Stebalien, line 55: we should probably remove all these random ops from the kernel while we're at it.

shared/src/message.rs
```diff
@@ -93,3 +93,27 @@ impl<'de> Deserialize<'de> for Message {
...
     }
 }
+
+pub mod params {
+    use serde::{Deserialize, Serialize};
+
+    use super::{Cbor, Message};
+
+    /// Params of the `validate` entrypoint, signature raw bytes and the whole filecoin message to be validated
+    #[derive(Debug, Serialize, Deserialize)]
```
Stebalien, line 103:
```suggestion
    #[derive(Debug, Serialize_tuple, Deserialize_tuple)]
```

shared/src/message.rs
```diff
@@ -93,3 +93,27 @@ impl<'de> Deserialize<'de> for Message {
...
+    pub struct ValidateParams {
+        pub signature: Vec<u8>,
```
Stebalien, line 105:
```suggestion
        #[serde(with = "serde_bytes")]
        pub signature: Vec<u8>,
```

fvm/src/kernel/default.rs
```diff
@@ -245,6 +256,14 @@ where
     }
 }
 
+impl<C> Validator for DefaultKernel<C>
```
Stebalien, line 259: Do we need a trait for this? It looks like a private method for the default kernel?

testing/integration/tests/fil-validate-actor/build.rs
```diff
@@ -0,0 +1,12 @@
+fn main() {
+    use substrate_wasm_builder::WasmBuilder;
+    WasmBuilder::new()
+        .with_current_project()
+        .import_memory()
+        .append_to_rust_flags("-Ctarget-feature=+crt-static")
+        .append_to_rust_flags("-Cpanic=abort")
+        .append_to_rust_flags("-Coverflow-checks=true")
+        // .append_to_rust_flags("-Clto=true")
```
Stebalien, line 9: Breaking something?

testing/integration/tests/fil-validate-actor/src/lib.rs
```diff
@@ -0,0 +1,41 @@
...
+    fvm_shared::message::params::ValidateParams,
+};
+
+#[cfg(not(target_arch = "wasm32"))]
+include!(concat!(env!("OUT_DIR"), "/wasm_binary.rs"));
+
+// /// Placeholder invoke for testing
+// #[cfg(target_arch = "wasm32")]
+// #[no_mangle]
+// pub fn invoke(_: u32) -> u32 {
```
Stebalien, line 15: still broken?

fvm/src/executor/validator.rs
```diff
@@ -0,0 +1,132 @@
...
+
+impl<K> ValidateExecutor for DefaultValidateExecutor<K>
+where
+    K: Kernel,
+{
+    type Validator = K;
+
+    /// validate a message from an abstract account with a delegate signature
+    fn validate_message(&mut self, msg: Message, sig: Vec<u8>) -> anyhow::Result<ValidateRet> {
+        const VALIDATION_GAS_LIMIT: i64 = i64::MAX; // TODO reasonable gas limit
```
Stebalien, line 46: Yeah... can you define a constant in fvm_shared? We can set it to `i64::MAX` for now, but we'll need to figure this out later.

fvm/src/executor/mod.rs
Stebalien: Hm. Actually, I think we have to. `execute_message` will need to call `validate_message` (or some form of it) on execution. Basically:

- Preflight stuff.
- Abstract validation (burns gas).
- Execution (burns more gas).

That is, of course, unless we go with the two-stage validation thing. I'm going to see if we can get a talk with anorth to discuss that, we're going to want to make decisions here sooner rather than later.

Stebalien:
> Honestly I would love to just have the whole message embedded in params, even if we end up duplicating values

1. We cannot duplicate the actual EVM message params, doubling on-chain storage of messages. No matter how much simpler that would make this.
2. The "params" section of the message needs to be ready to pass off to the receiving actor, so we can't just stuff the entire transaction in the params.

That means that the EVM actor _will_ need to piece the transaction back together from a signature and parameters, but that's _really_ not that hard.

fvm/src/call_manager/default.rs
mriise: hmm, yeah thats a good point, system actor -> us, though I will need to check any logic that gets done in syscalls to ensure this wouldn't backfire

fvm/src/call_manager/default.rs
mriise: my thinking as well, had prioritized getting things working at the time

fvm/src/executor/mod.rs
mriise: True, though in theory this doesn't need to be a super trait of Executor and could be used to assert that an executor is _only_ used for validate. Though I'm not to confident on its usefulness...

fvm/src/kernel/mod.rs
mriise: Just following the current system (sort of). Current use is different than the originally intended, where Validator was a trait only implemented on Kernels that meant to have it. `ValidateOps` is a bit closer to what it does now though.

testing/integration/tests/fil-validate-actor/build.rs
mriise: nope

fvm/src/syscalls/vm.rs
mriise: hmm, this was more in the direction of splitting context method into smaller structs defined in their own relative modules e.g. `NetworkContext`, `MessageContext` rather than removing ops entirely.

do you mean to remove these syscalls entirely and just use `.x_context` as the way to read them?

fvm/src/kernel/default.rs
mriise: macro is used in syscalls::charge_gas as well, though it could be made local and have that other one just be manually written.

fvm/src/call_manager/default.rs
mriise: As well as lines 675-710, yes. So the end return value is just to be the exit code the user wants to exit with, _not_ any block handle.

fvm/src/call_manager/default.rs
```diff
@@ -92,9 +94,14 @@ where
             backtrace: Backtrace::default(),
             exec_trace: vec![],
             invocation_count: 0,
+            execution_type: ExecutionType::Normal,
         })))
     }
 
+    fn set_execution_type(&mut self, execution_type: ExecutionType) {
+        self.execution_type = execution_type;
+    }
```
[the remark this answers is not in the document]
mriise: Hm, yeah this should never reasonably change during runtime.

fvm/src/executor/mod.rs
Stebalien: It likely won't be useful. In general, implement as a monolith _first_, then break things into smaller pieces once the boundaries become clear. Otherwise, you'll end up with the wrong boundaries, spend a lot of time trying to make it work, give up eventually, then end up with _really_ awkward code with the wrong abstractions (speaking from lots of experience doing this).

fvm/src/syscalls/vm.rs
Stebalien: I mean removing these kernel methods, specifically, and just sticking with the `x_context` methods. Yes.

Unless something is using them. It makes it easier to tell what's "relevant" and what's not.

mriise: Doing this in a different PR since its a bit out of scope for this one.

fvm/src/call_manager/default.rs
mriise: sorry i misread the file mentioned here, no i think that this has quite a lot of differences between this and execute_unchecked, which will be accented if we remove the return value from validate. Leaving this as-is for now.

Stebalien:
The code and design have deviated substantially, so I'm closing this for now.

PR #849, opened 2022-09-08
feat: optimize no-op bitfield operations

Stebalien:
We have quite a few bitfield operations where one bitfield is empty.
Optimize them.

### DIFF
--- a/ipld/bitfield/src/lib.rs
+++ b/ipld/bitfield/src/lib.rs
@@ -297,6 +297,12 @@ impl BitField {
                 .all(|bit| self.unset.contains(&bit))
     }
 
+    /// Returns `true` if the bit field is _trivially_ empty. This is significantly faster than
+    /// checking if it's actually empty, but can be used to short-circuite certain operations.
+    fn is_trivially_empty(&self) -> bool {
+        self.set.is_empty() && self.ranges.is_empty()
+    }
+
     /// Returns a slice of the bit field with the start index of set bits
     /// and number of bits to include in the slice. Returns `None` if the bit
     /// field contains fewer than `start + len` set bits.
@@ -332,17 +338,18 @@ impl BitField {
 
     /// Returns the union of the given bit fields as a new bit field.
     pub fn union<'a>(bitfields: impl IntoIterator<Item = &'a Self>) -> Self {
-        bitfields.into_iter().fold(Self::new(), |a, b| &a | b)
+        bitfields.into_iter().fold(Self::new(), |a, b| a | b)
     }
 
     /// Returns true if `self` overlaps with `other`.
     pub fn contains_any(&self, other: &BitField) -> bool {
-        self.ranges().intersection(other.ranges()).next().is_some()
+        !(self.is_trivially_empty() || other.is_trivially_empty())
+            && self.ranges().intersection(other.ranges()).next().is_some()
     }
 
     /// Returns true if the `self` is a superset of `other`.
     pub fn contains_all(&self, other: &BitField) -> bool {
-        other.ranges().difference(self.ranges()).next().is_none()
+        other.is_trivially_empty() || other.ranges().difference(self.ranges()).next().is_none()
     }
 }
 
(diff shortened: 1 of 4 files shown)


ipld/bitfield/src/lib.rs
```diff
@@ -351,14 +358,46 @@ impl BitOr<&BitField> for &BitField {
...
+            BitField::from_ranges(self.ranges().union(rhs.ranges()))
+        }
     }
 }
 
 impl BitOrAssign<&BitField> for BitField {
     #[inline]
     fn bitor_assign(&mut self, rhs: &BitField) {
-        *self = &*self | rhs;
+        if !rhs.is_trivially_empty() {
```
Stebalien, line 389: saves a clone

ipld/bitfield/src/lib.rs
```diff
@@ -351,14 +358,46 @@ impl BitOr<&BitField> for &BitField {
...
+            *self = &*self | rhs;
+        }
+    }
+}
+
+impl BitOrAssign<BitField> for BitField {
+    #[inline]
+    fn bitor_assign(&mut self, rhs: BitField) {
+        // by-value to optimize the case where one is empty.
+        let s = std::mem::take(self);
```
Stebalien, line 399: also may save a clone

ipld/bitfield/src/lib.rs
```diff
@@ -367,7 +406,20 @@ impl BitAnd<&BitField> for &BitField {
 
     #[inline]
     fn bitand(self, rhs: &BitField) -> Self::Output {
-        BitField::from_ranges(self.ranges().intersection(rhs.ranges()))
+        if self.is_trivially_empty() || rhs.is_trivially_empty() {
+            BitField::new()
```
Stebalien, line 410: This is free.

ipld/bitfield/src/lib.rs
```diff
@@ -367,7 +406,20 @@ impl BitAnd<&BitField> for &BitField {
...
+        } else {
+            BitField::from_ranges(self.ranges().intersection(rhs.ranges()))
+        }
+    }
+}
+
+impl BitAnd<BitField> for BitField {
```
Stebalien, line 417: Implementing by-value ops encourages users to perform by-value ops where possible (possibly leading to optimizations).

jennijuju:
(are we supposed to backport this to `release/v2`? (I can totally do it just wanna confirm we should

Stebalien:
> (are we supposed to backport this to release/v2? (I can totally do it just wanna confirm we should

Ah.... no. This create isn't actually a part of the FVM, we just stashed it in this repo because we're LAAAAAAAAAAAAAAZY.

We can safely release this from master.

Stebalien:
We've discussed breaking all the "ipld" creates into a new repo, but haven't because that's more work.

ipld/bitfield/src/lib.rs
ZenGround0: I don't quite follow.  The `|` op on 430 uses the `bitor`'s defined above.  Are you saying `impl BitOr<&BitField> for &BitField {` is less efficient than `impl BitOr<BitField> for &BitField {` apart from the clone?  All I can see different is the clone.

ipld/bitfield/src/lib.rs
```diff
@@ -351,14 +358,76 @@ impl BitOr<&BitField> for &BitField {
...
+        }
+    }
+}
+
+impl BitOr<&BitField> for BitField {
+    type Output = BitField;
+
+    #[inline]
+    fn bitor(self, rhs: &BitField) -> Self::Output {
+        if rhs.is_trivially_empty() {
```
ZenGround0, line 391: nit: unless I'm missing some speedup gain it would be nice to check in a consistent order.  Other bitors do self, then rhs

ipld/bitfield/src/lib.rs
ZenGround0: This uses `impl BitAnd<&BitField> for &BitField {` under the hood so there's no speedup to using this instead correct?  But you're saying just letting them work with values directly can lead to user side optimizations?

ZenGround0 (approved):
I'm not so strong on rust's memory model so some of the tricks used here go over my head.  However it looks correct.

ipld/bitfield/src/lib.rs
```diff
@@ -297,6 +297,12 @@ impl BitField {
                 .all(|bit| self.unset.contains(&bit))
     }
 
+    /// Returns `true` if the bit field is _trivially_ empty. This is significantly faster than
+    /// checking if it's actually empty, but can be used to short-circuite certain operations.
```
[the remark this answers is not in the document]
Stebalien: Yep. Basically, if it returns true, it's definitely empty. But if it returns false, we won't be able to know without doing a bit of work.

ipld/bitfield/src/lib.rs
Stebalien: I'm trying to avoid a clone if both are trivially empty. It probably won't matter in most case so I'll switch it back.

ipld/bitfield/src/lib.rs
Stebalien: That's the difference. But I'll see if I can simplify a bit.

Stebalien:
@ZenGround0 I broke the ops into a separate file and implemented them on all ref/non-ref combinations. It's very verbose, but should be easier to reason about.

ipld/bitfield/src/lib.rs
Kubuxu: Which is what I’ve been doing recently. 😂

Kubuxu (approved):
Other than the XOR question, LGTM.

PR #850, opened 2022-09-08
feat: make BitField::new const

Stebalien:
This won't actually improve performance, but may help with some ergonomics.


Stebalien:
Guess that isn't stable yet.

mriise:
rustlang tracking issue for this: https://github.com/rust-lang/rust/issues/71835

Stebalien:
cleaning up old branches...

fridrik01 (approved):

PR #860, opened 2022-09-09
feat: add f4 address support

Stebalien:
We still need to decide on a final length, but this lets us represent f4 addresses.

I also took the opportunity to optimize things a bit.

### DIFF
--- a/shared/src/address/mod.rs
+++ b/shared/src/address/mod.rs
@@ -36,6 +36,9 @@ pub const SECP_PUB_LEN: usize = 65;
 /// BLS public key length used for validation of BLS addresses.
 pub const BLS_PUB_LEN: usize = 48;
 
+/// Max length of f4 sub addresses.
+pub const MAX_SUBADDRESS_LEN: usize = 32; // TODO: decide on a final length.
+
 /// Defines first available ID address after builtin actors
 pub const FIRST_NON_SINGLETON_ADDR: ActorID = 100;
 
@@ -53,7 +56,7 @@ lazy_static::lazy_static! {
 /// Length of the checksum hash for string encodings.
 pub const CHECKSUM_HASH_LEN: usize = 4;
 
-const MAX_ADDRESS_LEN: usize = 84 + 2;
+const MAX_ADDRESS_LEN: usize = 84 + 2; // TODO: update for f4
 const MAINNET_PREFIX: &str = "f";
 const TESTNET_PREFIX: &str = "t";
 
@@ -117,6 +120,16 @@ impl Address {
             network: NETWORK_DEFAULT,
             payload: Payload::Actor(address_hash(data)),
         }
+    }
+
+    pub fn new_namespaced(ns: ActorID, subaddress: &[u8]) -> Result<Self, Error> {
+        Ok(Self {
+            network: NETWORK_DEFAULT,
+            payload: Payload::Namespaced {
+                namespace: ns,
+                subaddress: subaddress.try_into()?,
+            },
+        })
     }
 
     /// Generates new address using BLS pubkey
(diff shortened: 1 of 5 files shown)


shared/src/address/mod.rs
```diff
@@ -213,50 +259,74 @@ impl FromStr for Address {
...
+                }
+
+                // Validate checksum.
+                if blake2b_simd::Params::new()
+                    .hash_length(CHECKSUM_HASH_LEN)
+                    .to_state()
+                    .update(&[protocol as u8])
+                    .update(payload)
+                    .finalize()
+                    .as_bytes()
```
Stebalien, line 321: Doing this manually lets us avoid intermediate allocations.

Stebalien: (and pulling it into a helper doesn't really buy us much).

shared/src/address/payload.rs
```diff
@@ -3,9 +3,63 @@
...
 
-use super::{from_leb_bytes, to_leb_bytes, Error, Protocol, BLS_PUB_LEN, PAYLOAD_HASH_LEN};
+use super::{
+    from_leb_bytes, to_leb_bytes, Error, Protocol, BLS_PUB_LEN, MAX_SUBADDRESS_LEN,
+    PAYLOAD_HASH_LEN,
+};
+use crate::ActorID;
+
+#[derive(Copy, Clone, Eq, PartialEq, Hash, Debug)]
+pub struct Subaddress {
```
Stebalien, line 16: Please bikeshed the name.

shared/src/address/payload.rs
```diff
@@ -19,30 +73,36 @@ pub enum Payload {
     Actor([u8; PAYLOAD_HASH_LEN]),
     /// BLS key address, full 48 byte public key
     BLS([u8; BLS_PUB_LEN]),
+    /// Namespaced
+    Namespaced {
```
Stebalien, line 77: Definitely hate this name. Just call it `f4`? Or...?

Also, `::Actor` is now misnamed because these are both "actor".

Maybe rename to:

- `Stable`
- `Managed` (?) idk

mriise: `Managed` sounds more appropriate, f4 addresses aren't guaranteed to be stable.

shared/src/address/payload.rs
mriise: `ManagedAddress`?
 
 Really new address are aliases\delegated redirects to an actor, but those names have added baggage that doesn't apply.

shared/src/address/mod.rs
```diff
@@ -36,6 +36,9 @@ pub const SECP_PUB_LEN: usize = 65;
 /// BLS public key length used for validation of BLS addresses.
 pub const BLS_PUB_LEN: usize = 48;
 
+/// Max length of f4 sub addresses.
+pub const MAX_SUBADDRESS_LEN: usize = 32; // TODO: decide on a final length.
```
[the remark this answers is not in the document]
raulk: That is, if we plan to use f4 addresses to support runtimes extracted from other chains. Other use cases may find this limit limiting (e.g. text addresses, although one could hash those and rely on clients to transform?).

shared/src/address/payload.rs
raulk: `Delegated`?

mriise: `Delegate` I was hoping to use for the signature type name in account abstraction, which may cause confusion. Though account abstraction isnt finished and we can bikeshed that name as well

Stebalien: Delegated matches IP address subrange delegation.

shared/tests/address_test.rs
```diff
@@ -398,6 +428,18 @@ fn invalid_byte_addresses() {
             input: bls_s,
             expected: Error::InvalidPayloadLength(47),
         },
+        // Namespace Protocol
```
mriise, line 431:
```suggestion
        // Delegate Protocol
```

mriise (approved):
LGTM outside of what is still being discussed in FIPs https://github.com/filecoin-project/FIPs/discussions/459

PR #862, opened 2022-09-11
Change TokenAmount::from_whole to take any Into<BigInt> parameter

anorth:
This explicit `u64` has been annoying downstream many times. Happily it should require no change in existing callers that pass u64.

FYI @ZenGround0


anorth:
@Stebalien should I pull into release/v2 branch after this ?

Stebalien (approved):

Stebalien:
Merged and backported (I was cutting a release anyways).

PR #873, opened 2022-09-13
chore: optimize dependencies when testing

Stebalien:


Stebalien:
Testing says... not worth it. Not sure why caching isn't helping more, but it isn't.

PR #883, opened 2022-09-16
Format RawBytes debug as hex string rather than array of decimals

anorth:
The derived representation of a list of decimals is very unhelpful when debugging, e.g. CBOR-encoded values. The hex string will at least be usable in an CBOR inspector, and I think usually not less useful for cases when the bytes are representing something else (tho since this is the CBOR library, that'll be uncommon).

FYI @ZenGround0 when debugging actors.

### DIFF
--- a/ipld/encoding/src/cbor.rs
+++ b/ipld/encoding/src/cbor.rs
@@ -1,6 +1,7 @@
 // Copyright 2019-2022 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
 
+use std::fmt::{Debug, Formatter};
 use std::ops::Deref;
 use std::rc::Rc;
 
@@ -29,7 +30,7 @@ impl<T> Cbor for Option<T> where T: Cbor {}
 
 /// Raw serialized cbor bytes.
 /// This data is (de)serialized as a byte string.
-#[derive(Clone, PartialEq, Debug, Serialize, Deserialize, Hash, Eq, Default)]
+#[derive(Clone, PartialEq, Serialize, Deserialize, Hash, Eq, Default)]
 #[serde(transparent)]
 pub struct RawBytes {
     #[serde(with = "serde_bytes")]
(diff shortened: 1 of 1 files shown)


Stebalien (approved):
Yeah, this is much more usable.

anorth:
I'm having trouble landing this due to:
```
error: unused import: `crate::RawBytes`
   --> ipld/encoding/src/cbor.rs:102:9
    |
102 |     use crate::RawBytes;
    |         ^^^^^^^^^^^^^^^
```

My local environment
```
active toolchain
----------------

nightly-2022-05-09-x86_64-apple-darwin (overridden by '/Users/me/plabs/filecoin-project/ref-fvm/rust-toolchain')
rustc 1.62.0-nightly (cb1219871 2022-05-08)
```
replicates this warning, but if I _remove_ the import then build fails with `failed to resolve: use of undeclared type `RawBytes` at the obvious places, and a recommendation to import it.

My IDE (1.63) also requires the import.

This may be related to the discussion in https://github.com/filecoin-project/builtin-actors/pull/660, but I can't see how to straightforwardly apply the same technique here. @mriise can you help?

mriise:
yeah! its a pretty simple fix :)

ipld/encoding/src/cbor.rs
```diff
@@ -86,3 +87,28 @@ impl RawBytes {
...
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "RawBytes {{ ")?;
+        for byte in &self.bytes {
+            write!(f, "{:02x}", byte)?;
+        }
+        write!(f, " }}")
+    }
+}
+
+mod test {
```
mriise, line 101:
```suggestion
#[cfg(test)]
mod test {
```

anorth: Thanks 🤦

Issue #894, opened 2022-09-19
Implement Ord for Address


ethan-storswift:
Implement ordering for Address in order to use them as keys in BTreemaps. 

See [issue 51 of buildin-actors](https://github.com/filecoin-project/builtin-actors/issues/51)

PR #895, opened 2022-09-19
add ordering to address

ethan-storswift:
add ordering to address so they can be used as keys in BTreeMaps #894


Stebalien (approved):
Seems reasonable. But note, you can _usually_ pre-resolve addresses to IDs and use those.

PR #896, opened 2022-09-19
feat: add some initialization helpers to the sdk

Stebalien:
This replaces the custom assert macros. Instead of overriding the assert macros, the user would:

1. Call `fvm_sdk::initialize()` at the top of their actor/tests.
2. Use the default `assert` macros.

replaces #660


vyzo (commented):
So what happens if the user doesnt call the initializer?

vyzo:
ok, nothing deleterious.

vyzo (approved):

PR #901, opened 2022-09-23
EVM-899: Hamt::flush short circuit on no change

aakoshh:
fixes https://github.com/filecoin-project/builtin-actors/pull/709

The PR adds an optional `flushed_cid` field to the `Hamt` which is used to remember the CID it was initialized with until something changes and the `Hamt` if flushed again. If the field is set during `flush`, we have nothing to do and can return the already calculated CID. 

This saves the effort of serializing the root node and passing the data through the Wasm machinery for the CID to be calculated when there was no change made to the contents of the map.

If this PR is acceptable, it would be worth doing the same for the `Amt` in another PR.


Stebalien (approved):
Seems like a good optimization. It'll only save one write, but that write could be significant.

Issue #636, opened 2022-06-24
Perform SHA256 hashes natively via a syscall


Stebalien:
HAMTs currently perform SHA256 hashing inside the actors (for hashing HAMT keys). We should be doing this via a syscall.

mriise:
`hash(data: &[u8], mh: Code)` ?

Stebalien:
Yep. We have the syscall defined here: https://github.com/filecoin-project/ref-fvm/blob/923a6846c203ceaed06dd683ccc4ed37dd738c68/fvm/src/syscalls/crypto.rs#L48

However, we'll have to change the HAMT constructor to take a hash function so that the HAMT will work both _inside_ and _outside_ of the FVM.

Stebalien:
Fixed

PR #902, opened 2022-09-24
Optimization for #636 | Perform SHA256 hashes natively via a syscall

shamb0:
Hello @Stebalien & @fvm-team,

Tried to refactor the code to optimize for [#636](https://github.com/filecoin-project/ref-fvm/issues/636),

Have taken implementation approach of ...

* Passing the context specific hash function as `HashAlgorithm` trait object to following methods ...

```
Hamt::set()
Hamt::set_if_absent()
Hamt::get()
Hamt::contains_key()
Hamt::delete()
```

* Verified the changes with existing unit test app. All are passed except one.

* Hit with one Unit test failure error, checking the the cause for failure is in-progress.

* Observed some regression on bench results, not sure about the cause,  suspecting HW configuration of the machine, investigation in-progress.

* Code is ready for review.

### DIFF
--- a/fvm/CHANGELOG.md
+++ b/fvm/CHANGELOG.md
@@ -22,6 +22,10 @@ See `release/v2`
     - blake2b512
 - API BREAKING: add gas charges to the execution trace.
 
+## 2.0.0-alpha.1
+
+Bump major version for breaking changes.
+
 ## 1.1.0 [2022-06-27]
 
 - debug execution: implement actor redirects in engine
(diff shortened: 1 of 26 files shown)


Stebalien:
Thanks for taking a stab at this. Unfortunately, this implementation would make using the Hamt pretty difficult. Luckily, the HAMT type already supports passing in a custom hash function via the type:

https://github.com/filecoin-project/ref-fvm/blob/1dcb84d6450e52b007a77158488edec968cdc471/ipld/hamt/src/hamt.rs#L35

So, we need to:

1. Implement `HashAlgorithm` using the FVM SDK.
2. Specify that `HashAlgorithm` type in https://github.com/filecoin-project/builtin-actors/blob/19e4000e884b8951306ef6780677d72c9ea75438/runtime/src/lib.rs#L40 if we're building with `cfg(feature = "fil-actor")`.

Stebalien:
> Observed some regression on bench results, not sure about the cause, suspecting HW configuration of the machine, investigation in-progress.

There's probably overhead with:

- Atomics (you're using an Arc).
- Passing an extra parameter.

Specifying the hash function in the type should have no overhead.

shamb0:
```
> Thanks for taking a stab at this. Unfortunately, this implementation would make using the Hamt pretty difficult. Luckily, the HAMT type already supports passing in a custom hash function via the type:
> 
> https://github.com/filecoin-project/ref-fvm/blob/1dcb84d6450e52b007a77158488edec968cdc471/ipld/hamt/src/hamt.rs#L35
> 
> So, we need to:
> 
>     1. Implement `HashAlgorithm` using the FVM SDK.
> 
>     2. Specify that `HashAlgorithm` type in https://github.com/filecoin-project/builtin-actors/blob/19e4000e884b8951306ef6780677d72c9ea75438/runtime/src/lib.rs#L40 if we're building with `cfg(feature = "fil-actor")`.
```

Sure, Thanks for the comment & suggestion, will re-work on it ...

PR #905, opened 2022-09-26
Support for iron opcodes

vyzo:
Companion to https://github.com/filecoin-project/builtin-actors/pull/713

TODO:
- [x] `syscalls::gas::available`
- [x] `syscalls::actor::actor_balance` -- renamed to `syscalls::actor::balance_of`
- [x] `syscalls::env::timestamp`
- [x] `syscalls::env::blockhash`
- [x] ~`syscalls::env::gas_limit`~ -- moved into the invocation context.
- [x] ~`syscalls::env::gas_price`~ -- added gas_premium to the invocation context.


fvm/src/syscalls/actor.rs
```diff
@@ -112,3 +113,11 @@ pub fn install_actor(
...
     context.kernel.install_actor(typ)
 }
+
+pub fn actor_balance(
+    _context: Context<'_, impl Kernel>,
+    _addr_off: u32,
+    _addr_len: u32,
+) -> Result<sys::TokenAmount> {
+    todo!()
+}
```
raulk, line 123: `balance_of` sounds like a better name to prevent the stutter (`actor::actor_balance`).

This seems relatively easy to implement.

fvm/src/syscalls/env.rs
```diff
@@ -0,0 +1,23 @@
+use super::Context;
+use crate::kernel::{Kernel, Result};
+
+pub fn timestamp(_context: Context<'_, impl Kernel>) -> Result<u64> {
+    todo!()
+}
```
raulk, line 6: This will require the chain environment data to be passed from the client.

fvm/src/syscalls/env.rs
```diff
@@ -0,0 +1,23 @@
...
+
+pub fn blockhash(
```
raulk, line 8: This should be `tipset_cid`, the block hash is something different than what this should return.

fvm/src/syscalls/env.rs
```diff
@@ -0,0 +1,23 @@
+use super::Context;
+use crate::kernel::{Kernel, Result};
+
+pub fn timestamp(_context: Context<'_, impl Kernel>) -> Result<u64> {
```
raulk, line 4: `tipset_timestamp`?

raulk, line 4: Just noting that we had discussed to put all this data behind a single struct, following the same patter as the `InvocationContext`, e.g. `ChainContext`. I'm happy to keep it as separate ops for now, it's always easy to add another syscall to retrieve all info at once for when the actor wants more than just a single field.

fvm/src/syscalls/env.rs
```diff
@@ -0,0 +1,23 @@
+use super::Context;
```
raulk, line 1: Not sure if `env` is the appropriate module name here. Could also be `chain`.

fvm/src/syscalls/env.rs
```diff
@@ -0,0 +1,23 @@
...
+    todo!()
+}
+
+pub fn gas_limit(_context: Context<'_, impl Kernel>) -> Result<u64> {
+    todo!()
+}
+
+pub fn gas_price(_context: Context<'_, impl Kernel>) -> Result<u64> {
+    todo!()
+}
```
raulk, line 23: This is not correct. We already have the base_fee under `network::base_fee`. The EVM `GASPRICE` opcode returns the `effective_gas_price`, which is:

	effective_gas_price = priority_fee_per_gas + block.base_fee_per_gas

The piece we're missing is the gas premium (in the invocation context?), and the EVM runtime would fetch both the base fee and the gas premium to sum them and honour the GASPRICE behaviour.

fvm/src/syscalls/gas.rs
```diff
@@ -16,3 +16,7 @@ pub fn charge_gas(
     // Gas charges from actors are always in full gas units. We use milligas internally, so convert here.
     context.kernel.charge_gas(name, Gas::new(compute))
 }
+
+pub fn available(_context: Context<'_, impl Kernel>) -> Result<u64> {
+    todo!()
+}
```
raulk, line 22: Should be trivial to implement. `context.kernel.gas_available()` already exists.

fvm/src/syscalls/actor.rs
```diff
@@ -112,3 +113,11 @@ pub fn install_actor(
     let typ = context.memory.read_cid(typ_off)?;
     context.kernel.install_actor(typ)
 }
+
+pub fn actor_balance(
+    _context: Context<'_, impl Kernel>,
+    _addr_off: u32,
+    _addr_len: u32,
```
Stebalien, line 120: I'd consider:

```suggestion
    actor_id: u64,
```

This will require an extra syscall to lookup the actor ID in some cases, but it matches our other syscalls. All the calls that take an _address_ do so because the target actor may not exist (yet).

fvm/src/syscalls/env.rs
```diff
@@ -0,0 +1,23 @@
+use super::Context;
+use crate::kernel::{Kernel, Result};
+
+pub fn timestamp(_context: Context<'_, impl Kernel>) -> Result<u64> {
+    todo!()
+}
+
+pub fn blockhash(
+    _context: Context<'_, impl Kernel>,
+    _block: u32,
```
Stebalien, line 10: `epoch: i64`

fvm/src/syscalls/env.rs
Stebalien: (also needs to be a token amount)

fvm/src/syscalls/env.rs
Stebalien: IMO, we'd also want to move some fields out of the invocation context (epoch, network version, etc.) into the ChainContext (or EpochContext?).

(I'd prefer todo this _sooner_ rather than later, if possible)

mriise:
There are overlaps here with what I need to do for ValidateContext (if at all). Some fields in `InvokationContext` are more related to the message or the network, I think at this point we can just make a `xContext` struct per module to keep seperation. This would split current `InvokationContext` into `MessageContext` and `NetworkContext`, the latter might just be combined with the this PR's `ChainContext` depending on what `Ops` it needs. For syscalls that are added here as one offs, I think we should leave them as-is.

Doing it like this sort of deflates the purpose of context, but having consolidated structs of metadata is useful still, and we can continue to cache the result in the sdk as needed.

fvm/src/syscalls/actor.rs
vyzo: ok, will rename.

fvm/src/syscalls/env.rs
vyzo: sure!

fvm/src/syscalls/env.rs
vyzo: ok.

fvm/src/syscalls/env.rs
vyzo: we can rename, let's discuss. I am keeping it for now, but let's resolve before merging.

fvm/src/syscalls/env.rs
vyzo: so you want to remove this syscall entirely? I can add the gas premium in the invocation context.

fvm/src/syscalls/actor.rs
vyzo: ok, will use the actor id.

fvm/src/syscalls/env.rs
vyzo: ACK

fvm/src/syscalls/actor.rs
vyzo: done and implemnted.

fvm/src/syscalls/actor.rs
vyzo: done.

fvm/src/syscalls/gas.rs
vyzo: implemented.

fvm/src/syscalls/env.rs
vyzo: I can also move the gas limit to the invocation context as well, since they are intrinsically related.

fvm/src/syscalls/env.rs
vyzo: Sure, but I'd rather do this in a subsequent pr.

fvm/src/syscalls/env.rs
vyzo: Ok, instead of passing them in the env context, I added gas premium as an argument to Send and fished the limit from the gas tracker.
I also added both of them to the invocation context struct.

fvm/src/call_manager/default.rs
```diff
@@ -96,6 +106,7 @@ where
         method: MethodNum,
         params: Option<Block>,
         value: &TokenAmount,
+        gas_premium: &TokenAmount,
```
raulk, line 109: This looks awkward and leaky here. All parameters to send are **necessary** to execute the send; this is merely contextual information. Let's not put it here, but instead extend the `GasTracker` to hold this value and add a getter for it.

fvm/src/call_manager/default.rs
```diff
@@ -385,9 +403,19 @@ where
...
-            let kernel = K::new(cm, block_registry, from, to, method, value.clone());
+            let kernel = K::new(
+                cm,
+                block_registry,
+                from,
+                to,
+                method,
+                value.clone(),
+                gas_premium.clone(),
+                gas_limit,
```
raulk, line 417: Doesn't the Kernel already have access to the `gas_limit` via the `call_manager.gas_tracker`? Why does it need to be handed off the field explicitly? (for premium, see above)

fvm/src/call_manager/mod.rs
```diff
@@ -42,7 +43,13 @@ pub trait CallManager: 'static {
     type Machine: Machine;
 
     /// Construct a new call manager.
-    fn new(machine: Self::Machine, gas_limit: i64, origin: (ActorID, Address), nonce: u64) -> Self;
+    fn new(
+        machine: Self::Machine,
+        gas_limit: i64,
+        origin: (ActorID, Address),
+        nonce: u64,
+        chain_context: ChainContext,
```
raulk, line 51: The ChainContext is a Machine level field, not a CallManager level field. Let's hoist it up and remove it from here. (The ChainContext won't change across message applications within a single Machine instantiation)

raulk: We can also move the `epoch` into the chain context, like @Stebalien suggested elsewhere.

fvm/src/kernel/default.rs
```diff
@@ -441,6 +456,29 @@ where
...
+
+    /// epoch tipset cid
+    fn tipset_cid(&self, epoch: i64) -> Result<Option<Cid>> {
+        if epoch < 0 {
+            return Err(syscall_error!(IllegalArgument; "epoch is negative").into());
+        }
+        let tipsets = &self.call_manager.chain_context().tipsets;
+        if (epoch as usize) < tipsets.len() {
+            return Ok(Some(tipsets[epoch as usize]));
+        }
```
raulk, line 476: For extra safety and formality, we should _also_ validate that the specific lookback is within the admitted range (256 for Ethereum, 900 if we want to follow Filecoin's finality).

fvm/src/kernel/mod.rs
```diff
@@ -72,13 +73,16 @@ pub trait Kernel:
...
+    #[allow(clippy::too-many-arguments)]
     fn new(
         mgr: Self::CallManager,
         blocks: BlockRegistry,
         caller: ActorID,
         actor_id: ActorID,
         method: MethodNum,
         value_received: TokenAmount,
+        gas_premium: TokenAmount,
+        gas_limit: u64,
```
raulk, line 85: See above. Can get the limit from the `call_manager.gas_tracker`, and we should add the premium there too.

fvm/src/kernel/mod.rs
```diff
@@ -345,3 +359,12 @@ pub trait DebugOps {
     /// Returns error on malformed name, returns Ok and logs the error on system/os errors.
     fn store_artifact(&self, name: &str, data: &[u8]) -> Result<()>;
 }
+
+/// Environment ops
+pub trait EnvOps {
```
raulk, line 364: For better naming consistency.

```suggestion
pub trait ChainOps {
```

sdk/src/sys/env.rs
```diff
@@ -0,0 +1,13 @@
+//! Syscalls for working with gas.
+
+super::fvm_syscalls! {
+    module = "env";
```
raulk, line 5:
```suggestion
    module = "chain";
```

raulk (changes requested):

fvm/src/call_manager/default.rs
vyzo: oh, didn't notice that -- yeah no need for it then.

fvm/src/call_manager/default.rs
vyzo: ok.

fvm/src/call_manager/mod.rs
vyzo: Ok, will hoist.

fvm/src/kernel/default.rs
vyzo: this should probably happen at call site? We don't know if it is an eth or fil invocation here, so we merely check against the length of the vector.

fvm/src/kernel/mod.rs
vyzo: ack.

fvm/src/kernel/mod.rs
vyzo: ok.

fvm/src/kernel/default.rs
raulk: The FVM should place a hard limit. I would vote for 900 epochs because it matches our finality. The EVM runtime should validate that the argument is 256 max.

raulk:
BTW @vyzo -- I agree with @mriise's comment above. Thinking about it, all attributes in `ChainContext` are actually network-related attributes that _must_ be provided by the client, so we can consolidate into the `network` syscall module and package, and into a `NetworkContext`, instead of accepting the package/namespace sprawl here.

vyzo:
ok

fvm/src/call_manager/default.rs
vyzo: done.

fvm/src/kernel/mod.rs
vyzo: done.

fvm/src/call_manager/default.rs
vyzo: done.

vyzo:
hoisted chain context stuffs into a NetworkContext struct in Machine, which includes the epoch as well.

fvm/src/call_manager/mod.rs
vyzo: done.

fvm/src/kernel/mod.rs
vyzo: done.

sdk/src/sys/env.rs
vyzo: done.

fvm/src/kernel/default.rs
vyzo: ok, 900 it is -- so we get an error (as opposed to a None) if we pass more than that.

raulk (commented):
_Almost_ right, just a bit more consolidation (NetworkOps and moving base_fee and nv into the NetworkContext), and we're good.

fvm/src/call_manager/default.rs
```diff
@@ -71,8 +71,14 @@ where
...
     type Machine = M;
 
-    fn new(machine: M, gas_limit: i64, origin: (ActorID, Address), nonce: u64) -> Self {
-        let mut gas_tracker = GasTracker::new(Gas::new(gas_limit), Gas::zero());
+    fn new(
+        machine: M,
+        gas_limit: i64,
+        origin: (ActorID, Address),
+        nonce: u64,
+        gas_premium: TokenAmount,
```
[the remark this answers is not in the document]
vyzo: yeah, make sense.

fvm/src/kernel/default.rs
```diff
@@ -441,6 +449,33 @@ where
...
+
+        let tipsets = &self.call_manager.context().network_context.tipsets;
+        if (epoch as usize) < tipsets.len() {
+            return Ok(Some(tipsets[epoch as usize]));
+        }
+
+        Ok(None)
+    }
+}
```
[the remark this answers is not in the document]
vyzo: yeah, i guess i should!

fvm/src/machine/mod.rs
```diff
@@ -174,6 +194,18 @@ impl NetworkConfig {
...
+    /// The network epoch at which the Machine runs.
+    pub epoch: ChainEpoch,
+
+    /// The UNIX timestamp (in seconds) of the current tipset
+    pub timestamp: u64,
+
+    /// The tipset CIDs for the last finality
+    pub tipsets: Vec<Cid>,
+}
+
```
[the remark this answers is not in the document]
vyzo: sure, why not.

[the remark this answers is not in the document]
vyzo: I moved the base fee, but the nv is rather awkward to move -- it is in the NetworkConfig struct and I think it might be best to leave it there for now.

fvm/src/kernel/mod.rs
```diff
@@ -345,3 +356,12 @@ pub trait DebugOps {
...
 }
+
+/// Environment ops
+pub trait ChainOps {
+    /// current tipset timestamp
+    fn tipset_timestamp(&self) -> u64;
+
+    /// epoch tipset cid
+    fn tipset_cid(&self, epoch: i64) -> Result<Option<Cid>>;
+}
```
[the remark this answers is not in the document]
vyzo: done.

fvm/src/machine/mod.rs
```diff
@@ -220,11 +224,6 @@ pub struct MachineContext {
     /// The initial state root on which this block is based.
     pub initial_state_root: Cid,
 
-    /// The base fee that's in effect when the Machine runs.
-    ///
-    /// Default: 0.
-    pub base_fee: TokenAmount,
-
```
raulk, line 195: There's more refactor/consolidation we'll need to do here between `NetworkConfig` and `NetworkContext`.

`NetworkConfig` is confusing because it actually contains **machine** configuration:

- max call depth
- max wasm stack
- actor debugging
- actor redirects
- builtin overrides

As well as some network attributes which should live in `NetworkContext`:

- network version
- price list

But we don't need to do it _now_.

raulk (approved):

PR #910, opened 2022-09-28
feat: auto-load wasm modules from blockstore

Stebalien:
In general, wasm modules should already be cached. However, pre-loading during tests is expensive, so it's nice to allow lazy loading.

This is likely still useful in M2.2. Even though the module _should_ be pre-compiled, the module cache will still be a _cache_.

Also note: A "baseline" compiler is coming to wasmtime (https://github.com/bytecodealliance/wasmtime/pull/4907) that will make lazy-loading a more viable option.

replaces #907 #714

### DIFF
--- a/fvm/src/kernel/default.rs
+++ b/fvm/src/kernel/default.rs
@@ -93,6 +93,10 @@ where
             value_received,
         }
     }
+
+    fn machine(&self) -> &<Self::CallManager as CallManager>::Machine {
+        self.call_manager.machine()
+    }
 }
 
 impl<C> DefaultKernel<C>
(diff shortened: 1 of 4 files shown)


Stebalien:
@rllola @hunjixin does this work for you? It should lazy-load the appropriate actor code on-demand.

Stebalien:
Hm. Well, the integration tests are still really slow....

hunjixin:
this should resolve my issue. but need test.

Stebalien:
Ok, the slowness is that the cache wasn't found for some reason (assume it just got too large, but we'll see if the problem persists)

PR #921, opened 2022-10-04
chore: update rust & fix clippy lints

Stebalien:
Otherwise, anyhow fails to compile. If we weren't compiling on nightly, this wouldn't be an issue. Unfortunately.... we're using substrate-wasm-bundler.


ipld/hamt/src/node.rs
```diff
@@ -184,14 +183,13 @@ where
...
         let hash = H::hash(q);
-        self.get_value(&mut HashBits::new(&hash), bit_width, 0, q, store)
+        self.get_value(&mut HashBits::new(&hash), bit_width, q, store)
     }
 
     fn get_value<Q: ?Sized, S: Blockstore>(
         &self,
         hashed_key: &mut HashBits,
         bit_width: u32,
-        depth: u64,
```
Stebalien, line 194: This was probably an artifact from a previous version. We avoid infinite recursion by pulling bits off the hashed key, which has a limited size anyways.

Stebalien:
We can also punt if this heavily conflicts.

PR #922, opened 2022-10-04
feat: implement max limit and checksum from f4 fip

Stebalien:
- limit: 54 byte subaddress.
- checksum: 4 byte blake2b

### DIFF
--- a/shared/src/address/mod.rs
+++ b/shared/src/address/mod.rs
@@ -38,7 +38,7 @@ pub const SECP_PUB_LEN: usize = 65;
 pub const BLS_PUB_LEN: usize = 48;
 
 /// Max length of f4 sub addresses.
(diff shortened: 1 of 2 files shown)


shared/src/address/mod.rs
```diff
@@ -266,6 +276,29 @@ impl FromStr for Address {
             }
         };
 
+        fn validate_and_split_checksum<'a>(
+            protocol: Protocol,
+            prefix: Option<&[u8]>,
+            payload: &'a [u8],
+        ) -> Result<&'a [u8], Error> {
+            if payload.len() < CHECKSUM_HASH_LEN {
```
mriise, line 284: <= ?

Stebalien: Technically the payload can be empty.

Stebalien: Or, at least, that's not what we're asserting here.

mriise (approved):

PR #923, opened 2022-10-05
feat: Add set_default_network to allow dynamically tagging Network in Address at runtime

hanabi1224:
This PR tries to fix the below TODOs by introducing `fn set_default_network`.  Please let me know your feedback and suggestions, thanks!

```rust
// TODO pull network from config (probably)
// TODO: can we do this using build flags?
pub const NETWORK_DEFAULT: Network = Network::Mainnet;
```

Downstream issues blocked by this:
https://github.com/ChainSafe/forest/issues/2001
https://github.com/ChainSafe/forest/issues/2000


Stebalien:
Could you give a short summary of the design along with some rational? E.g., why not make the `NETWORK_DEFAULT` an atomic?

Stebalien:
We could also add a compile-time flag, or even use a compile-time environment variable.

hanabi1224:
> Could you give a short summary of the design along with some rational? E.g., why not make the NETWORK_DEFAULT an atomic?

Hi @Stebalien yes, I can make NETWORK_DEFAULT an atomic, but I think it's still better to expose mod level wrapper API to hide impl details (using atomic, RwLock or sth else).  A wrapper `AddressContext` is not necessary when it only contains one field but it's more extensible if more similar flags are added.   Let me know your preference. I can remove the current `AddressContext` wrapper and refactor the struct API into mod API

Stebalien:
> yes, I can make NETWORK_DEFAULT an atomic, but I think it's still better to expose mod level wrapper API to hide impl details (using atomic, RwLock or sth else) for both crate or crate users

I agree, but the current API seems over designed. Really, we just need a way to:

- Set the current network.
- Get the current network.

That just requires exporting two functions:

- `default_network()`
- `set_default_network()`

(possibly scoped in a submodule)

Having an actual `AddressContext` type makes it look like there's some address context that can be passed around at runtime, which isn't the case.

>  A wrapper AddressContext is not necessary when it only contains one field but it's more extensible if more similar flags are added.

We don't actually _have_ any other address related thing that might be toggled, so this likely isn't worth optimizing for.

hanabi1224:
@Stebalien fixed as suggested, plz take another look.

shared/src/address/network.rs
```diff
@@ -1,12 +1,18 @@
 // Copyright 2019-2022 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
 
+use num_enum::{FromPrimitive, IntoPrimitive};
```
Stebalien, line 4: Please use `num_derive`, not `num_enum`.

shared/src/address/default.rs
```diff
@@ -0,0 +1,43 @@
+// Copyright 2019-2022 ChainSafe Systems
```
Stebalien, line 1: Let's just put this in the `network.rs` module. An `address::default` module doesn't really make sense.

Stebalien (commented):
LGTM modulo a few nits. Mainly, let's avoid adding new dependencies that we don't really need.

shared/src/address/network.rs
hanabi1224: fixed

shared/src/address/default.rs
hanabi1224: fixed

Stebalien:
Hm. It looks like there are still some compile failures. Could you take a quick look? (`cargo clippy --all --all-targets` should do it).

hanabi1224:
> Hm. It looks like there are still some compile failures. Could you take a quick look? (cargo clippy --all --all-targets should do it).

Fixed!

Issue #924, opened 2022-10-05
Add support for legacy amt v0


creativcoder:
Context: https://github.com/ChainSafe/forest/issues/1624

PR #925, opened 2022-10-05
Add support for legacy amt v0

creativcoder:
This PR adds support for serializing and deserializing amt v0 data structures, which are still used in filecoin for reading headers etc.

Fixes: #924

### DIFF
--- a/ipld/amt/src/amt.rs
+++ b/ipld/amt/src/amt.rs
@@ -12,7 +12,8 @@ use itertools::sorted;
 
 use super::ValueMut;
 use crate::node::{CollapsedNode, Link};
-use crate::root::{RootImpl, Version as AmtVersion, V0, V3};
+use crate::root::version::{Version as AmtVersion, V0, V3};
+use crate::root::RootImpl;
 use crate::{
     init_sized_vec, nodes_for_height, Error, Node, DEFAULT_BIT_WIDTH, MAX_HEIGHT, MAX_INDEX,
 };
@@ -65,10 +66,10 @@ where
         Self::new_with_bit_width(block_store, DEFAULT_BIT_WIDTH)
     }
 
-    /// Construct new Amt with given bit width (v3)
+    /// Construct new Amt with given bit width
     pub fn new_with_bit_width(block_store: BS, bit_width: u32) -> Self {
         Self {
-            root: RootImpl::<V, Ver>::new_with_bit_width(bit_width),
+            root: RootImpl::new_with_bit_width(bit_width),
             block_store,
         }
     }
--- a/ipld/amt/src/amt.rs
+++ b/ipld/amt/src/amt.rs
@@ -19,6 +19,7 @@ use crate::{
 };
 
 #[derive(Debug)]
+#[doc(hidden)]
 pub struct AmtImpl<V, BS, Ver> {
     root: RootImpl<V, Ver>,
     block_store: BS,
@@ -61,7 +62,7 @@ where
     BS: Blockstore,
     Ver: AmtVersion,
 {
-    /// Constructor for Root AMT node (v0)
+    /// Constructor for Root AMT node
     pub fn new(block_store: BS) -> Self {
         Self::new_with_bit_width(block_store, DEFAULT_BIT_WIDTH)
     }
--- a/ipld/amt/src/amt.rs
+++ b/ipld/amt/src/amt.rs
@@ -12,14 +12,14 @@ use itertools::sorted;
 
 use super::ValueMut;
 use crate::node::{CollapsedNode, Link};
-use crate::root::RootImpl;
+use crate::root::{RootImpl, Version as AmtVersion, V0, V3};
 use crate::{
-    init_sized_vec, nodes_for_height, Error, Node, Root, DEFAULT_BIT_WIDTH, MAX_HEIGHT, MAX_INDEX,
+    init_sized_vec, nodes_for_height, Error, Node, DEFAULT_BIT_WIDTH, MAX_HEIGHT, MAX_INDEX,
 };
 
 #[derive(Debug)]
-pub struct AmtImpl<V, BS, const VER: u8> {
-    root: RootImpl<V, VER>,
+pub struct AmtImpl<V, BS, Ver> {
+    root: RootImpl<V, Ver>,
     block_store: BS,
 }
 
@@ -44,30 +44,31 @@ pub struct AmtImpl<V, BS, const VER: u8> {
 /// // Generate cid by calling flush to remove cache
 /// let cid = amt.flush().unwrap();
 /// ```
-pub type Amt<V, BS> = AmtImpl<V, BS, 3>;
+pub type Amt<V, BS> = AmtImpl<V, BS, V3>;
 /// Legacy amt V0
-pub type Amtv0<V, BS> = AmtImpl<V, BS, 0>;
+pub type Amtv0<V, BS> = AmtImpl<V, BS, V0>;
 
-impl<V: PartialEq, BS: Blockstore> PartialEq for Amt<V, BS> {
+impl<V: PartialEq, BS: Blockstore, Ver: PartialEq> PartialEq for AmtImpl<V, BS, Ver> {
     fn eq(&self, other: &Self) -> bool {
         self.root == other.root
     }
 }
 
-impl<V, BS> Amt<V, BS>
+impl<V, BS, Ver> AmtImpl<V, BS, Ver>
 where
     V: DeserializeOwned + Serialize,
     BS: Blockstore,
+    Ver: AmtVersion,
 {
-    /// Constructor for Root AMT node
+    /// Constructor for Root AMT node (v0)
     pub fn new(block_store: BS) -> Self {
         Self::new_with_bit_width(block_store, DEFAULT_BIT_WIDTH)
     }
 
-    /// Construct new Amt with given bit width.
+    /// Construct new Amt with given bit width (v3)
     pub fn new_with_bit_width(block_store: BS, bit_width: u32) -> Self {
         Self {
-            root: Root::new(bit_width),
+            root: RootImpl::<V, Ver>::new_with_bit_width(bit_width),
             block_store,
         }
     }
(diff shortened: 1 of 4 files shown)


creativcoder:
* Using types would not allow us to have the pub type alias with compile time version declaration. 
  With types, we would need to expose a generic top level Amt. 
  With const version, I have type aliased the v3 to just Amt, and the change would only be affecting the v0 code.
* Using const we don't need to store an instance of version within the struct.
* Since version is only going to be an integral (in this particular case) i thought using const generic would be the simplest.
* With types we wouldn't be able to match it without an instance in serde methods.

Do we foresee any issues long term with this approach?

Stebalien:
> Using types would not allow us to have the pub type alias with compile time version declaration.
With types, we would need to expose a generic top level Amt.
With const version, I have type aliased the v3 to just Amt, and the change would only be affecting the v0 code.


Ah, I mean replace the consts with unit structs (possibly implementing some trait. I.e.:


```rust
pub struct V0;
pub struct V3;

pub trait AmtVersion {} // ideally private..., but I'm pretty sure rust won't let us do that?

impl AmtVersion for V0 {}
impl AmtVersion for V3 {}

pub struct AmtImpl<V, BS, VER: AmtVersion> { ... }
```

It's _basically_ identical to the const version. The main differences are:

- We can use associated types in the type signature. I.e. `AmtImpl<Thing, MyBS, OtherType::Version>`. IIRC, this isn't stable for consts.
- We can restrict the allowed versions. Right now, someone can technically specify a version that makes no sense and we don't really check it anywhere.
- Given different future "features", we can add traits for these "features" and implement these traits on the versions that support them. Then we can say things like `AmtImpl<Thing, MyBS, V: HasSomeFeature>`.
- We can implement amt-version specific logic on the _version_ type itself. E.g., add a serialize/deserialize method.

On the other hand, the main benefit of the const is that it's easier to write "runtime" checks that check the const, that actually disappear at compile time (basically, it gives you specialization without specialization).

Basically, I just want to make sure you've seen and considered both options before proceeding. There are tradeoffs both ways and I haven't fully considered either option.

creativcoder:
Yeah, you're right about being able to use any integer value for the version. Using traits with unit structs, is definitely extensible. I'm gonna try that, and update soon.

creativcoder:
@mriise :eyes:  This is amazing! thanks for sharing.

ipld/amt/src/amt.rs
```diff
@@ -37,32 +44,31 @@ use crate::{
...
-pub struct Amt<V, BS> {
-    root: Root<V>,
-    block_store: BS,
-}
+pub type Amt<V, BS> = AmtImpl<V, BS, V3>;
+/// Legacy amt V0
+pub type Amtv0<V, BS> = AmtImpl<V, BS, V0>;
 
-impl<V: PartialEq, BS: Blockstore> PartialEq for Amt<V, BS> {
+impl<V: PartialEq, BS: Blockstore, Ver: PartialEq> PartialEq for AmtImpl<V, BS, Ver> {
```
Stebalien, line 51: We shouldn't _need_ PartialEq for `Ver`, right?

ipld/amt/src/amt.rs
```diff
@@ -37,32 +44,31 @@ use crate::{
...
     pub fn new(block_store: BS) -> Self {
         Self::new_with_bit_width(block_store, DEFAULT_BIT_WIDTH)
     }
 
-    /// Construct new Amt with given bit width.
+    /// Construct new Amt with given bit width (v3)
     pub fn new_with_bit_width(block_store: BS, bit_width: u32) -> Self {
         Self {
-            root: Root::new(bit_width),
+            root: RootImpl::<V, Ver>::new_with_bit_width(bit_width),
```
Stebalien, line 71: Is this type not inferred?

ipld/amt/src/root.rs
```diff
@@ -1,62 +1,126 @@
...
+    const NUMBER: usize;
+}
+
+impl Version for V0 {
+    const NUMBER: usize = 0;
+}
+
+impl Version for V3 {
+    const NUMBER: usize = 3;
+}
```
Stebalien, line 27: I'd put these in a `version` submodule.

ipld/amt/src/amt.rs
```diff
@@ -37,32 +44,31 @@ use crate::{
...
+    Ver: AmtVersion,
 {
-    /// Constructor for Root AMT node
+    /// Constructor for Root AMT node (v0)
     pub fn new(block_store: BS) -> Self {
         Self::new_with_bit_width(block_store, DEFAULT_BIT_WIDTH)
     }
 
-    /// Construct new Amt with given bit width.
+    /// Construct new Amt with given bit width (v3)
```
Stebalien, line 68: These constructors aren't related to the version.

Stebalien (commented):
Mostly nits at this point.

ipld/amt/src/root.rs
```diff
@@ -1,62 +1,126 @@
...
+use crate::{init_sized_vec, Node, DEFAULT_BIT_WIDTH};
+
+#[derive(PartialEq, Debug)]
+pub struct V0;
+#[derive(PartialEq, Debug)]
+pub struct V3;
+
+pub trait Version {
+    const NUMBER: usize;
+}
```
mriise, line 19: I would be careful to fully expose this trait, as now users could implement their own versions and get trapped on unreachable if they ever use it. I would either static assert the version, or make it pub(crate) or something.

ipld/amt/src/amt.rs
creativcoder: It shouldn't, I too found it weird but here's the error:

```
error[E0369]: binary operation `==` cannot be applied to type `RootImpl<V, Ver>`
  --> ipld/amt/src/amt.rs:53:19
   |
53 |         self.root == other.root
   |         --------- ^^ ---------- RootImpl<V, Ver>
   |         |
   |         RootImpl<V, Ver>
   |
help: consider restricting type parameter `Ver`
   |
51 | impl<V: PartialEq, BS: Blockstore, Ver: std::cmp::PartialEq> PartialEq for AmtImpl<V, BS, Ver> {
```

ipld/amt/src/amt.rs
creativcoder: Ah it does, removed turbofishes.

ipld/amt/src/root.rs
creativcoder: added

ipld/amt/src/amt.rs
creativcoder: removed

ipld/amt/src/root.rs
creativcoder: I have moved it behind a version submodule and pub(crate)'d it.

creativcoder:
btw do we require a minor version bump for this?

Stebalien:
> btw do we require a minor version bump for this?

It looks completely backwards compatible, so no.

ipld/amt/src/amt.rs
Stebalien: Hm. Ok, so, I think that's just a derive issue (you could manually implement it). But honestly, it doesn't really matter.

ipld/amt/src/amt.rs
```diff
@@ -12,10 +12,18 @@ use itertools::sorted;
...
 use crate::node::{CollapsedNode, Link};
+use crate::root::version::{Version as AmtVersion, V0, V3};
+use crate::root::RootImpl;
 use crate::{
-    init_sized_vec, nodes_for_height, Error, Node, Root, DEFAULT_BIT_WIDTH, MAX_HEIGHT, MAX_INDEX,
+    init_sized_vec, nodes_for_height, Error, Node, DEFAULT_BIT_WIDTH, MAX_HEIGHT, MAX_INDEX,
 };
 
+#[derive(Debug)]
+pub struct AmtImpl<V, BS, Ver> {
```
Stebalien, line 22: Can we add:

```rust
#[doc(hidden)]
```

ipld/amt/src/amt.rs
```diff
@@ -37,32 +45,31 @@ use crate::{
...
 
-impl<V, BS> Amt<V, BS>
+impl<V, BS, Ver> AmtImpl<V, BS, Ver>
 where
     V: DeserializeOwned + Serialize,
     BS: Blockstore,
+    Ver: AmtVersion,
 {
-    /// Constructor for Root AMT node
+    /// Constructor for Root AMT node (v0)
```
Stebalien, line 64: Not v0 specific.

ipld/amt/src/root.rs
```diff
@@ -1,62 +1,128 @@
...
             node: Node::Leaf {
                 vals: init_sized_vec(bit_width),
             },
+            ver: PhantomData,
+        }
+    }
+}
+
+impl<V> RootImpl<V, self::version::V0> {
+    pub(crate) fn new() -> Rootv0<V> {
```
Stebalien, line 60: Let's make the visibility of `new` and `new_with_bit_width` match (just pick the same one, either one).

Stebalien (commented):
Some super-nits but this is otherwise ready to merge.

Stebalien:
Hm. Except that there are some clippy lints and warnings.

ipld/amt/src/root.rs
creativcoder: Sure, using `pub(crate)`

ipld/amt/src/amt.rs
creativcoder: Added

ipld/amt/src/amt.rs
creativcoder: Fixed

creativcoder:
I'm not sure why there's still a warning on type alias of RootImpl as I did use them both in [root.rs](https://github.com/creativcoder/ref-fvm/blob/913751785914eac44db08a17174b00421872d696/ipld/amt/src/root.rs#L158)

Stebalien:
Ah, well, tests don't count. I'd just define these aliases inside the tests.

creativcoder:
Ah I overlooked the point that Root isn't meant to be part of public API. Moved the type aliases.

Stebalien:
Thanks!

PR #928, opened 2022-10-06
feat: simplify syscall returns

Stebalien:
Some history:

- At first, we returned an Option because these syscalls would return a special sentinel value for "not found".
- Then we changed the syscalls to return the NotFound error code, but we never updated the kernel methods.

Now, I'm adding another kernel method. But in my case, to keep this pattern, I'd need to return `Option<Option<Address>>`... so I'm proposing that we change these other kernel functions instead.

### DIFF
--- a/fvm/src/kernel/default.rs
+++ b/fvm/src/kernel/default.rs
@@ -223,10 +223,7 @@ where
             // exists; if missing, it fails the self destruct.
             //
             // In FVM we check unconditionally, since we only support nv13+.
-            let beneficiary_id = self
-                .resolve_address(beneficiary)?
-                .context("beneficiary doesn't exist")
-                .or_error(ErrorNumber::NotFound)?;
+            let beneficiary_id = self.resolve_address(beneficiary)?;
 
             if beneficiary_id == self.actor_id {
                 return Err(syscall_error!(Forbidden, "benefactor cannot be beneficiary").into());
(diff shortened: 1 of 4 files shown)


Stebalien:
This is in-line with how all the other kernel functions behave. If the kernel returns an error, we propegate it. Otherwise, we convert the value.

fvm/src/kernel/default.rs
```diff
@@ -745,18 +742,23 @@ impl<C> ActorOps for DefaultKernel<C>
 where
     C: CallManager,
 {
-    fn resolve_address(&self, address: &Address) -> Result<Option<ActorID>> {
-        self.call_manager.state_tree().lookup_id(address)
+    fn resolve_address(&self, address: &Address) -> Result<ActorID> {
```
vyzo, line 745: This is going to cause some problems, as actors typically treat kernel errors as fatal and dont look into them.

Also, semantically, is NotFound even a thing? Thats what Options are for.

vyzo (commented):
I am not convinces this is the right thing to do.

Why is NotFound even a thing?
Isnt this what Options are for?
Absence of value -> automaitcally Option.

Also, this complicates actor code as we now have to start looking into errors instead of blanket aborting in them.
Ergonomically, this is objectively worse.

Either way, i am just raising my concerns, if you think its the right thing todo, i wont block you.
But imo it should NotFound that goes from the error codes.

fvm/src/kernel/default.rs
Stebalien: We check for _specific_ error codes in specific syscalls (when expected).

Stebalien:
I'm changing the kernel, not the syscalls. The syscalls will behave _identically_ after this change. I'm just making the kernel interface better match the syscalls in implementation.

vyzo:
well, you have bigger claws :)
Just saying though, what good is an Option?

Stebalien:
> well, you have bigger claws :)

I'm asking you for your opinion because I value it. If I can't convince you that this patch is an improvement, then there's a pretty good chance it isn't.

---

Here's some context: https://github.com/filecoin-project/ref-fvm/issues/738#issuecomment-1101423891

In pure rust, I'd return an option. However, the syscall boundry is a bit more limited. We changed this because:

- Many of our APIs were subtly inconsistent.
- These cases _usually_ indicate a failure of some form.
- Performance & simplicity. We already _had_ a way to indicate typed failures through error numbers. Wrapping a fallible type within a fellable type wasn't great.
    - Note: _rust_ is often cleaver about this and can optimize things like `Result<Option<Foo>>`, but that's not something we could do.

In the case of this _particular_ patch, I'm just moving logic from the syscall handlers into the kernel itself (that's why it actually removes lines of code and doesn't affect any tests).

vyzo (approved):
ok, i am mostly convinced -- the part that does it is not having to change actors.

Stebalien:
Yeah, I'm basically treating this as a "kernel" like interface, standard unix syscall warts and limitations included.

PR #929, opened 2022-10-06
fix: update set_default_network to use store api instead

hanabi1224:
atomic `store` API should be better than `fetch_update` here.


Stebalien (approved):
Yeah, good catch.

Issue #892, opened 2022-09-19
EVM Storage: Keep the root node small


aakoshh:
Part of https://github.com/filecoin-project/ref-fvm/issues/859
Depends on https://github.com/filecoin-project/ref-fvm/issues/888

The HAMT consists of a root `Node`, which contains a sparse vector of `Pointer`s. Each `Pointer` can be a `Link` to another `Node` or `Values` containing a bucket of up to 3 `KeyValuePair`s. With the default value of 8 bit width resulting in 256 possible places in the vector, and the EVM storing 256bit keys and 256bit values, this can result in a maximum `Node` size of `256*3*(256+256)` which is around 40kB. 

Since EVM hashes keys in mappings, these will be distributed across the key space evenly, so it's reasonable to think they might saturate the root node to a high level before the `Value`s are replaced by `Link`s. This will weigh down on the contract because the root node is always retrieved, even we don't want to access any of the entries in the mapping. 

We might want to improve upon this by having an option in the HAMT to keep the root node (depth=0) free of `Values` and only allow `Link` from the beginning. 

That costs an extra lookup for each variable in the top level contract as well, so it might be further optimised by looking for the long 0x00000000 keys which map to the low slots in the contract. Maybe the first bucket in the HAMT should be allowed to have `Values`, although most likely it won't make a difference because if there are more than 3 variables they are sunk down anyway.

Issue #891, opened 2022-09-19
EVM Storage: Skip/Extend empty levels in the HAMT


aakoshh:
Part of https://github.com/filecoin-project/ref-fvm/issues/859
Depends on https://github.com/filecoin-project/ref-fvm/issues/890

The HAMT has buckets of size 3 for KV pairs; once there are more items under the same key-prefix, it sinks them into sub-nodes. 

After disabling key-hashing, all items in a dynamic array will fall under the same key prefix. Keys are 256 bits long, and with the default 8 bit length (nibble size) that results in a maximum of 32 levels in the HAMT. Because the dynamic array puts keys next to each other, we will actually have 32 levels right after the 4th item is appended to the array. 

Following 32 almost empty nodes to get to the array, and then writing back all 32 new nodes once an item is appended is wasteful. We can take insiration from the Merkle Patricia Trie which contains extension nodes to skip ahead empty levels, right to the next non-empty blocks. Optimised Sparse Merkle Trees like Jellyfish also substitute a sub-tree with a node, if that node is the only non-empty descendant. That will not be the case for us, because as the array grows, it will spill over to the next node.

maciejwitowski:
Closing as part of https://github.com/filecoin-project/ref-fvm/issues/859

PR #930, opened 2022-10-06
EVM-891: HAMT link extension

aakoshh:
part of https://github.com/filecoin-project/ref-fvm/issues/891

## Description

The PR adds an option to the HAMT to use _extensions_ in `Pointer::Link` to skip levels in the tree where the keys would map into the same bucket. This is not enabled by default because with the default `Sha256` hashing this is probably not an issue, and also for backwards compatibility on the CID level. It's being added to support the HAMT as a backend for the EVM, where the Solidity compiler already applies hashing on the keys plus an offset to put array entries next to each other. 

As a first step we circumvented the HAMT hashing in the EVM actor in https://github.com/filecoin-project/builtin-actors/pull/730
As the charts showed this resulted in very deep trees for arrays, where most levels apart from the bottom only contain a single pointer. This PR should see those trees shrink, the number of reads and writes go down. 

Since then we discussed that potentially the AMT would also be a viable data structure for this purpose, if it had similar skipping to what's being added here. If the EVM switched to AMT, I'm not sure how much value is in the HAMT having the skip feature. It's relevant based on the hashing strategy that the user implements.

## Changes
- [x] Add a `hamt::Config` to contain `bit_width` and a new `use_extensions` flag. Disabled by default.
- [x] If extensions are enabled, find the longest prefix in keys once a bucket is filled up and a subshard needs to be created.
- [x] During searches/deletes, check if a key matches the extension in the link. If not, don't retrieve the node from the store.
- [x] Use `Ipld::Map` for serialization of `Pointer::Link` _iff_ there is a non-empty extension in it.
- [x] Refactored unit tests to allow different HAMT configurations
- [x] Added fuzz tests for the extensions using the same corpora as the simple case 

## Notes

I saw that the `depth` was recently removed from the recursive methods, however I wanted to keep it until I can try another optimisation in https://github.com/filecoin-project/ref-fvm/issues/892

In retrospect it would have been much simpler to store the list of indexes in the `Extension` at each level, which would have made splitting easier. I decided early on to use bit packing instead based on how `HashBits` works to save storage space; if the `bit_width` is much smaller than 8 this would result in fewer number of bytes in the `path` than if we stored numbers. At the time I thought about storing indexes as `u32` which `next` returns, but that would have made no sense actually since it always fits an `u8`, so the actual difference might not be that much; definitely not with the default `bit_width` where it's identical.

### DIFF
--- a/testing/fuzz-corpora.md
+++ b/testing/fuzz-corpora.md
@@ -1,2 +1,8 @@
 Symlink `./fuzz-corpora` to https://github.com/filecoin-project/ref-fvm-fuzz-corpora repo
 to utilize existing corpus during fuzzing.
+
+For example, assuming `ref-fvm-fuzz-corpora` is next to `ref-fvm`, run this in `testing`:
+
+```shell
+ln -s ../../ref-fvm-fuzz-corpora/ fuzz-corpora
+```


Stebalien:
Well, I think this was definitely a success. The buckets in this "hamt" based version better handle "random" keys than the AMT would, so you were probably right to start here.

I'm wondering if we should just fork this HAMT into a new "ordered map" datastructure with an optional hash function (feeling free to break things). That seems like a reasonable basis for both an AMT and a HAMT, and we'd be able to avoid storing values in intermediate nodes (as you've pointed out).

aakoshh:
I can see how you would rather have a simpler HAMT codebase and let it stick to its basic use case, without burdening it with the extension support that normally doesn't come up because hashes don't overlap too long, and the HAMT stores data in the nodes at the shortest path it can (at the cost of read/write amplification). 

On the other hand maintaining yet another tree is burdensome as well, unless we can say exactly what it needs to do. I feel we're getting close with the HAMT for the EVM, we can configure it so that it behaves like a B-tree with skipping.

Let me do the final task https://github.com/filecoin-project/ref-fvm/issues/892 so we can see the final measurements.

aakoshh:
Okay so here's an idea for a data structure or HAMT variant that could work well:
* Data only lives in the leaves, but we use extensions/skipping to go straight to them or the next node with branches. We don't pack key-value pairs in intermediate nodes that clog up the tree. 
* A leaf is still wide and can store 2**bit_width elements, which is good for arrays. 
* But because mapping keys are dispersed, and they are put into individual nodes because they only have a common prefix, but a unique suffix to their key, they live alone, and during retrieval we don't read any other unrelated KV pairs in.

aakoshh:
I did an experiment with the above suggestion about not putting data into intermediary nodes in https://github.com/filecoin-project/ref-fvm/pull/946 and https://github.com/filecoin-project/builtin-actors/pull/745

From the numbers it looks like this doesn't make sense for the EVM, but it might do for other HAMT users with larger values in the tree.

ipld/hamt/fuzz/corpus/extensions
```diff
@@ -0,0 +1 @@
+../../../../testing/fuzz-corpora/corpus/hamt_simple
```
Kubuxu, line 1: I would suggest a separate corpus directory. The corpus optimization and compaction is done based on coverage information, I think the two fuzzing targets will remove corpus elements which are not exercising code path useful to them.

Kubuxu:
Are we sure that the link extension doesn't break the independence of HAMT on the order of operations? If it does we might want to consider the consequences.

aakoshh:
> Are we sure that the link extension doesn't break the independence of HAMT on the order of operations?

If you mean the that the CIDs will be the same, you are right that I did not test for this. On the deletion side the the `clean` function that should take care of it, but the existing tests use hardcoded CIDs, and I didn't capture new ones. I can at least add some assertions there that should compare before/after CIDs not against the hardcoded ones but as invariants. 

Also a good question whether the order of inserts lead to the same CID. I could add some `quickcheck` tests for that, similar to the fuzz tests but under the unit tests.

ipld/hamt/fuzz/corpus/extensions
Kubuxu: I would suggest creating a new directory in the submodule with a copy of the hamt_simple corpus, you can push straight to master there IMO (CI automation does that when it finds new valuable test cases).

> Every time I point cargo at a new ref-fvm branch on Github from the builtin-actors repo it takes ages to clone another copy of the fuzz targets submodule

Do you mean cloning the fuzz-corpora takes a long time? If so we could probably discard old commits there.

Looking at the CI, the corpus updating broke at some point, I'm not sure why. I will open an issue.


> Isn't there a way to just fuzz on random input for a limited time and only print failures on CI?

There isn't much point, fuzzing takes a while and works best with good corpus. One of the CID bugs was found after weeks of fuzzing on the CI.

aakoshh: > Do you mean cloning the fuzz-corpora takes a long time?

The fuzz corpora and some other test vectors. For example the following shows how long it takes to initialize these two for a branch of `builtin-actors` which is built against a specific branch of `ref-fvm`:

```console
❯ cargo cache -a                                                                                                                                                                                                 ─╯
Clearing cache...

... just to make sure I don't have it ...
❯ make all                                                                                                                                                                                                     ─╯
cargo test --test storage_footprint
    Updating git submodule `https://github.com/filecoin-project/fvm-test-vectors`
    Updating git submodule `https://github.com/filecoin-project/ref-fvm-fuzz-corpora.git`
   Compiling fvm_ipld_blockstore v0.1.1 (https://github.com/filecoin-project/ref-fvm?branch=evm-891-hamt-link-extension#7fe44ee3)
   
^Cmake: *** [Makefile:31: measure-storage-footprint] Interruptuntime                                                                                                                                             

    ~/projects/consensuslab/builtin-actors/actors/evm   evm-891-hamt-link-extens… ················································································ ✘ INT   2m 32s   1.63.0  16:49:09 ─╮
❯ cargo cache -a                                                                                                                                                                                                 ─╯
Clearing cache...

Cargo cache '/home/aakoshh/.cargo':

Total:                                      1.97 GB => 1.40 GB
  19 installed binaries:                             469.61 MB
  Registry:                             784.14 MB => 710.38 MB
    Registry index:                                  358.90 MB
    2757 crate archives:                             351.47 MB
    272 => 0 crate source checkouts:          73.77 MB => 0  B
  Git db:                               713.76 MB => 224.85 MB
    32 bare git repos:                               224.85 MB
    2 => 0 git repo checkouts:               488.91 MB => 0  B

Size changed 1.97 GB => 1.40 GB (-562.67 MB, -28.59%)

    ~/projects/consensuslab/builtin-actors/actors/evm   evm-891-hamt-link-extens… ···································································································  1.63.0  16:49:45 ─╮
❯                                                                                                                                                                                                                ─╯
```

So it took 2.5 minutes to clone those two, and after that clearing it away shows it was 560MB that had to be downloaded.
By my count it took about 1 minute to do the fuzz corpora, although looking at the data on disk, it doesn't seem to be that big actually. 

Anyway, happy to do the copy, no problem, thanks for explaining how it works!

Kubuxu:
Another randomized test we had in go world was: perform all these random operations on HAMT (insert/update, delete), take the CID of resulting HAMT, separately coalese operations (last insert or delete survives) and apply this simplified sequence on hamt, compare the CID of HAMTs.
I'm not saying we have to add it, I think the quickcheck convinces me.

ipld/hamt/fuzz/corpus/extensions
Kubuxu: > Interesting, I need to learn more about how this fuzzing library works.

It is less about the lib itself (as it is a wrapper/entry point for libfuzzer) but tooling around it. We use ClusterFuzzLite for fuzzing on the CI. libfuzzer is able to do coverage guided fuzzing, the corpus minimization is also feature from it but you have call it separately which ClusterFuzzLite does.

aakoshh:
> Another randomized test we had in go world was

Thanks @Kubuxu, sounds like a good test that captures everything :+1: Added and removed the previous.

ipld/hamt/fuzz/corpus/extensions
aakoshh: Thanks! 
As a heads up I also added a fuzz target in https://github.com/filecoin-project/ref-fvm/pull/946

Kubuxu: Heads up on possible bug in CI fuzzing: https://github.com/filecoin-project/ref-fvm/issues/959

aakoshh: Can't see it yet, but I'll update that PR too, thanks.

aakoshh:
The verdict was that we will not add extension support to the current HAMT, but rather make a new data structure that will have it always on. But we do want to keep the next PR which keep the root free of data, and the other refactoring in this PR which deals with the tests. 

I will merge this PR into a feature branch and then restore the original HAMT.

PR #931, opened 2022-10-06
fix: make address networks global

Stebalien:
reverts the revert of #923 and #929, then replaces most of the logic (we'll squash before merging)

We relied on `new_id` being "const" in several places, so the previous patches weren't going to work. I had a few options:

1. Revert.

2. Make it a compile time option. This is strictly less flexible than the current option.

3. Use a third value for "default" or "unset" that simply follows the system default. This was probably a better solution than the existing one, but made the Eq implementation a bit tricky.

So, instead of all of these solutions, I took a closer look at the issue:

1. Any solution that bundles "networks" with addresses will have a bit of a race: any address constructed _before_ the network is set will be wrong.

2. Addresses don't really _have_ networks. _Textual_ addresses do, but addresses lose their networks as soon as they're parsed.

So I opted for a completely new option (that, honestly, more closely matches lotus's behavior):

1. Remove networks from addresses.
2. Enforce that addresses have the correct network _on parse_.
3. Use the "current" network _on format_.

In this world, implementations should set the correct network before trying to parse any addresses. However, if they get this wrong, they'll get an actual _error_ instead of getting addresses that _should_ be equivalent but aren't.

---

Breaking:

- Removes `network` from the `Address` type. I can't actually find any calls in forest or the builtin actors outside of tests.
- Removes `set_network` from the `Address` type. Again, this wasn't called outside of tests.
- Enforces that all textual addresses match the correct network on parse. This is the change most likely to cause issues, if any.

Alternatives to this last point would be to accept any valid "network" and ignore it (what lotus does). But I'd really rather not as this could be quite useful for catching bugs. Anyways, this is more likely to catch _bugs_ from addresses that hard-code a specific network than anything else.

### DIFF
--- a/fvm/src/machine/mod.rs
+++ b/fvm/src/machine/mod.rs
@@ -27,10 +27,10 @@ pub use engine::{Engine, EngineConfig, MultiEngine};
 
 mod boxed;
 
-pub const REWARD_ACTOR_ADDR: Address = Address::new_id(2);
+pub const REWARD_ACTOR_ADDR_ID: u64 = 2;
 
 /// Distinguished Account actor that is the destination of all burnt funds.
-pub const BURNT_FUNDS_ACTOR_ADDR: Address = Address::new_id(99);
+pub const BURNT_FUNDS_ACTOR_ADDR_ID: u64 = 99;
 
 /// The Machine is the top-level object of the FVM.
 ///
--- a/fvm/src/machine/mod.rs
+++ b/fvm/src/machine/mod.rs
@@ -27,10 +27,10 @@ pub use engine::{Engine, EngineConfig, MultiEngine};
 
 mod boxed;
 
-pub const REWARD_ACTOR_ADDR_ID: u64 = 2;
+pub const REWARD_ACTOR_ADDR: Address = Address::new_id(2);
 
 /// Distinguished Account actor that is the destination of all burnt funds.
-pub const BURNT_FUNDS_ACTOR_ADDR_ID: u64 = 99;
+pub const BURNT_FUNDS_ACTOR_ADDR: Address = Address::new_id(99);
 
 /// The Machine is the top-level object of the FVM.
 ///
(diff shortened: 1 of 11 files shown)


Stebalien:
@lemmih please let me know if this works for you, or if you need another solution. The current solution is effectively broken as addresses parsed from binary will magically get a network based on the current default network. If addresses actually encoded the network in the binary format, I'd say that the constructor should just take the network as a parameter. But as that's _not_ the case, this appears to be more of a "ux" thing.

Note: It's still _possible_ to parse an address from a different network, you just need to explicitly specify this (by parsing with `Network::Testnet.parse_address(...)`. I'm also happy to expose additional functionality if you need it.

Stebalien:
Cc @hanabi1224.

aakoshh:
I wholeheartedly support this change. I had issues with it in Forest API where addresses switch network:

https://filecoinproject.slack.com/archives/C029LPZ5N73/p1658499046980439?thread_ts=1657886906.185139&cid=C029LPZ5N73

fvm/src/machine/mod.rs
```diff
@@ -27,10 +27,10 @@ pub use engine::{Engine, EngineConfig, MultiEngine};
 
 mod boxed;
 
-pub const REWARD_ACTOR_ADDR: Address = Address::new_id(2);
+pub const REWARD_ACTOR_ADDR_ID: u64 = 2;
 
 /// Distinguished Account actor that is the destination of all burnt funds.
-pub const BURNT_FUNDS_ACTOR_ADDR: Address = Address::new_id(99);
+pub const BURNT_FUNDS_ACTOR_ADDR_ID: u64 = 99;
```
lemmih, line 33: We can go back to defining `BURNT_FUNDS_ACTOR_ADDR`, right?

lemmih:
> @lemmih please let me know if this works for you, or if you need another solution. The current solution is effectively broken as addresses parsed from binary will magically get a network based on the current default network. If addresses actually encoded the network in the binary format, I'd say that the constructor should just take the network as a parameter. But as that's _not_ the case, this appears to be more of a "ux" thing.
> 
> Note: It's still _possible_ to parse an address from a different network, you just need to explicitly specify this (by parsing with `Network::Testnet.parse_address(...)`. I'm also happy to expose additional functionality if you need it.

This works for Forest.

lemmih (approved):

PR #934, opened 2022-10-07
fix: cleanup tipset timestamp/cid

### LINKED ISSUE #927: The tipset_cid syscall bikeshed
This syscall currently:

- Takes an offset starting at the current epoch.
- Returns "illegal argument" in almost all error cases.
- Returns "0" (an empty CID) if we try to lookup an epoch before epoch 0.

The bikeshed:

- The _randomness functions_ take an _epoch_, not an offset. Really, this function claims to take an epoch as well but doesn't.
    - We definitely need to add some documentation.
    - Differing from the randomness functions makes _some_ sense as we generally use those functions to "tie" commitments to the chain, but I'm not sure if it's worth the confusion.
- Errors
    - If we're looking back too far, we should use the LimitExceeded error code.
    - If we do take an absolute epoch, we'd return IllegalArgument for "negative" epochs.
    - If we continue to take a relative epoch, we should return IllegalOperation in this case (I think?). This case almost never matters.

@vyzo any strong opinions on any of these cases?

Stebalien:
fixes #927

- Use i64 for timestamps because that's technically more correct.
  - And that's the best correct.
- Use the ChainEpoch type where appropriate.
- Use Result instead of Option for incorrect epochs, and use a descriptive error.
- Take the epoch instead of an offset in tipset_cid.
- Docs, all the docs.
- Fill in some todos
- Add a TODOs/issues for missing things.

### DIFF
--- a/fvm/src/kernel/default.rs
+++ b/fvm/src/kernel/default.rs
@@ -672,24 +672,23 @@ where
         &self.call_manager.context().network_context.base_fee
     }
 
-    fn tipset_timestamp(&self) -> u64 {
+    fn tipset_timestamp(&self) -> i64 {
         self.call_manager.context().network_context.timestamp
     }
 
-    fn tipset_cid(&self, epoch: i64) -> Result<Option<Cid>> {
+    fn tipset_cid(&self, epoch: ChainEpoch) -> Result<Cid> {
         if epoch < 0 {
             return Err(syscall_error!(IllegalArgument; "epoch is negative").into());
         }
-        if epoch >= 900 {
-            return Err(syscall_error!(IllegalArgument; "epoch out of finality range").into());
+        let offset = self.call_manager.context().network_context.epoch - epoch;
+        if offset < 0 {
+            return Err(syscall_error!(IllegalArgument; "epoch is in the future").into());
         }
-
         let tipsets = &self.call_manager.context().network_context.tipsets;
-        if (epoch as usize) < tipsets.len() {
-            return Ok(Some(tipsets[epoch as usize]));
+        if offset >= tipsets.len() as i64 {
+            return Err(syscall_error!(LimitExceeded; "tipset lookback exceeded limit").into());
         }
-
-        Ok(None)
+        Ok(tipsets[offset as usize])
     }
 }
 
--- a/fvm/src/machine/mod.rs
+++ b/fvm/src/machine/mod.rs
@@ -167,6 +167,7 @@ impl NetworkConfig {
             network: self.clone(),
             network_context: NetworkContext {
                 epoch,
+                // TODO #933
                 timestamp: 0,
                 tipsets: vec![],
                 base_fee: TokenAmount::zero(),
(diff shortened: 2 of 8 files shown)


fvm/src/machine/mod.rs
```diff
@@ -199,7 +200,7 @@ pub struct NetworkContext {
     pub epoch: ChainEpoch,
 
     /// The UNIX timestamp (in seconds) of the current tipset
-    pub timestamp: u64,
+    pub timestamp: i64,
```
Stebalien, line 203: Look... maybe we make a time machine? I'm happy to revert, this just made me twitch a bit.

sdk/src/error.rs
```diff
@@ -11,3 +11,11 @@ pub enum ActorDeleteError {
     #[error("deletion beneficiary does not exist")]
     BeneficiaryDoesNotExist,
 }
+
+#[derive(Copy, Clone, Debug, Error)]
+pub enum EpochBoundsError {
```
Stebalien, line 16: (yeah, yeah, here come the rust police)

Basically, this way we can distinguish between "you gave us a bad epoch" and "I can't do that dave".

fvm/src/kernel/default.rs
```diff
@@ -672,24 +672,23 @@ where
...
     }
 
-    fn tipset_timestamp(&self) -> u64 {
+    fn tipset_timestamp(&self) -> i64 {
         self.call_manager.context().network_context.timestamp
     }
 
-    fn tipset_cid(&self, epoch: i64) -> Result<Option<Cid>> {
+    fn tipset_cid(&self, epoch: ChainEpoch) -> Result<Cid> {
         if epoch < 0 {
```
Stebalien, line 680: So... We _could_ interpret this as "relative to the current epoch". Thoughts? Disgust?

Stebalien:
(@vyzo I'm looking for opinions, not just a signoff)

fvm/src/machine/mod.rs
vyzo: dunno, u64 seems fine. But it doesnt matter much either!

sdk/src/error.rs
vyzo: I dont think anyone is going to distinguish that programmatically, the error message is what matters most.

In theory it _could_ be useful, in practice we'll just `?` raise it.

fvm/src/kernel/default.rs
vyzo: uhm, lets just make it absolute and not bother with edge cases.

I am fine with returning Result here instead of an Option, esp. if this happens through an extern in the near future.

vyzo (commented):
Overall I am in favour of these changes, some fine points in the comments.

sdk/src/error.rs
Stebalien: Well, I have three alternatives:

- Panic in the sdk method.
- Return an Option with no additional information.
- Return a Result with a syscall error number.

fvm/src/machine/mod.rs
vyzo: actually, it is u64 at the source (tipset) so i suggest we keep it as is.

sdk/src/error.rs
vyzo: yeah, lets just return some detail... maybe it will be useful one day :)

vyzo (approved):

PR #945, opened 2022-10-10
FIX: HashBits::next when bit_width does not divide 256 and the full hash is consumed

aakoshh:
Fixes `HashBits::next` returned an error when we have consumed all but the last bits of a key, and what's left is less than the bit width. The PR changes it so that it consumes whatever is left, and only fails on the next call when the whole key has already been consumed. 

The following table shows width of the last chunk of the 32 byte long hash with various bit widths:

| bit_width | last nibble width |
|-|-|
8 | 8 
7 | 4 
6 | 4
5 | 1
4 | 4
3 | 1
2 | 2
1 | 1

The reason this probably hasn't come up before is that with hashing it's very unlikely that to keys share all but their last bits. However with the hybrid hashing+offset based strategy used by Solidity this is exactly what happens for arrays, and then some bit widths work while others don't.


Stebalien (approved):
Forgot about this one. Thanks!

PR #946, opened 2022-10-10
EVM-892: Try putting data into the leaves.

aakoshh:
build on https://github.com/filecoin-project/ref-fvm/pull/935 
depends on https://github.com/filecoin-project/ref-fvm/pull/945

This PR adds a `push_data_into_leaves` option to the HAMT `Config` to see what would happen if we didn't store key-value pairs in intermediary nodes in the tree. The idea is that mapping entries cause write amplification which could be avoided this way.


aakoshh:
It was decided that this feature is not currently worth the extra code complexity. Since we're splitting out the data structure for the EVM, where it didn't bring any benefits, I'm not going to merge it.

PR #965, opened 2022-10-13
feat: remove RawBytes from APIs

Stebalien:
Use raw byte slices/vecs everywhere. Otherwise, this type kind of leaks into all APIs.

### DIFF
--- a/shared/src/message.rs
+++ b/shared/src/message.rs
@@ -2,9 +2,8 @@
 // SPDX-License-Identifier: Apache-2.0, MIT
 
 use anyhow::anyhow;
-use fvm_ipld_encoding::de::{Deserialize, Deserializer};
-use fvm_ipld_encoding::ser::{Serialize, Serializer};
-use fvm_ipld_encoding::{Cbor, RawBytes};
+use fvm_ipld_encoding::tuple::*;
+use fvm_ipld_encoding::{strict_bytes, Cbor};
 
 use crate::address::Address;
 use crate::econ::TokenAmount;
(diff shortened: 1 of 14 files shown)


Stebalien:
So, I got kind of annoyed at having a "special" type for byte vectors and tried simplifying things a bit. This is more of an experiment than anything but I'd like some feedback before we go further.

shared/src/message.rs
```diff
@@ -2,28 +2,33 @@
...
 
 use crate::address::Address;
 use crate::econ::TokenAmount;
 use crate::MethodNum;
 
 /// Default Unsigned VM message type which includes all data needed for a state transition
 #[cfg_attr(feature = "testing", derive(Default))]
-#[derive(PartialEq, Clone, Debug, Hash, Eq)]
+#[derive(PartialEq, Clone, Debug, Hash, Eq, Serialize_tuple, Deserialize_tuple)]
 pub struct Message {
```
Stebalien, line 15: I'm really confused as to why we manually deserialized/serialized this...

Stebalien:
In terms of "breakingness"... Given the runtime in the builtin actors, we can pretty easily upgrade them and slowly port them over.

Stebalien:
But after doing this... I think _most_ of these cases should be full on "blocks" with codecs.

anorth:
Needing to add `#[serde(with = "strict_bytes")]` everywhere its in a parameter block is a bit of a downside to me. One of the best things about TokenAmount was fixing that for bigints. There is a little bit of an overhaul needed in serialisation types/traits tho, e.g. so that `invoke` dispatch tables can have a fixed trait for param/return serialisable types and thus be macro-ified.

Stebalien:
Moving back to discussion-mode in https://github.com/filecoin-project/ref-fvm/issues/987.

PR #970, opened 2022-10-14
impl TryFrom<Payload> for DelegatedAddress

mriise:


shared/src/address/payload.rs
```diff
@@ -151,6 +151,17 @@ impl From<&Payload> for Protocol {
     }
 }
 
+impl TryFrom<Payload> for DelegatedAddress {
+    type Error = ();
+
+    fn try_from(value: Payload) -> Result<Self, Self::Error> {
+        match value {
+            Payload::Delegated(d) => Ok(d),
+            _ => Err(()),
```
Stebalien, line 160: We should probably do what we do for ID addresses and introduce a new `Error::NonDelegatedAddress` error. It's not great, but it's better than unit.

shared/src/address/payload.rs
```diff
@@ -151,6 +151,17 @@ impl From<&Payload> for Protocol {
     }
 }
 
+impl TryFrom<Payload> for DelegatedAddress {
```
Stebalien, line 154: Also implement by reference (not sure if necessary... but idiomatic).

Stebalien (commented):
nits but otherwise lgtm.

shared/src/address/payload.rs
mriise: not really less or more idiomatic IMO, its just whether or not we are going to reuse payload after this call.

Stebalien (approved):

PR #999, opened 2022-10-20
feat: resolve addresses by looking in the state-tree

Stebalien:
Before, we'd look in the actor state itself. But that's no longer necessary.

This patch:

1. Saves us a bit of gas/complexity.
2. Paves the way for allowing sends from f4 addresses.


mriise (approved):

raulk:
Awesome. This will require a migration though to hoist f1 and f3 addresses to the state tree, and to remove them from the internal state of the account actor.

PR #1002, opened 2022-10-21
Hack: allow f4-embryos to be used as accounts

Stebalien:
There are two commits, we can decide to drop the second:

1. Commit 1 allows sending from an f4 embryo.
2. Commit 2 allows validating messages from an f4 embryo. However, this likely requires https://github.com/filecoin-project/FIPs/discussions/497 to actually be useful. We can pull this off as a last minute hack, but it's not great.

### DIFF
--- a/fvm/Cargo.toml
+++ b/fvm/Cargo.toml
@@ -52,7 +52,7 @@ default-features = false
 features = ["cranelift", "pooling-allocator", "parallel-compilation"]
 
 [features]
-f4-as-account = []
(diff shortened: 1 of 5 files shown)


Stebalien:
Note: we _don't_ have to update fvm_shared in the actors, this just affects the crypto logic.

raulk (approved):

raulk:
Merging as a rebase since the commits are well differentiated.

PR #1005, opened 2022-10-21
EAM singleton

rllola:
Added EAM singleton in the integration testing module.


Stebalien (approved):
I've applied a few fixes. LGTM.

PR #1014, opened 2022-10-21
fix compile issues with f4-as-account feature.

raulk:
These were not caught in CI because we don't build with `--all-features` due to that triggering CUDA dependencies. Need to figure out a solution.

Also adds a test, even though it can only be triggered manually now by calling `cargo test` with `--feature f4-as-account`.

### DIFF
--- a/testing/integration/tests/embryo_sender_test.rs
+++ b/testing/integration/tests/embryo_sender_test.rs
@@ -1,0 +1,67 @@
+mod bundles;
+
+use bundles::*;
+use fvm::executor::{ApplyKind, Executor};
+use fvm::machine::Machine;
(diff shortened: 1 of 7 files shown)


testing/integration/tests/embryo_sender_test.rs
```diff
@@ -0,0 +1,67 @@
...
+    let balance = tester
+        .executor
+        .unwrap()
+        .state_tree()
+        .get_actor(&receiver)
+        .expect("couldn't find receiver actor")
+        .expect("actor state didn't exist")
+        .balance;
+
+    println!("{}", balance)
```
raulk, line 66: Need an actual assertion here.

Issue #1007, opened 2022-10-21
MEM: Reinstate test vectors


aakoshh:
Part of https://github.com/filecoin-project/ref-fvm/issues/851 

To measure the effects of loading memory and to determine a suitable minimum required memory in https://github.com/filecoin-project/ref-fvm/issues/369 we want to use the test vectors under https://github.com/filecoin-project/ref-fvm/tree/release/v2/testing/conformance

However the ones in the README refer to `specs_actor_v6` which have been removed, and https://github.com/filecoin-project/fvm-test-vectors/pull/7 which adds `specs_actor_v7` has not been merged yet. What we have now are the `fil_7_*` vectors under `0001-ohsnap-01`. According to @raulk it matters what we use for testing, and ideally a set of `V8` test vectors should be extracted from mainnet. 

This issue is just an update of the READMEs to remove the reference to v6 and start using v7.

Issue #851, opened 2022-09-08
Memory expansion gas


raulk:
Need to calculate the right values for memory expansion gas (expanding memory pages), and then implement the policy/formula.

Related:

- #341 
- #361
- #369

aakoshh:
I collected notes from a preliminary investigation in https://hackmd.io/yU4i53-7RmWXIHpaAYNn6Q?view

Issue #944, opened 2022-10-10
Fix TVX


Stebalien:
Currently, [tvx](https://github.com/filecoin-project/lotus/tree/master/cmd/tvx) doesn't use the FVM (and so fails to compute the correct gas values). Unfortunately, when made to use the FVM, it trips up over some incomplete state.

I've made an attempt to fix this at https://github.com/filecoin-project/lotus/tree/steb/fix-tvx-fvm, but I was never able to really solve the issue. Unfortunately, this is making it hard to extract test vectors for nv18.

maciejwitowski:
@Stebalien looks like [post code-freeze](https://github.com/filecoin-project/ref-fvm/milestone/19) to me. Wdyt?

Stebalien:
Yeah, that works for me. We won't be extracting test vectors till things stabilize anyways.

Stebalien:
Fixed by @arajasek

aakoshh:
There seems to be some [missing parentheses](https://github.com/filecoin-project/lotus/blob/master/cmd/tvx/extract_message.go#L419-L423) in the TVX code.

PR #1019, opened 2022-10-24
MEM-1007: Reinstate tests

aakoshh:
Resolves https://github.com/filecoin-project/ref-fvm/issues/1007 

## Testing V7 vectors

This PR runs the `conformance` tests against https://github.com/filecoin-project/fvm-test-vectors/pull/7 which contain test vectors for NetworkVersion=15 and ActorVersion=7. 

On this particular test-vector PR, which hasn't been merged since May, we still have the specs V6 tests which have since been [removed](https://github.com/filecoin-project/fvm-test-vectors/pull/6) from `master`. However because we still have that, I added some selectors and filters to skip these. 

The PR has some enhancements and fixes to the tests:
* Added the ability to handle `VECTOR` being a subdirectory of the corpus, not just a specific test
* Added the file name as a context to failing tests and failing test setups to make it easier to identify which vector requires non-existing network versions for example
* Fixed the CID in of post condition sanity check message

I executed the tests as follows:

```bash
TEST_VECTOR_POSTCONDITION_MISSING_ACTION=warn VECTOR=test-vectors/corpus/specs_actors_v7 cargo test conformance -- --nocapture  
```

## Cannot merge into master 

The PR is opened against a feature branch for https://github.com/filecoin-project/ref-fvm/issues/851 based on https://github.com/filecoin-project/ref-fvm/tree/release/v2 and not `master` of `ref-fvm` because on `master` we could no longer run V7 tests. The earliest [bundle loaded is V10](https://github.com/filecoin-project/ref-fvm/commit/08930e15a1725aa832802b33f63be627ac495c21#diff-233e3f8899cdbdbcb89b350177f2da7fa44a54de85e5a9deff112b527c8104b3) now and the tests were [disabled](https://github.com/filecoin-project/ref-fvm/commit/08930e15a1725aa832802b33f63be627ac495c21#diff-4f7536c054ed679e20336524a23d7244e6887277efa891d7532935d0b3405559). 

Maybe there's a way to load multiple actor bundles in the test, not just the latest?

_UPDATE_: I tried to cherry-pick all these changes onto `master` and load both V7 and V10 actor bundles. Unfortunately it didn't work because `NetworkVersion::V15` is [not supported](https://github.com/filecoin-project/ref-fvm/blob/master/fvm/src/machine/default.rs#L67) by the `DefaultMachine`, only V18 is. Here's the branch for what it's worth: https://github.com/filecoin-project/ref-fvm/tree/mem-1007-reinstate-tests-master

## Problems with the tests 

The tests under the _extracted_ directory work but there are some problems with the V7 specs. 

### Missing post-condition CIDs 

All tests seem to be caught out by the sanity check which ascertains that the post-condition CID is loaded into the blockstore before running the test; they are not. TBH I don't understand why they should be, because later we check that the return value of the test matches the post-condition - how can we be sure that it should be in the block store before applying the message? In any case I added a `TEST_VECTOR_POSTCONDITION_MISSING_ACTION` env var to allow us to still use these test vectors.

### Failing tests

There are two failing tests. 

#### Wrong exit code in bad sender
```
[FAIL] vector: test-vectors/corpus/specs_actors_v7/TestAggregateBadSender/6f9aa4df047387cdb61f7328f3697c99e53d6151045880902595ca1ac60d334e-t0101-t0102-storageminer-26.json | variant: ohsnap
	|> reason: exit code of msg 0 did not match; expected: ExitCode { value: 18 }, got ExitCode { value: 8 }. Error: message failed with backtrace:
00: f0102 (method 26) -- caller f0101 is not one of supported (8)
```
* The tests are built against actors version ~7.5
* Looking at how [validate_immediate_caller_is](https://github.com/filecoin-project/builtin-actors/blob/v7.2.0/actors/runtime/src/runtime/fvm.rs#L126) is implemented in 7.2 it returns `SYS_FORBIDDEN`
* The `miner` actor in version `7.2` is built against `ref-fvm` version `0.6.0` but the [exit code 8 is reserved](https://github.com/filecoin-project/ref-fvm/blob/fvm_shared%40v0.6.1/shared/src/error/mod.rs#L71)
* `SYS_FORBIDDEN` is defined in the `builtin-actors` own [runtime](https://github.com/filecoin-project/builtin-actors/blob/v7.2.0/actors/runtime/src/actor_error.rs#L15)
* In the latest version `9.1.0` this actually returns [forbidden](https://github.com/filecoin-project/builtin-actors/blob/v9.0.1/runtime/src/runtime/fvm.rs#L124) defined as [USR_FORBIDDEN](https://github.com/filecoin-project/builtin-actors/blob/v9.0.1/runtime/src/actor_error.rs#L31) which [has the expected code](https://github.com/filecoin-project/ref-fvm/blob/fvm_shared%40v0.6.1/shared/src/error/mod.rs#L92)
* It looks like there's [still confusion](https://github.com/filecoin-project/builtin-actors/blob/a6250c71cd099b781912b71ae4dec9809aec3fb9/actors/market/tests/on_miner_sectors_terminate.rs#L306) left in the code over what the correct return value is
* but there's an [explicit test vector commit](https://github.com/filecoin-project/fvm-test-vectors/pull/7/commits/95a9b412c3f659a9530d1a9083a68e844f4486ae) that confirms 8 is the expected behaviour.

So the test is right to fail, and this has been fixed in later versions of the `builtin-actors`; it looks [correct](https://github.com/filecoin-project/builtin-actors/blob/v8.0.0/runtime/src/runtime/fvm.rs#L129) in `8.0.0`.

#### Wrong exit code on partition index failure
```
[FAIL] vector: test-vectors/corpus/specs_actors_v7/TestWrongPartitionIndexFailure/b4e5b1bb610305fc8cf81bfb859c76b9acdacb109a591354e7a6b43bb8e7b61f-t0100-t0101-storageminer-27.json | variant: ohsnap
	|> reason: exit code of msg 0 did not match; expected: ExitCode { value: 17 }, got ExitCode { value: 16 }. Error: message failed with backtrace:
00: f0101 (method 27) -- error checking sector health (16)
```

* The miner actor returns [USR_ILLEGAL_ARGUMENT](https://github.com/filecoin-project/builtin-actors/blob/v7.2.0/actors/miner/src/lib.rs#L982)
* The test expected [USR_NOT_FOUND](https://github.com/filecoin-project/ref-fvm/blob/fvm_shared%40v0.6.1/shared/src/error/mod.rs#L90)
* which comes [later in prove_replica_updates](https://github.com/filecoin-project/builtin-actors/blob/v7.2.0/actors/miner/src/lib.rs#L1231), that corresponds to method 27
* :man_shrugging: 

## Next steps

Strictly speaking I don't need all the tests working to be able to collect the memory statistics, so I can go ahead like this. 

Ideally there would be a way to use the fixed actors for the network versions that the tests expect, but looking at [NetworkVersion](https://github.com/filecoin-project/ref-fvm/blob/master/shared/src/version/mod.rs) this doesn't seem to be the intention, with the next bundle intended for the next network version only.

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,14 +45,7 @@ Here's what you'll find in each directory:
   - Contains the test vector runner, as well as benchmarking utilities on top of it.
   - The conformance test runner feeds the test vector corpus located at https://github.com/filecoin-project/fvm-test-vectors into ref-fvm, in order to validate spec conformance.
   - The benchmarking utilities use the `criterion` Rust library to measure the performance and overhead of ref-fvm across various facets.
-  - Instructions
-    - To run a specific test vector, run `VECTOR=test-vectors/corpus/specs_actors_v6/REST_OF_TEST_VECTOR.json cargo test -- conformance --nocapture`
-    - To bench a specific test vector, run `VECTOR=test-vectors/corpus/specs_actors_v6/REST_OF_TEST_VECTOR.json cargo bench -- conformance --nocapture`
-    - To bench the system's overhead for the setup of the machine for a given test vector, run `VECTOR=test-vectors/corpus/specs_actors_v6/REST_OF_TEST_VECTOR.json cargo bench -- overhead --nocapture`. Note that the vector choice doesn't matter much, because the Machine initialization procedure is identicall for all vectors.
-    - To get a perf flamegraph, run `CARGO_PROFILE_BENCH_DEBUG=true VECTOR=testing/conformance/test-vectors/corpus/specs_actors_v6/REST_OF_TEST_VECTOR.json  cargo flamegraph --bench bench_conformance -- --nocapture`. The output SVG will be in `flamegraph.svg`.
-  - Overhead measurement scenarios. There are two overhead measurement scenarios included.
-    1. `bench_init_only`: measure the overhead of running the benchmark itself, it doesn't send any messages to the FVM to process.
-    2. `bench_500_simple_state_access`: measures the overhead of calling the `pubkey_address` method on an account actor 500 times, this is the most lightweight message possible to send that actually executes actor logic (unlike a bare send).
+  - See the [instructions](./testing/conformance/README.md#instructions) about how to run the tests and the benchmarks.
   - Disclaimers
     - Benchmarks are currently very slow to run, setup and teardown. This is due to using default WASM cache, and will be fixed soon.
 
@@ -63,7 +56,7 @@ Here's what you'll find in each directory:
 - Alpha:
   - Declared when: all test vectors passing, integrated into Lotus via FFI.
   - Focus: theoretical correctness.
-- Beta: 
+- Beta:
   - Declared when: all the above + syncing mainnet consistently, keeping up with chain consistently, i.e. when Phase 0 from the [FVM milestone roadmap](https://filecoin.io/blog/posts/introducing-the-filecoin-virtual-machine/) is reached.
   - Focus: production-readiness, performance, live consensus correctness.
 - RC:
--- a/testing/conformance/README.md
+++ b/testing/conformance/README.md
@@ -6,7 +6,7 @@ of tests and benchmarks.
 ## Instructions
 
 - To run all tests, just run `cargo test`.
-- To run all test vectors under a specific directory, run eg. `VECTOR=test-vectors/corpus/extracted cargo test conformance`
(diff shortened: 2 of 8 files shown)


Stebalien:
So, we removed these from FVM v3 because we've made breaking changes and no longer support anything before nv18. This is good for testing, but we'll need to fix TVX (#944) before we can actually move forward here (and extract test vectors from a test network).

Issue #1008, opened 2022-10-21
MEM: Configure Wasmtime to use the pooling allocator


aakoshh:
Part of https://github.com/filecoin-project/ref-fvm/issues/851

We agreed to try to use the pooling allocator instead of the on-demand one.

Issue #1010, opened 2022-10-21
MEM: Collect memory statistics with test vectors


aakoshh:
Part of https://github.com/filecoin-project/ref-fvm/issues/851

Depends on https://github.com/filecoin-project/ref-fvm/issues/1007 and https://github.com/filecoin-project/ref-fvm/issues/1009

Run the test vectors and use the `ResourceLimiter` to measure the minimum and maximum gas used by any of the calls.

PR #1027, opened 2022-10-25
MEM-1008: Use pooling

aakoshh:
Closes https://github.com/filecoin-project/ref-fvm/issues/1008 

## Changes

- [x] Added `NetworkConfig::max_inst_memory_bytes` which is used to configure the `wasmtime::Config::static_memory_maximum_size`
- [x] Use the `PoolingInstanceAllocator` 
- [x] Use `NetworkConfig::max_call_depth` to configure the `InstanceLimits` in the pool

### Unreachable
```
$ TEST_VECTOR_POSTCONDITION_MISSING_ACTION=ignore VECTOR=test-vectors/corpus/extracted/0006-ohsnap-01/fil_7_storageminer/ExtendSectorExpiration/SysErrOutOfGas/ext-0007-fil_7_storageminer-ExtendSectorExpiration-SysErrOutOfGas-3.json cargo test conformance -- --nocapture
...
[FAIL] vector: test-vectors/corpus/extracted/0006-ohsnap-01/fil_7_storageminer/ExtendSectorExpiration/SysErrOutOfGas/ext-0007-fil_7_storageminer-ExtendSectorExpiration-SysErrOutOfGas-3.json | variant: ohsnap
	|> reason: exit code of msg 0 did not match; expected: ExitCode { value: 7 }, got ExitCode { value: 4 }. Error: message failed with backtrace:
00: f0704941 (method 8) -- wasm `unreachable` instruction executed (4)
...
```
The error is produced where a [wasmtime Trap is turned into an Abort](https://github.com/filecoin-project/ref-fvm/blob/release/v2/fvm/src/syscalls/error.rs#L63). It originates from a [Trap::UnreachableCodeReached](https://github.com/bytecodealliance/wasmtime/blob/main/crates/wasmtime/src/trap.rs#L131) error.

After adding a `eprintln!("TRAP: {}", t);`, enabling backtraces and running the tests with `WASMTIME_BACKTRACE_DETAILS=1` I got the following extra info:

```
TRAP: wasm trap: wasm `unreachable` instruction executed
wasm backtrace:
    0: <unknown>!rust_oom
    1: <unknown>!__rg_oom
    2: <unknown>!__rust_alloc_error_handler
    3: <unknown>!alloc::alloc::handle_alloc_error::rt_error::hb1e1f60d94085472
    4: <unknown>!core::ops::function::FnOnce::call_once::h89a39298b2714490
    5: <unknown>!core::intrinsics::const_eval_select::h5c4a832c8650d6ed
    6: <unknown>!alloc::alloc::handle_alloc_error::he3013cfdd978b6f5
    7: <unknown>!alloc::raw_vec::handle_reserve::hf600f1e9108c55df
    8: <unknown>!alloc::raw_vec::RawVec<T,A>::reserve::do_reserve_and_handle::h2f3e81b1b52c0ebe
    9: <unknown>!alloc::vec::Vec<T,A>::reserve::h258af6d92a1a6b0b
   10: <unknown>!alloc::vec::Vec<T,A>::extend_desugared::h5f104e657ab9f33d
   11: <unknown>!<alloc::vec::Vec<T,A> as alloc::vec::spec_extend::SpecExtend<T,I>>::spec_extend::h57e699656d1b77b5
   12: <unknown>!<alloc::vec::Vec<T> as alloc::vec::spec_from_iter_nested::SpecFromIterNested<T,I>>::from_iter::ha8b0d9c2e8546a33
   13: <unknown>!<alloc::vec::Vec<T> as alloc::vec::spec_from_iter::SpecFromIter<T,I>>::from_iter::hbcfe19d4be10c310
   14: <unknown>!core::iter::adapters::try_process::habb22eef7121fc9e
   15: <unknown>!<fil_actors_runtime::runtime::fvm::FvmRuntime<B> as fil_actors_runtime::runtime::Runtime<B>>::transaction::h0fcb8b02e078b114
   16: <unknown>!fil_actor_miner::Actor::extend_sector_expiration::h1d9fdd9586e5b45b
   17: <unknown>!<fil_actor_miner::Actor as fil_actors_runtime::runtime::actor_code::ActorCode>::invoke_method::h765233dc279409cd
   18: <unknown>!fil_actors_runtime::runtime::fvm::trampoline::hd4eb9bb8e05e896c
   19: <unknown>!invoke
   20: <unknown>!<wasm function 6043>
```

So it run out of memory somewhere `extend_sector_expiration` [calling](https://github.com/filecoin-project/builtin-actors/blob/v7.2.0/actors/miner/src/lib.rs#L2040-L2259) `transaction` :thinking: 

EDIT: This has been fixed now.

### DIFF
--- a/fvm/src/machine/engine.rs
+++ b/fvm/src/machine/engine.rs
@@ -20,9 +20,6 @@ use crate::machine::NetworkConfig;
 use crate::syscalls::{bind_syscalls, InvocationData};
 use crate::Kernel;
 
-// Copied from `wasmtime_environ`.
-const WASM_PAGE_SIZE: u64 = 0x10000;
-
 /// A caching wasmtime engine.
 #[derive(Clone)]
 pub struct Engine(Arc<EngineInner>);
@@ -95,7 +92,7 @@ pub fn default_wasmtime_config(
             count: instance_count,
             // Adjust the maximum amount of host memory that can be committed to an instance to
             // match the static linear memory size we reserve for each slot.
-            memory_pages: instance_memory_maximum_size / WASM_PAGE_SIZE,
+            memory_pages: instance_memory_maximum_size / (wasmtime_environ::WASM_PAGE_SIZE as u64),
             ..Default::default()
         },
     });
--- a/fvm/src/machine/engine.rs
+++ b/fvm/src/machine/engine.rs
@@ -20,6 +20,9 @@ use crate::machine::NetworkConfig;
 use crate::syscalls::{bind_syscalls, InvocationData};
 use crate::Kernel;
 
+// Copied from `wasmtime_environ`.
+const WASM_PAGE_SIZE: u64 = 0x10000;
+
 /// A caching wasmtime engine.
 #[derive(Clone)]
 pub struct Engine(Arc<EngineInner>);
@@ -90,6 +93,9 @@ pub fn default_wasmtime_config(
         strategy: PoolingAllocationStrategy::ReuseAffinity,
         instance_limits: InstanceLimits {
             count: instance_count,
+            // Adjust the maximum amount of host memory that can be committed to an instance to
+            // match the static linear memory size we reserve for each slot.
+            memory_pages: instance_memory_maximum_size / WASM_PAGE_SIZE,
             ..Default::default()
         },
     });
--- a/fvm/src/machine/engine.rs
+++ b/fvm/src/machine/engine.rs
@@ -10,7 +10,10 @@ use fvm_ipld_blockstore::Blockstore;
 use fvm_wasm_instrument::gas_metering::GAS_COUNTER_NAME;
 use fvm_wasm_instrument::parity_wasm::elements;
 use wasmtime::OptLevel::Speed;
-use wasmtime::{Global, GlobalType, Linker, Memory, MemoryType, Module, Mutability, Val, ValType};
+use wasmtime::{
+    Global, GlobalType, InstanceAllocationStrategy, InstanceLimits, Linker, Memory, MemoryType,
+    Module, Mutability, PoolingAllocationStrategy, Val, ValType,
+};
 
 use crate::gas::WasmGasPrices;
 use crate::machine::NetworkConfig;
@@ -28,7 +31,9 @@ pub struct MultiEngine(Arc<Mutex<HashMap<EngineConfig, Engine>>>);
 /// The proper way of getting this struct is to convert from `NetworkConfig`
 #[derive(Clone, Eq, PartialEq, Hash)]
 pub struct EngineConfig {
+    pub max_call_depth: u32,
     pub max_wasm_stack: u32,
+    pub max_inst_memory_bytes: u64,
     pub wasm_prices: &'static WasmGasPrices,
     pub actor_redirect: Vec<(Cid, Cid)>,
 }
@@ -36,7 +41,9 @@ pub struct EngineConfig {
 impl From<&NetworkConfig> for EngineConfig {
     fn from(nc: &NetworkConfig) -> Self {
         EngineConfig {
+            max_call_depth: nc.max_call_depth,
             max_wasm_stack: nc.max_wasm_stack,
+            max_inst_memory_bytes: nc.max_inst_memory_bytes,
             wasm_prices: &nc.price_list.wasm_rules,
             actor_redirect: nc.actor_redirect.clone(),
         }
@@ -71,8 +78,24 @@ impl Default for MultiEngine {
     }
 }
 
-pub fn default_wasmtime_config() -> wasmtime::Config {
+pub fn default_wasmtime_config(
+    instance_count: u32,
+    instance_memory_maximum_size: u64,
+) -> wasmtime::Config {
     let mut c = wasmtime::Config::default();
+
+    // wasmtime default: OnDeand
+    // We want to pre-allocate all permissible memory to support the maximum allowed recursion limit.
+    c.allocation_strategy(InstanceAllocationStrategy::Pooling {
+        strategy: PoolingAllocationStrategy::ReuseAffinity,
+        instance_limits: InstanceLimits {
+            count: instance_count,
+            ..Default::default()
+        },
+    });
+
+    // wasmtime default: 4GB
+    c.static_memory_maximum_size(instance_memory_maximum_size);
 
     // wasmtime default: false
     // We don't want threads, there is no way to ensure determisism
@@ -168,7 +191,10 @@ impl Deref for Engine {
 
 impl Engine {
     pub fn new_default(ec: EngineConfig) -> anyhow::Result<Self> {
-        Engine::new(&default_wasmtime_config(), ec)
+        Engine::new(
+            &default_wasmtime_config(ec.max_call_depth, ec.max_inst_memory_bytes),
+            ec,
+        )
     }
 
     /// Create a new Engine from a wasmtime config.
@@ -372,6 +398,8 @@ impl Engine {
             last_milligas_available: 0,
             memory: self.0.dummy_memory,
         };
+
+        //        todo!("Configure resource limiter");
 
         let mut store = wasmtime::Store::new(&self.0.engine, id);
         let ggtype = GlobalType::new(ValType::I64, Mutability::Var);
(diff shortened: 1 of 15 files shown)


aakoshh:
@Stebalien @raulk I appreciate any ideas you might have on debugging where the unreachable error in the wasm code might come from :pray:

aakoshh:
Raising the memory to 8 and 16 GB didn't help. Since this test is for OutOfGas, maybe it will consume an unreasonable amount of memory. I will try to see how much it needed as part of https://github.com/filecoin-project/ref-fvm/issues/1010 using the OnDemand allocator and a ResourceLimiter.

Stebalien:
Hm. That makes no sense. We're using wasm32, so only ~3GiB should even be _addressable_. It should be impossible to _ask_ for more than that.

aakoshh:
Mystery solves, it was another config in https://github.com/filecoin-project/ref-fvm/pull/1027/commits/ee9d9797bc3449b4a5f34207b75a4e9c4c152f7a

The 10MB limit actually rang a bell, but I forgot about this one since I looked at the code last week.

fvm/src/machine/engine.rs
```diff
@@ -10,13 +10,19 @@ use fvm_ipld_blockstore::Blockstore;
...
+    Module, Mutability, PoolingAllocationStrategy, Val, ValType,
+};
 
 use crate::gas::WasmGasPrices;
 use crate::machine::NetworkConfig;
 use crate::syscalls::{bind_syscalls, InvocationData};
 use crate::Kernel;
 
+// Copied from `wasmtime_environ`.
+const WASM_PAGE_SIZE: u64 = 0x10000;
```
Stebalien, line 24: nit: comment that this is 16k.

fvm/src/machine/engine.rs
```diff
@@ -71,9 +81,32 @@ impl Default for MultiEngine {
     }
 }
 
-pub fn default_wasmtime_config() -> wasmtime::Config {
+pub fn default_wasmtime_config(
+    instance_count: u32,
+    instance_memory_maximum_size: u64,
```
Stebalien, line 86: These should ideally be defaults that we adjust later.

fvm/src/machine/engine.rs
```diff
@@ -71,9 +81,32 @@ impl Default for MultiEngine {
...
 
+    // wasmtime default: OnDeand
+    // We want to pre-allocate all permissible memory to support the maximum allowed recursion limit.
+    c.allocation_strategy(InstanceAllocationStrategy::Pooling {
+        strategy: PoolingAllocationStrategy::ReuseAffinity,
+        instance_limits: InstanceLimits {
+            count: instance_count,
+            // Adjust the maximum amount of host memory that can be committed to an instance to
+            // match the static linear memory size we reserve for each slot.
+            memory_pages: instance_memory_maximum_size / WASM_PAGE_SIZE,
```
Stebalien, line 98: nit: we should probably just ask the user to specify the number of pages (defined to be 16KiB per wasm spec).

Stebalien (commented):
Mostly nits, but this looks good to merge into master.

fvm/src/machine/engine.rs
aakoshh: Thanks, good point. It's actually 64KiB, so it's not intuitive at all. If you're okay with that, I'll add a dependency to `wasmtime_environ` itself to avoid duplicating it. The docs said that any crate other than `wasmtime` is considered "internal", and the comments on this suggested this is a WebAssembly specific value that won't change, but maybe this is the cleanest way to make sure we agree on the same value.

fvm/src/machine/engine.rs
aakoshh: I did it this way for a couple of reasons:
1. The defaults are defined on the `NetworkConfig` and it seemed better to have one set of defaults than two that can get out of sync and cause confusion.
2. As you see the `instance_memory_maximum_size` is used in multiple places: inside `InstanceLimits` and to set the value of `static_memory_maximum_size`. It seems like it would be difficult to ask that everyone using the config and wanting to set this remembers to set both values correctly. 
3. The parameters passed to `allocation_strategy` are more complex and the `instance_count` is just one of them; what we actually want to set is `EngineConfig::max_call_depth`, but this again is a dependent setting that you have to get right with all the other unrelated defaults like the affinity, and you can't just set it as a field, as it's part of an enum.

It seemed to me like the cleanest way to keep all relevant configuration visible in this method is to ask for these two values to be given to the method.

This method is only ever called from `Engine::new_default`, so maybe we could just make it private to this module, and then the scope of "later" is very clear.

fvm/src/machine/engine.rs
aakoshh: I wanted to stick to bytes because that seems to be the way in `wasmtime::Config`. If you look at [static_memory_maximum_size](https://github.com/bytecodealliance/wasmtime/blob/main/crates/wasmtime/src/config.rs#L1109) it also does this division. 

It seems better to not require the user to possess extra knowledge about the wasm spec. Especially since it seems like if 16KB is in the spec then this value is [different](https://github.com/bytecodealliance/wasmtime/blob/main/crates/environ/src/lib.rs#L65-L66). 

The docs of `memory_pages` say the default value of 160 means 10MB, which works out with 160*65,536=10,485,760. `pooling.rs` will [use WASM_PAGE_SIZE to limit memory](https://github.com/bytecodealliance/wasmtime/blob/main/crates/runtime/src/instance/allocator/pooling.rs#L746), so I think it's more natural to hide this from the user and just ask in terms of bytes.

aakoshh:
Need to debug this integration test error:

```
---- native_stack_overflow stdout ----
STACKOVERFLOW RESULT = ApplyRet { msg_receipt: Receipt { exit_code: ExitCode { value: 4 }, return_data: RawBytes { bytes: [] }, gas_used: 338095 }, penalty: TokenAmount(0.000001), miner_tip: TokenAmount(0.0), base_fee_burn: TokenAmount(0.0), over_estimation_burn: TokenAmount(0.0), refund: TokenAmount(0.0), gas_refund: 0, gas_burned: 9999661905, failure_info: Some(MessageBacktrace(Backtrace { frames: [Frame { source: 10000, method: 1, code: ExitCode { value: 4 }, message: "wasm `unreachable` instruction executed" }], cause: None })), exec_trace: [] }
STACKOVERFLOW RESULT = ApplyRet { msg_receipt: Receipt { exit_code: ExitCode { value: 10 }, return_data: RawBytes { bytes: [] }, gas_used: 10000000000 }, penalty: TokenAmount(0.000001), miner_tip: TokenAmount(0.0), base_fee_burn: TokenAmount(0.0), over_estimation_burn: TokenAmount(0.0), refund: TokenAmount(0.0), gas_refund: 0, gas_burned: 0, failure_info: Some(MessageBacktrace(Backtrace { frames: [
Frame { source: 10000, method: 1026, code: ExitCode { value: 10 }, message: "fatal error" }, 
Frame { source: 10000, method: 1025, code: ExitCode { value: 10 }, message: "fatal error" }, 
... { source: 10000, method: 2, code: ExitCode { value: 10 }, message: "fatal error" }
], 
cause: Some(Fatal { error_msg: "[from=f1d3nehuc4u3l5mn7hazppnogf3oe6l6ymaicbkhi, to=f010000, seq=1, m=2, h=0]: Limit of 1024 concurrent instances has been reached", backtrace: "disabled backtrace" }) })), exec_trace: [] }
thread 'native_stack_overflow' panicked at 'assertion failed: `(left == right)`
  left: `10`,
 right: `3221225475`', testing/integration/tests/main.rs:243:5
```

fvm/src/machine/engine.rs
Stebalien: My concern is rounding. Alternatively, https://github.com/filecoin-project/ref-fvm/pull/1073

fvm/src/machine/engine.rs
Stebalien: Yeah, makes sense. See https://github.com/filecoin-project/ref-fvm/pull/1073.

fvm/src/machine/engine.rs
aakoshh: I sort of swept it under my mental carpet by thinking any remainder will be truncated and we'll have something that will work. Or that we can add some validation somewhere to ensure it's a multiple of the page size.

PR #1036, opened 2022-10-26
MEM-1010: Memory statistics

aakoshh:
Closes https://github.com/filecoin-project/ref-fvm/issues/1010

This PR will make the resource limiter configurable and use it to collect memory statistics during test vector execution, so we can figure out how to configure the system. 

This, combined with temporarily reverting to the OnDemand instance allocator, will be first used to figure out what's going on in the test failure we see in https://github.com/filecoin-project/ref-fvm/pull/1027

The type and creation of the limiter was moved to the `Machine` which already has some associated types that carry over to the components that build on it, and provides an easy point of pluggability. It also has all the parameters to create a limiter for the default case.

The statistics are as follows with both the pooling and the on-demand allocator:

```
memory stats:
 init.min: 1179648
 init.max: 1179648
 exec.min: 1245184
 exec.max: 10747904
```

So instances start by asking for 1MB memory, and some don't go much higher, while others go up to 11MB.


Stebalien:
@aakoshh did you run these tests on conformance tests, or mainnet?

Stebalien:
Ah, conformance.

PR #1037, opened 2022-10-27
Add quickcheck::Arbitrary for some types

connormullett:
ChainSafe's project, Forest, recently started using `quickcheck` for property testing our custom implementations of `serde`. Part of quickcheck is using `Arbitrary` and we would like to push some of our changes upstream so we don't have to use wrappers and ease testing for new implementors in the future. Some of these types already have an arbitrary implementation from a different crate hidden behind the "arb" feature flag. I've implemented this side-by-side hidden behind #[cfg(test)] as to not mess with the regular codebase and apply breaking changes. 

The types that have arbitrary implemented are as follows:
- Signature
- SignatureType
- Address
- Payload
- DelegatedAddress
- Message
- PoStProof

The tracking issue in Forest can be found [here](https://github.com/ChainSafe/forest/issues/2130)


connormullett:
> Unfortunately, `cfg(test)` only applies to the crate being tested, not dependencies. You'll need to use put these behind the "arb" feature gate.

You're right, I'll stick them behind the feature gate. That should fit our needs

shared/src/message.rs
```diff
@@ -93,3 +93,21 @@ impl<'de> Deserialize<'de> for Message {
...
+
+#[cfg(feature = "arb")]
+impl quickcheck::Arbitrary for Message {
+    fn arbitrary(g: &mut quickcheck::Gen) -> Self {
+        Self {
+            to: Address::new_id(u64::arbitrary(g)),
+            from: Address::new_id(u64::arbitrary(g)),
+            version: i64::arbitrary(g),
+            sequence: u64::arbitrary(g),
+            value: TokenAmount::from_atto(u64::arbitrary(g)),
```
lemmih, line 105: Let's implement `Arbitrary` for `TokenAmount`.

```suggestion
            value: TokenAmount::arbitrary(g),
```

fvm/Cargo.toml
```diff
@@ -50,6 +50,7 @@ fvm = { path = ".", features = ["testing"], default-features = false }
 version = "1.0.1"
 default-features = false
 features = ["cranelift", "pooling-allocator", "parallel-compilation"]
+quickcheck = "1"
```
lemmih, line 53: This might be a copy&paste mistake. In any case, you haven't touched the `fvm` crate so this line shouldn't be here.

shared/Cargo.toml
```diff
@@ -43,6 +43,8 @@ rand = "0.8"
 rand_chacha = "0.3"
 serde_json = "1.0.56"
 multihash = { version = "0.16.3", default-features = false, features = ["multihash-impl", "sha2", "sha3", "ripemd"] }
+quickcheck = "1"
+arbitrary = "1.2.0"
```
lemmih, line 47: These aren't dev dependencies and they should both be optional.

shared/src/address/payload.rs
```diff
@@ -19,6 +19,20 @@ pub struct DelegatedAddress {
...
 
+#[cfg(feature = "arb")]
+impl quickcheck::Arbitrary for DelegatedAddress {
+    fn arbitrary(g: &mut quickcheck::Gen) -> Self {
+        Self {
+            namespace: ActorID::arbitrary(g),
+            length: usize::arbitrary(g),
+            buffer: arbitrary::Arbitrary::arbitrary(&mut arbitrary::Unstructured::new(
+                &Vec::arbitrary(g),
+            ))
```
lemmih, line 30: Avoid using the `arbitrary` crate. Everything should be doable using only `quickcheck`.

shared/src/message.rs
```diff
@@ -93,3 +93,21 @@ impl<'de> Deserialize<'de> for Message {
         })
     }
 }
+
+#[cfg(feature = "arb")]
+impl quickcheck::Arbitrary for Message {
+    fn arbitrary(g: &mut quickcheck::Gen) -> Self {
+        Self {
+            to: Address::new_id(u64::arbitrary(g)),
```
lemmih, line 101: We have an arbitrary implementation for `Address`. Let's use it:

```suggestion
            to: Address::arbitrary(g),
```

shared/src/sector/post.rs
```diff
@@ -29,6 +29,30 @@ pub struct PoStProof {
...
+                RegisteredPoStProof::StackedDRGWinning512MiBV1,
+                RegisteredPoStProof::StackedDRGWinning32GiBV1,
+                RegisteredPoStProof::StackedDRGWinning64GiBV1,
+                RegisteredPoStProof::StackedDRGWindow2KiBV1,
+                RegisteredPoStProof::StackedDRGWindow8MiBV1,
+                RegisteredPoStProof::StackedDRGWindow512MiBV1,
+                RegisteredPoStProof::StackedDRGWindow32GiBV1,
+                RegisteredPoStProof::StackedDRGWindow64GiBV1,
+            ])
+            .unwrap();
```
lemmih, line 48: Let's implement `Arbitrary` for `RegisteredPoStProof` separately.

lemmih (changes requested):

fvm/Cargo.toml
connormullett: The ActorState changes were going to be apart of this PR, opted to make it separate after this PR gets through. So yes, it is a copy paste mistake. Removing

shared/src/message.rs
connormullett: :+1:

shared/src/message.rs
connormullett: :+1:

shared/src/address/payload.rs
connormullett: I havent found a way to nicely create a [u8; 20] with quickcheck. Just using the arbitrary implementation we already had in Forest. Any recommendations would be nice.

shared/src/econ/mod.rs
```diff
@@ -106,6 +106,15 @@ impl fmt::Debug for TokenAmount {
     }
 }
 
+#[cfg(feature = "arb")]
+impl quickcheck::Arbitrary for TokenAmount {
+    fn arbitrary(g: &mut quickcheck::Gen) -> Self {
+        Self {
+            atto: BigInt::arbitrary(g),
+        }
```
lemmih, line 114:
```suggestion
        TokenAmount::from_atto(BigInt::arbitrary(g))
```

shared/src/message.rs
```diff
@@ -93,3 +93,21 @@ impl<'de> Deserialize<'de> for Message {
         })
     }
 }
+
+#[cfg(feature = "arb")]
+impl quickcheck::Arbitrary for Message {
+    fn arbitrary(g: &mut quickcheck::Gen) -> Self {
+        Self {
+            to: Address::arbitrary(g),
+            from: Address::new_id(u64::arbitrary(g)),
```
lemmih, line 102:
```suggestion
            from: Address::arbitrary(g),
```

shared/src/address/payload.rs
```diff
@@ -75,6 +87,30 @@ pub enum Payload {
...
 }
 
+#[cfg(feature = "arb")]
+impl quickcheck::Arbitrary for Payload {
+    fn arbitrary(g: &mut quickcheck::Gen) -> Self {
+        let id = Payload::ID(u64::arbitrary(g));
+        let secp = Payload::Secp256k1(
+            arbitrary::Arbitrary::arbitrary(&mut arbitrary::Unstructured::new(&Vec::arbitrary(g)))
+                .unwrap(),
+        );
```
lemmih, line 97: Avoid using the `arbitrary` crate.

lemmih (changes requested):

Stebalien:
Is this PR blocked on something?

connormullett:
> Is this PR blocked on something?

I'm no longer affiliated with ChainSafe and assumed someone would take over this as it's tied to the work I was doing. I'll finish this PR

shared/src/address/payload.rs
connormullett: figured out

shared/src/address/payload.rs
```diff
@@ -19,6 +19,18 @@ pub struct DelegatedAddress {
     buffer: [u8; MAX_SUBADDRESS_LEN],
 }
 
+#[cfg(feature = "arb")]
+impl quickcheck::Arbitrary for DelegatedAddress {
+    fn arbitrary(g: &mut quickcheck::Gen) -> Self {
+        let buffer: [u8; MAX_SUBADDRESS_LEN] = core::array::from_fn(|_| u8::arbitrary(g));
+        Self {
+            namespace: ActorID::arbitrary(g),
+            length: usize::arbitrary(g),
```
lemmih, line 28: This should be bounded by `MAX_SUBADDRESS_LEN`.

shared/src/address/payload.rs
```diff
@@ -75,6 +87,21 @@ pub enum Payload {
     Delegated(DelegatedAddress),
 }
 
+#[cfg(feature = "arb")]
+impl quickcheck::Arbitrary for Payload {
+    fn arbitrary(g: &mut quickcheck::Gen) -> Self {
+        let id = Payload::ID(u64::arbitrary(g));
+        let secp = Payload::Secp256k1([u8::arbitrary(g); PAYLOAD_HASH_LEN]);
```
lemmih, line 94: This generates a single arbitrary `u8` and then repeats it `PAYLOAD_HASH_LEN` times.

lemmih (changes requested):

lemmih:
> > Is this PR blocked on something?
> 
> I'm no longer affiliated with ChainSafe and assumed someone would take over this as it's tied to the work I was doing. I'll finish this PR

Thanks. Feel free to close this PR if it is in any way an inconvenience, though.

shared/src/address/payload.rs
connormullett: Fixed

shared/src/address/payload.rs
connormullett: fixed

lemmih (approved):
Looks good to me. Not sure why the checks are failing.

lemmih:
@Stebalien Everything should be good to go.

Stebalien (approved):

Stebalien:
Thanks!

PR #1039, opened 2022-10-31
MEM-1011: Charge gas for memory after execution.

aakoshh:
Charges gas for memory expansions after each Wasm execution. It doesn't charge during the expansions themselves because the reason for failure would be lost there. Current cost is `Zero`.


Stebalien:
Digging into this...

- We're going to need to charge as we allocate instead of afterwards.
- Instrumenting the code may actually be the best way to do this if it makes the memory limiter simpler. I assume charging _as_ we allocate inside the memory limiter would cause some annoying dependency cycles.
- We _also_ need to charge/limit for _initial_ memory (something I didn't even realize was a thing). https://github.com/filecoin-project/ref-fvm/issues/1055. We can handle this in a separate PR/work stream.

aakoshh:
@Stebalien thanks for the comments. 

> We're going to need to charge as we allocate instead of afterwards.

This is how this PR started, and I forgot to update the title until now, but I thought the way it currently works does exactly that. 

By tapping into the `Rules::memory_grow_cost` the code will be instrumented via injection so that it charges gas _before_ it's grown. Isn't that what you call allocation? 

> Instrumenting the code may actually be the best way to do this if it makes the memory limiter simpler. I assume charging as we allocate inside the memory limiter would cause some annoying dependency cycles.

Indeed, using the injection provided by the framework made the limiter much simpler. Trying to charge inside the limiter was problematic not necessarily from the dependency perspective but because the return value is just a `bool`, it would have been impossible to distinguish between hitting a memory limit and running out of gas. 

But this is not a problem with the injection approach as it will check the gas limits before trying to grow the memory, so the error we receive will be the same as running out of gas at any other point in wasm.

> We also need to charge/limit for initial memory (something I didn't even realize was a thing). https://github.com/filecoin-project/ref-fvm/issues/1055. We can handle this in a separate PR/work stream.

Just wanted to double check what's going on here. 

1. The pooling allocator [instantiates the contents of the memory and creates a static construct over it](https://github.com/bytecodealliance/wasmtime/blob/v2.0.1/crates/runtime/src/instance/allocator/pooling.rs#L403-L426). You might 
2. [MemoryImage::initialize](https://github.com/bytecodealliance/wasmtime/blob/v2.0.1/crates/runtime/src/cow.rs#L345) makes the initial memory size accessible and fills it with the appropriate contents, if available.
3. [Memory::new_static](https://github.com/bytecodealliance/wasmtime/blob/v2.0.1/crates/runtime/src/memory.rs#L634) calls `limit_new` to [check with the limiter](https://github.com/bytecodealliance/wasmtime/blob/v2.0.1/crates/runtime/src/memory.rs#L723) if the minimum/maximum are within limits. 

I think you're right that we have to do something about charging for the initial memory. The limits would be applied, but those are probably permissive enough not to cause an issue, apart from the fact that the memory is first initialised and _then_ the limits get a chance to reject them. But charging seems to be missing. 

I had an interim approach (one of the commits) where I estimated the available gas based on how much gas the user has, which would have provided a lower limit, and then I would have charged for it in the end. But with the instrumentation approach, it looks like the initial memory might not be something the module explicitly has to grow to, and thus escapes gas. It's hard to tell because it depends on how the Wasm code was generated. I will check try to check it with the tests; if the modules grow to their initial size I should see two calls with `memory_growing(current = 0, desired = minimum)`. 

Thanks for pointing this out!

aakoshh:
I will merge this for now and make a new PR for charging for initial memory; might make it easier to fix the resource limiter.

PR #1049, opened 2022-11-06
add support for actor events.

### LINKED ISSUE #1048: Epic: Actor events implementation
This issue tracks the implementation of FIP draft for Actor Events: https://github.com/filecoin-project/FIPs/pull/483.

## Basics

- [x] Common types in fvm_shared.
- [x] ~`validate` method in dagcbor codec to structurally validate event payloads~
	- [ ] https://github.com/ipld/libipld/issues/158
	- Moved to https://github.com/filecoin-project/ref-fvm/issues/1111.
- [x] `record_event` syscall in SDK and kernel; accumulation of events in CallManager.
- [x] Return events in receipts via Executor.
- [x] Snake through events via filecoin-ffi to client.
- [x] Gas hooks and (parameterized) price formulae for recording events.
- [X] ~Implement and integrate event store?~
	- Moved to https://github.com/filecoin-project/ref-fvm/issues/1108.

## EVM runtime integration

- [x] Implement LOG opcodes.

## Gas params

- [X] Parameterize gas for events.
	- Moved to https://github.com/filecoin-project/ref-fvm/issues/1109.

## PRs

- [x] ref-fvm
	- [x] https://github.com/filecoin-project/ref-fvm/pull/1049
- [x] builtin-actors
	- [x] https://github.com/filecoin-project/builtin-actors/pull/839
- [x] filecoin-ffi
	- [x] https://github.com/filecoin-project/filecoin-ffi/pull/332
- [x] fil-actor-utils
	- [x] https://github.com/filecoin-project/filecoin-actor-utils/pull/1

raulk:
Reference implementation of https://github.com/filecoin-project/FIPs/pull/483.
Part of #1048.

## TODO
- [x] Make events AMT root in Receipt optional.
- [x] Implement gas handling.
- [ ] ~Validate that values in event entries are valid DAG-CBOR.~
	- Moved to #1082.
- [ ] ~Macro for emitting events.~
	- P3, not necessary at this stage; will be more impactful with native actors (M2.2).

### DIFF
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -53,7 +53,7 @@ pub struct InnerDefaultCallManager<M: Machine> {
     /// Limits on memory throughout the execution.
     limits: M::Limiter,
     /// Accumulator for events emitted in this call stack.
-    events: Vec<StampedEvent>,
+    events: EventsAccumulator,
 }
 
 #[doc(hidden)]
@@ -103,7 +103,7 @@ where
             exec_trace: vec![],
             invocation_count: 0,
             limits,
-            events: vec![],
+            events: Default::default(),
         })))
     }
 
@@ -166,11 +166,20 @@ where
         f: impl FnOnce(&mut Self) -> Result<InvocationResult>,
     ) -> Result<InvocationResult> {
         self.state_tree_mut().begin_transaction();
+        self.events.create_layer();
+
         let (revert, res) = match f(self) {
             Ok(v) => (!v.exit_code().is_success(), Ok(v)),
             Err(e) => (true, Err(e)),
         };
         self.state_tree_mut().end_transaction(revert)?;
+
+        if revert {
+            self.events.discard_last_layer();
+        } else {
+            self.events.merge_last_layer();
+        }
+
         res
     }
 
@@ -191,6 +200,8 @@ where
         if machine.context().tracing {
             exec_trace.extend(gas_tracker.drain_trace().map(ExecutionEvent::GasCharge));
         }
+
+        let events = events.finish();
 
         (
             FinishRet {
@@ -244,7 +255,7 @@ where
     }
 
     fn append_event(&mut self, evt: StampedEvent) {
-        self.events.push(evt)
+        self.events.append_event(evt)
     }
 }
 
@@ -576,5 +587,69 @@ where
         );
         self.call_stack_depth -= 1;
         res
+    }
+}
+
+/// Stores events in layers as they are emitted by actors. As the call stack progresses, when an
+/// actor exits normally, its events should be merged onto the previous layer (merge_last_layer).
+/// If an actor aborts, the last layer should be discarded (discard_last_layer). This will also
+/// throw away any events collected from subcalls (and previously merged as those subcalls returned
+/// normally).
+///
+/// For things to operate normally, this struct must be constructed with an empty vector as the
+/// first layer. This is where final retained events will eventually be folded into. The Default
+/// implementation constructs this object correctly.
+pub struct EventsAccumulator(Vec<Vec<StampedEvent>>);
+
+impl EventsAccumulator {
+    fn append_event(&mut self, evt: StampedEvent) {
+        self.0
+            .last_mut()
+            .expect("expected an event accumulator layer on append_event")
+            .push(evt)
+    }
+
+    fn create_layer(&mut self) {
+        self.0.push(Default::default())
+    }
+
+    fn merge_last_layer(&mut self) {
+        let mut last = self
+            .0
+            .pop()
+            .expect("expected an event accumulator layer on merge_last_layer");
+
+        if !last.is_empty() {
+            let prev = self
+                .0
+                .last_mut()
+                .expect("expected second event accumulator layer on merge_last_layer");
+            if prev.is_empty() {
+                let _ = std::mem::replace::<Vec<StampedEvent>>(prev, last);
+            } else {
+                prev.append(&mut last)
+            }
+        }
+    }
+
+    fn discard_last_layer(&mut self) {
+        self.0
+            .pop()
+            .expect("expected an event accumulator layer on discard_last_layer");
+    }
+
+    fn finish(mut self) -> Vec<StampedEvent> {
+        let ret = self
+            .0
+            .pop()
+            .expect("expected final event accumulator layer on finish");
+        assert!(self.0.is_empty(), "non-empty event accumulator on finish");
+        ret
+    }
+}
+
+impl Default for EventsAccumulator {
+    fn default() -> Self {
+        EventsAccumulator(vec![Vec::default()])
     }
 }
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -103,7 +103,7 @@ where
             exec_trace: vec![],
             invocation_count: 0,
             limits,
-            events: Vec::with_capacity(8),
+            events: vec![],
         })))
     }
 
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -593,63 +593,36 @@ where
 /// Stores events in layers as they are emitted by actors. As the call stack progresses, when an
 /// actor exits normally, its events should be merged onto the previous layer (merge_last_layer).
 /// If an actor aborts, the last layer should be discarded (discard_last_layer). This will also
-/// throw away any events collected from subcalls (and previously merged as those subcalls returned
+/// throw away any events collected from subcalls (and previously merged, as those subcalls returned
 /// normally).
-///
-/// For things to operate normally, this struct must be constructed with an empty vector as the
-/// first layer. This is where final retained events will eventually be folded into. The Default
-/// implementation constructs this object correctly.
-pub struct EventsAccumulator(Vec<Vec<StampedEvent>>);
+#[derive(Default)]
+pub struct EventsAccumulator(Vec<StampedEvent>, Vec<usize>);
 
 impl EventsAccumulator {
     fn append_event(&mut self, evt: StampedEvent) {
-        self.0
-            .last_mut()
-            .expect("expected an event accumulator layer on append_event")
-            .push(evt)
+        self.0.push(evt)
     }
 
     fn create_layer(&mut self) {
-        self.0.push(Default::default())
+        self.1.push(self.0.len());
     }
 
     fn merge_last_layer(&mut self) {
-        let mut last = self
-            .0
+        self.1
             .pop()
-            .expect("expected an event accumulator layer on merge_last_layer");
-
-        if !last.is_empty() {
-            let prev = self
-                .0
-                .last_mut()
-                .expect("expected second event accumulator layer on merge_last_layer");
-            if prev.is_empty() {
-                let _ = std::mem::replace::<Vec<StampedEvent>>(prev, last);
-            } else {
-                prev.append(&mut last)
-            }
-        }
+            .expect("no index in the event accumulator when calling merge_last_layer");
     }
 
     fn discard_last_layer(&mut self) {
+        let idx = self
+            .1
+            .pop()
+            .expect("no index in the event accumulator when calling discard_last_layer");
+        self.0.truncate(idx)
+    }
+
+    fn finish(self) -> Vec<StampedEvent> {
+        assert!(self.1.is_empty(), "non-empty event accumulator on finish");
         self.0
-            .pop()
-            .expect("expected an event accumulator layer on discard_last_layer");
-    }
-
-    fn finish(mut self) -> Vec<StampedEvent> {
-        let ret = self
-            .0
-            .pop()
-            .expect("expected final event accumulator layer on finish");
-        assert!(self.0.is_empty(), "non-empty event accumulator on finish");
-        ret
-    }
-}
-
-impl Default for EventsAccumulator {
-    fn default() -> Self {
-        EventsAccumulator(vec![Vec::default()])
     }
 }
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -175,9 +175,9 @@ where
         self.state_tree_mut().end_transaction(revert)?;
 
         if revert {
-            self.events.discard_last_layer();
+            self.events.discard_last_layer()?;
         } else {
-            self.events.merge_last_layer();
+            self.events.merge_last_layer()?;
         }
 
         res
@@ -596,33 +596,42 @@ where
 /// throw away any events collected from subcalls (and previously merged, as those subcalls returned
 /// normally).
 #[derive(Default)]
-pub struct EventsAccumulator(Vec<StampedEvent>, Vec<usize>);
+pub struct EventsAccumulator {
+    events: Vec<StampedEvent>,
+    idxs: Vec<usize>,
+}
 
 impl EventsAccumulator {
     fn append_event(&mut self, evt: StampedEvent) {
-        self.0.push(evt)
+        self.events.push(evt)
     }
 
     fn create_layer(&mut self) {
-        self.1.push(self.0.len());
+        self.idxs.push(self.events.len());
     }
 
-    fn merge_last_layer(&mut self) {
-        self.1
-            .pop()
-            .expect("no index in the event accumulator when calling merge_last_layer");
+    fn merge_last_layer(&mut self) -> Result<()> {
+        self.idxs.pop().map(|_| {}).ok_or_else(|| {
+            ExecutionError::Fatal(anyhow!(
+                "no index in the event accumulator when calling merge_last_layer"
+            ))
+        })
     }
 
-    fn discard_last_layer(&mut self) {
-        let idx = self
-            .1
-            .pop()
-            .expect("no index in the event accumulator when calling discard_last_layer");
-        self.0.truncate(idx)
+    fn discard_last_layer(&mut self) -> Result<()> {
+        let idx = self.idxs.pop().ok_or_else(|| {
+            ExecutionError::Fatal(anyhow!(
+                "no index in the event accumulator when calling discard_last_layer"
+            ))
+        })?;
+        self.events.truncate(idx);
+        Ok(())
     }
 
     fn finish(self) -> Vec<StampedEvent> {
-        assert!(self.1.is_empty(), "non-empty event accumulator on finish");
-        self.0
+        // Ideally would assert here, but there's risk of poisoning the Machine.
+        // Cannot return a Result because the call site expects infallibility.
+        // assert!(self.idxs.is_empty());
+        self.events
     }
 }
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -4,6 +4,7 @@ use fvm_ipld_encoding::{to_vec, RawBytes, DAG_CBOR};
 use fvm_shared::address::{Address, Payload};
 use fvm_shared::econ::TokenAmount;
 use fvm_shared::error::{ErrorNumber, ExitCode};
+use fvm_shared::event::StampedEvent;
 use fvm_shared::sys::BlockId;
 use fvm_shared::{ActorID, MethodNum, METHOD_SEND};
 use num_traits::Zero;
@@ -48,6 +49,8 @@ pub struct InnerDefaultCallManager<M> {
     exec_trace: ExecutionTrace,
     /// Number of actors that have been invoked in this message execution.
     invocation_count: u64,
+    /// Accumulator for events emitted in this call stack.
+    events: Vec<StampedEvent>,
 }
 
 #[doc(hidden)]
@@ -93,6 +96,7 @@ where
             backtrace: Backtrace::default(),
             exec_trace: vec![],
             invocation_count: 0,
+            events: Vec::with_capacity(8),
         })))
     }
 
@@ -186,6 +190,7 @@ where
             backtrace,
             mut gas_tracker,
             mut exec_trace,
+            events,
             ..
         } = *self.0.take().expect("call manager is poisoned");
 
@@ -202,6 +207,7 @@ where
                 gas_used,
                 backtrace,
                 exec_trace,
+                events,
             },
             machine,
         )
@@ -245,6 +251,10 @@ where
 
     fn invocation_count(&self) -> u64 {
         self.invocation_count
+    }
+
+    fn append_event(&mut self, evt: StampedEvent) {
+        self.events.push(evt)
     }
 }
 
--- a/fvm/src/executor/default.rs
+++ b/fvm/src/executor/default.rs
@@ -16,7 +16,7 @@ use fvm_shared::ActorID;
 use num_traits::Zero;
 
 use super::{ApplyFailure, ApplyKind, ApplyRet, Executor};
-use crate::call_manager::{backtrace, CallManager, InvocationResult};
+use crate::call_manager::{backtrace, Backtrace, CallManager, InvocationResult};
 use crate::gas::{Gas, GasCharge, GasOutputs};
 use crate::kernel::{Block, ClassifyResult, Context as _, ExecutionError, Kernel};
 use crate::machine::{Machine, BURNT_FUNDS_ACTOR_ADDR, REWARD_ACTOR_ADDR};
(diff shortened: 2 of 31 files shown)


fvm/src/gas/price_list.rs
```diff
@@ -271,6 +277,150 @@ lazy_static! {
...
             exec_instruction_cost: Gas::new(4),
         },
+
+        event_emit_base_cost: Zero::zero(),
+        event_per_entry_cost: Zero::zero(),
+        event_entry_index_cost: Zero::zero(),
+        event_per_byte_cost: Zero::zero(),
+    };
+
+    static ref HYGGE_PRICES: PriceList = PriceList {
```
raulk, line 287: giphy (9)

fvm/src/gas/price_list.rs
```diff
@@ -684,13 +840,35 @@ impl PriceList {
...
+    }
 }
 
 /// Returns gas price list by NetworkVersion for gas consumption.
 pub fn price_list_by_network_version(network_version: NetworkVersion) -> &'static PriceList {
     match network_version {
         NetworkVersion::V15 => &OH_SNAP_PRICES,
-        _ => &SKYR_PRICES,
+        NetworkVersion::V16 | NetworkVersion::V17 => &SKYR_PRICES,
+        NetworkVersion::V18 => &HYGGE_PRICES,
```
raulk, line 870: 👀

fvm/src/machine/default.rs
```diff
@@ -261,6 +265,33 @@ where
...
+        if events.is_empty() {
+            return Ok(None);
+        }
+
+        let blockstore = self.blockstore();
+
+        let amt_cid = {
+            let mut amt = Amt::new_with_bit_width(blockstore, EVENTS_AMT_BITWIDTH);
+            // TODO this can be zero-copy if the AMT supports a batch set operation that takes an
+            //  iterator of references and flushes the batch at the end.
```
raulk, line 278: @Stebalien any quick ideas here?

fvm/src/gas/price_list.rs
```diff
@@ -703,13 +859,35 @@ impl PriceList {
     pub fn grow_memory_gas(&self, grow_memory_bytes: usize) -> Gas {
         self.wasm_rules.memory_expansion_per_byte_cost * grow_memory_bytes as i64
     }
+
+    #[inline]
+    pub fn on_actor_event(&self, evt: &ActorEvent) -> GasCharge {
```
mriise, line 864: this follows with naming of the other gas methods, but knee-jerk reaction of this feels like its part of an event lib (maybe its my just JS showing though)

fvm/src/machine/boxed.rs
```diff
@@ -60,13 +61,13 @@ impl<M: Machine> Machine for Box<M> {
     }
 
     #[inline(always)]
-    fn into_store(self) -> Self::Blockstore {
-        (*self).into_store()
+    fn flush(&mut self) -> Result<Cid> {
+        (**self).flush()
```
mriise, line 65: ? clippy or something cause this re-ordering?

vyzo (commented):
looks ok, but i want steb review here.

shared/src/version/mod.rs
```diff
@@ -32,21 +32,21 @@ pub enum NetworkVersion {
...
     /// HyperDrive
     V13,
     /// Chocolate v6
     V14,
     /// OhSnap v7
     V15,
-    /// Skyr v8
+    /// Skyr (builtin-actors v8)
     V16,
-    /// Candy v9
```
mriise, line 47: RIP candy

testing/integration/tests/events_test.rs
```diff
@@ -0,0 +1,92 @@
+mod bundles;
+use bundles::*;
+use fil_events_actor::WASM_BINARY as ADDRESS_BINARY;
```
mriise, line 3: `EVENTS_BINARY`?

mriise (approved):
LGTM

As for doing zero copy things for AMT, it would likely require introducing a lifetime to Nodes (leading to the classic rust lifetime parameter creep).

We could do type shenanigans but that would be a larger project probably...

fvm/src/call_manager/default.rs
```diff
@@ -100,6 +103,7 @@ where
             exec_trace: vec![],
             invocation_count: 0,
             limits,
+            events: Vec::with_capacity(8),
```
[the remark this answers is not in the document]
raulk: It's hard to predict; it's possible that the chain becomes dominated by EVM smart contracts, most of which emit events. A 256 byte pre-allocation is a fair price to pay to minimise re-allocations (if we start empty and grow from there). Alternatively we could use a LinkedList here.

fvm/src/gas/price_list.rs
raulk: :D Probably your JS history plays a major role in that bias here. It didn't even occur to me.

fvm/src/machine/boxed.rs
raulk: Implementation order now follows trait declaration order.

shared/src/version/mod.rs
raulk: Ecstatically so.

testing/integration/tests/events_test.rs
raulk: Yep, will fix.

fvm/src/executor/default.rs
```diff
@@ -66,46 +68,62 @@ where
...
+                            value.as_ref().map(|v| v.size() as usize).unwrap_or(0),
+                        ))?;
+                    }
+
+                    Ok(ret)
+                });
+                let (res, machine) = cm.finish();
+
+                // Flush all events to the store.
+                let events_root = match machine.commit_events(res.events.as_slice()) {
```
raulk, line 110: All of the above is noise due to a newline introduced by `cargo fmt`. Here is where the change actually starts.

fvm/src/syscalls/event.rs
```diff
@@ -0,0 +1,14 @@
...
+use crate::kernel::Result;
+use crate::Kernel;
+
+pub fn emit_event(
+    context: Context<'_, impl Kernel>,
+    event_off: u32, // ActorEvent
+    event_len: u32,
+) -> Result<()> {
+    let evt: ActorEvent = context.memory.read_cbor(event_off, event_len)?;
+    context.kernel.emit_event(evt)
```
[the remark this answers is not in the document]
Stebalien: So, um. I'm going to make you hate me. We should do this with a gas-charging serde visitor. But we can do that in a followup patch (I'm "happy" to write it).

Basically, we can charge gas _as_ we deserialize and stop immediately if/when we run out without ever doing any work we haven't charged for.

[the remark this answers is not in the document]
Stebalien: (we can probably extend this concept to other `read_cbor` cases)

fvm/tests/dummy.rs
```diff
@@ -201,6 +203,10 @@ impl Machine for DummyMachine {
     fn new_limiter(&self) -> Self::Limiter {
         DummyLimiter::default()
     }
+
+    fn commit_events(&self, _events: &[StampedEvent]) -> kernel::Result<Option<Cid>> {
+        todo!()
```
Stebalien, line 208: Implementing this as a helper function would let you skip these todo's. (or you could give them a default implementation, up to you).

shared/src/event/mod.rs
```diff
@@ -0,0 +1,70 @@
...
+
+/// A key value entry inside an Event.
+#[derive(Serialize_tuple, Deserialize_tuple, PartialEq, Eq, Clone, Debug)]
+pub struct Entry {
+    /// A bitmap conveying metadata or hints about this entry.
+    pub flags: Flags,
+    /// The key of this event.
+    pub key: String,
+    /// Any DAG-CBOR encodeable type.
+    pub value: RawBytes,
```
Stebalien, line 60: It would be nice if this were an `IgnoreAny` (or `Ipld` if we wanted to actually keep it). By using `RawBytes`, we've "hidden" any internal CIDs from the codec.

But I don't feel _too_ strongly about that.

fvm/src/machine/default.rs
Stebalien: https://github.com/filecoin-project/ref-fvm/pull/1083

shared/src/event/mod.rs
Stebalien: Ok, I guess I feel somewhat strongly about this. I'd say that the value should either by "arbitrary bytes", or an _actual_ CBOR value. Having cbor as a byte-string within another cbor-value is icky.

fvm/src/call_manager/default.rs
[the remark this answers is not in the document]
raulk: Ack, no pre-allocation it is.

fvm/src/executor/default.rs
```diff
@@ -66,46 +68,62 @@ where
...
-                return (Err(e), cm.finish().1);
-            }
-
-            let params = if msg.params.is_empty() {
-                None
-            } else {
-                Some(Block::new(DAG_CBOR, msg.params.bytes()))
-            };
+        let (res, gas_used, mut backtrace, exec_trace, events_root, events) =
+            self.map_machine(|machine| {
```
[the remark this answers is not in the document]
raulk: Sure, I can do that in this PR.

fvm/src/syscalls/event.rs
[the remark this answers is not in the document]
raulk: We already have these gas fees in place here:

1. Syscall fee, which we had discussed to extend with dynamic prices based on parameter lengths.
2. Event emission -- flat rate per event.
3. Event emission -- per byte cost of serialized event. This is a storage cost.
4. Event emission -- per indexed field.

In a **_perfect_** world, we would charge for the cost of serde by metering it -- we have discussed about instrumenting codec logic and charging gas for it, but my impression is that that's a _tomorrow_ thing.

In absence of the above, what would the pricing policy of this gas-charging serde visitor be? Would we charge for anything other than bytes consumed? Can't we include the projected cost for that in (1) or (3)?

fvm/src/machine/mod.rs
```diff
@@ -97,6 +98,10 @@ pub trait Machine: 'static {
 
     /// Creates a new limiter to track the resources of a message execution.
     fn new_limiter(&self) -> Self::Limiter;
+
+    /// Commits the events to the machine by building the events AMT, and making sure that events
+    /// are written to the store.
+    fn commit_events(&self, events: &[StampedEvent]) -> Result<Option<Cid>>;
```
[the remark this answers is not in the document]
raulk: In the future it will, though. We are planning to fold events into a block-level index structure at some point. My understanding is that the logic to populate that data structure would go here, and be returned when finalizing the Machine.

fvm/tests/dummy.rs
raulk: Ack but https://github.com/filecoin-project/ref-fvm/pull/1049#discussion_r1023165039.

shared/src/event/mod.rs
raulk: I agree with the concern. But in practice I'm not convinced it changes much. What's important is that the validation logic is in place somewhere. This being represented as an `Ipld::Ipld` here makes things harder for actor code (because `Entry` is a shared type). An `IgnoredAny` would simply not work for the same reason. We would need to bifurcate this type into the FVM-side (which could use `IgnoredAny` or  `Ipld::Ipld`) and the SDK-side type. And I'm not sure what the right field type would be for the latter if not `RawBytes`?

fvm/src/machine/default.rs
raulk: Awesome, thanks! But we have one impediment to using that API as-is. Would love you hear your thoughts here: https://github.com/filecoin-project/ref-fvm/pull/1083#discussion_r1022683506.

fvm/src/call_manager/default.rs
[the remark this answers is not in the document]
raulk: 547d8c27

fvm/src/syscalls/event.rs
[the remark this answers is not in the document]
Stebalien: The problem with the current version is that we charge _after_ processing the event. If we use a serde visitor, se'd be able to charge for 2-4 while parsing (and stop parsing early if we run out of gas).

fvm/src/machine/mod.rs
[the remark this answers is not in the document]
Stebalien: Hm. I _assume_ that logic would go in the client. But I could see it going here as well.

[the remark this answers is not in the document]
Stebalien: (basically, the FVM usually deals with the state-tree only)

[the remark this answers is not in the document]
raulk: Assuming that we _don't_ return events to the client in ApplyRet (which we're doing now, but because to support transient call cases, and not sure if it needs to be there, might remove), doing it there would involve the extra work of re-fetching the events from the store _just_ to populate the structure.

FWIW, I think forming the receipts HAMT should _also_ be done in the Machine, with us returning the receipt_root to the client. IMO we whould trend towards consolidating all consensus-relevant execution logic inside the FVM. Right now there is still a gray area, but I'm strongly advocating for that future event indexing logic to live here.

fvm/src/call_manager/default.rs
```diff
@@ -51,6 +52,8 @@ pub struct InnerDefaultCallManager<M: Machine> {
     invocation_count: u64,
     /// Limits on memory throughout the execution.
     limits: M::Limiter,
+    /// Accumulator for events emitted in this call stack.
+    events: Vec<StampedEvent>,
```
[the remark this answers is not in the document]
raulk: Yeah, this is unfortunately a bit of a can of worms.

1. Filecoin does not revert side effects on non-zero exit codes (i.e. you can change your state root and abort with a non-zero exit code, and the protocol doesn’t revert your state change).
2. Events are trickier though, because they are fire-and-forget. Once an event is emitted, you can't take a compensating action to un-emit it.
    - By contrast, you can always revert a state root update. For example, you track your previous state root, update your state root, detect a failure scenario, and revert back to your original state root, before returning a non-zero exit code.
3. Ethereum _does_ revert side effects (including storage updates) on REVERT. We currently enforce the rollback of storage mutations by using a transaction block in the EVM runtime (i.e. entirely actor-side), but this begs to introduce protocol support for rollback on failure, which would create a new precedent.

I can think of two ways to handle this:

1. Layer events in the call manager and the last layer on a non-zero exit. This would prevent us from intentionally emitting events on failure (e.g. I could reasonably conceive a "window post failed" event).
2. Enable actors to throw away events emitted through their execution. The EVM runtime would use this feature, and we could make the transactional Runtime discard events by default on failure. Future native actors (or built-in actors) could choose retain events on failure.

shared/src/event/mod.rs
Stebalien: So, these behave _completely_ differently.

- `Ipld` and `IgnoreAny` means the data would actually be cbor.
- `RawBytes` means that the data would be a byte array (that _may_ be cbor if we validate it).

The important thing is that any links in that raw data wouldn't be resolvable. I.e., any IPLD tooling would treat it as "just bytes".

Stebalien: > This being represented as an Ipld::Ipld here makes things harder for actor code (because Entry is a shared type). An IgnoredAny would simply not work for the same reason.

The "right" solution is likely to make the `Event` generic over this type. Then we can use whatever type we want for the value, depending on the context.

- When validating, we can use `IgnoredAny` (once we fix it to actually validate fully, but we don't really care about validation right now anyways).
- The most "general" version is `Ipld`, it just gets expensive because it allocates a lot.
- We should implement a `RawValue` equivalent for IPLD: https://docs.rs/serde_json/latest/serde_json/value/struct.RawValue.html. We just don't have that right now.

raulk: > Ipld and IgnoreAny means the data would actually be cbor.

It just changes the timing at which you verify. Using `Ipld` or `IgnoreAny` would give us that guarantee during deserialization. Validating explicitly gives us that guarantee later. But in all cases you would walk away guaranteeing that the input is DAG-CBOR.

> The important thing is that any links in that raw data wouldn't be resolvable. I.e., any IPLD tooling would treat it as "just bytes".

How much of a problem is that?

Inside the FVM, we never expect to resolve these links (for all we know, they could be pointing at some NFT stored in the public IPFS network). I'd argue that we have no interest in processing Entry values at all, beyond validating that they're correct DAG-CBOR. And the reason we enforce DAG-CBOR is to give observers (which will be vary varied) _some_ guarantees about the structure of the data. Otherwise, they will have zero ability to reason about the events they observe.

fvm/src/syscalls/event.rs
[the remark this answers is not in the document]
raulk: Would the syscall fee with parameter length costs, charged at the time of the syscall, cover the concern here?

What I'm trying to figure out is what exactly are we charging for with this gas-charging serde visitor, if we are not charging for compute? Aren't we just charging for bytes (which can be covered in a simpler manner through (1))?

fvm/src/call_manager/default.rs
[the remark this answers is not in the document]
Stebalien: No, Filecoin totally reverts everything on non-zero exit codes.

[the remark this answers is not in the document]
arajasek: ^What Steb said -- we do so in both [the legacy VM](https://github.com/filecoin-project/lotus/blob/8df5b81f24d66c82d409d57a1ccc437e5f0d2465/chain/vm/vm.go#L562-L567) and [the FVM](https://github.com/filecoin-project/ref-fvm/blob/0801fd8f523692c0d17cf8c41fae2143bcff332a/fvm/src/call_manager/default.rs#L167).

I think that makes most (all?) of the problems you're describing go away.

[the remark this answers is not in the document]
raulk: I was in doubt and it sounded strange that it didn't, but I wasn't able to find where it was happening. And that happens through the layering in the state tree: https://github.com/filecoin-project/ref-fvm/blob/0801fd8f523692c0d17cf8c41fae2143bcff332a/fvm/src/call_manager/default.rs#L160

shared/src/event/mod.rs
Stebalien: > How much of a problem is that?

It breaks tooling:

- If the user linked to data, pinning a receipt wouldn't pin any data linked-to by an event.
- Selectors wouldn't be able to see inside events.

> I'd argue that we have no interest in processing Entry values at all, beyond validating that they're correct DAG-CBOR.

Either we care or we don't.

- If we care, it should be an embedded CBOR object. E.g., if we ever want to maybe consume events on-chain.
- If we don't, we can say that the user _should_ use CBOR, but I see no reason to spend extra time validating (because it doesn't actually _matter_ to us).

Given this discussion, I'm leaning towards the second.

Stebalien: > Given this discussion, I'm leaning towards the second.

Which is basically what we have, just with an extra "we're never going to validate this" guarantee.

fvm/src/executor/default.rs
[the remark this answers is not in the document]
raulk: https://github.com/filecoin-project/ref-fvm/pull/1049/commits/3a8d2e5a2a0d25270296f35c7eede00e555bccda

fvm/src/gas/price_list.rs
```diff
@@ -275,6 +281,150 @@ lazy_static! {
             memory_expansion_per_byte_cost: Zero::zero(),
         },
 
+        event_emit_base_cost: Zero::zero(),
+        event_per_entry_cost: Zero::zero(),
+        event_entry_index_cost: Zero::zero(),
+        event_per_byte_cost: Zero::zero(),
+    };
+
+    static ref HYGGE_PRICES: PriceList = PriceList {
```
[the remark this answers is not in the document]
raulk: This didn't work. I'd like us to find a safer way other than redeclaring via copy/paste, but now is not the time.

fvm/src/syscalls/event.rs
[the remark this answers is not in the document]
Stebalien: We would be charging for compute.

To be clear, I'm not suggesting that we change the charging model. We're _already_ charging for all the operations in 2-4. However, a user can get us to do a lot of free work before we get around to charging because we have to first _parse_ the structure entirely before we can determine how much it should cost.

I'm just trying to find a way to charge _as_ we parse, so we can abort early.

But this isn't even consensus critical, so we can punt till later.

fvm/src/call_manager/default.rs
```diff
@@ -593,63 +593,36 @@ where
...
-/// throw away any events collected from subcalls (and previously merged as those subcalls returned
+/// throw away any events collected from subcalls (and previously merged, as those subcalls returned
 /// normally).
-///
-/// For things to operate normally, this struct must be constructed with an empty vector as the
-/// first layer. This is where final retained events will eventually be folded into. The Default
-/// implementation constructs this object correctly.
-pub struct EventsAccumulator(Vec<Vec<StampedEvent>>);
+#[derive(Default)]
+pub struct EventsAccumulator(Vec<StampedEvent>, Vec<usize>);
```
Stebalien, line 599: uber nit: just give these names

fvm/src/call_manager/default.rs
```diff
@@ -568,3 +589,40 @@ where
...
+    }
+
+    fn create_layer(&mut self) {
+        self.1.push(self.0.len());
+    }
+
+    fn merge_last_layer(&mut self) {
+        self.1
+            .pop()
+            .expect("no index in the event accumulator when calling merge_last_layer");
```
Stebalien, line 613: I'd prefer to return fatal errors here (what we do in the state-tree). Panics tend to poison the entire machine.

Stebalien (approved):
Two nits, but everything else can be done in followups. We might as well merge.

shared/src/event/mod.rs
raulk: Hm, how about we table this discussion until I get around to updating the FIP? Then we can discuss with a larger audience there.

In short, I still think it's beneficial to enforce DAG-CBOR as these events are meant for external observability, and not having a consistent data format hinders such observability. I do not think we should **reason** _at all_ about these payloads. No link discovery, no reachability analysis, nothing. Validating structural correctness is where I'd draw the line in terms of FVM responsibility.

If we don't reason about these, I'd argue it's not a requirement to use types that are transparent to IPLD tooling (e.g. `Ipld::Ipld`, I don't know about `IgnoredAny`) in the implementation. We may opt to do so for technical reasons, though, not arguing that. But bearing in mind that that would require more type sophistication here (I can sense your eyes light up 😆).

fvm/src/machine/mod.rs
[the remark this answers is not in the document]
Stebalien: I'm strongly against it. In general, the FVM should _just_ be the VM. It manages the state-tree, applies messages, and returns receipts. It doesn't care about the receipt HAMT, tipsets, the message lists, etc.

1. Different chains (e.g., in IPC) may want to use the FVM but may also have different chain structures and indexing requirements. E.g., I expect some chains will use cryptographic accumulators, vector commitments, etc.
2. Filecoin clients will likely want to index events in some form of database _anyways_, so it's not like they won't have to process them.

testing/integration/tests/fil-events-actor/src/actor.rs
```diff
@@ -116,7 +116,9 @@ pub fn invoke(params: u32) -> u32 {
 
             if counter > 0 {
                 let params = fvm_ipld_encoding::to_vec(&counter).expect("failed to serialize");
-                sdk::send::send(&our_addr, EMIT_SUBCALLS_REVERT, params.into(), Zero::zero());
+                let _ =
+                    sdk::send::send(&our_addr, EMIT_SUBCALLS_REVERT, params.into(), Zero::zero())
+                        .ok();
```
Stebalien, line 121: You don't need the `ok()` (but I'm not sure why you're not just calling `unwrap` to assert that this worked.

raulk: The Result here is deliberately unused, and clippy was complaining. One of these sends is going to fail (see below), and ignoring that failure is part of the test scenario. I can document it.

fvm/src/call_manager/default.rs
raulk: Converted all to errors except for `finish()`. The usage site expects infallibility. I'll just drop the assert in that case.

fvm/src/call_manager/default.rs
raulk: Ok.

testing/integration/tests/fil-events-actor/src/actor.rs
Stebalien: Why clippy, WHY!!!

Issue #839, opened 2022-09-03
logging: replace `debug::enabled` with `debug::log_level`


raulk:
The only remaining part is exposing the log level. I'd like to replace `debug::enabled` with `debug::log_level`.

_Originally posted by @Stebalien in https://github.com/filecoin-project/ref-fvm/issues/173#issuecomment-1069404726_

PR #1050, opened 2022-11-07
feat: set max log when init logger

hunjixin:
set max level in init_logging


Stebalien:
We definitely don't want to set that by default. See #839 for the right fix.

Stebalien:
Wait, no, I misread. You're only doing this when debugging is enabled. Sorry about that.

Stebalien (approved):

Stebalien:
Thanks!

Issue #1051, opened 2022-11-07
GAS: Determine which syscalls need (re-)pricing


aakoshh:
Similar effort to https://github.com/filecoin-project/ref-fvm/pull/465

Capture the time-to-gas ratio of syscalls, to facilitate an analysis as in https://arxiv.org/pdf/1905.00553v1.pdf

Another, more deterministic approach is explained in https://github.com/near/nearcore/blob/78ae3b1696b308e25a61a6e896e484886829837c/runtime/runtime-params-estimator/emu-cost/README.md I'm not sure how IO is accounted for there.

Stebalien:
The syscalls that definitely don't need to be repriced are mostly miner/crypto related ones:

- crypto::verify_seal
- crypto::verify_post
- crypto::compute_unsealed_sector_cid
- crypto::verify_consensus_fault
- crypto::verify_aggregate_seals
- crypto::verify_replica_update
- crypto::batch_verify_seals

aakoshh:
I went through the syscalls :eyes: 

Here are some rough notes I took: 
[syscall-notes.txt](https://github.com/filecoin-project/ref-fvm/files/10004629/syscall-notes.txt)


Things I thought might be worth looking into:
* `vm::abort`: As noted by @Stebalien it allows the user to set an arbitrarily long `message` which `DefaultCallManager::send_resolved` will put in the `backtrace`, which finally is copied into the `ApplyRet` in `DefaultExecutor::finish_message`. Maybe worth limiting the size of that message, or make them pay for it.
* `ipld::block_stat`: This one is notable because seemingly all it does is find a block in the `BlockRegistry` and return its size, charging a fixed fee for it. Other methods that also access the block registry don't have a similar fixed fee. 
* `ipld::block_read`: It charges in proportion to the amount of data read, but shouldn't it charge a fixed fee as well for the registry lookup. like `ipld::block_stat`?
* `ipld::block_link`: It charges in proportion to the amount of data written to the store, but shouldn't it charge a fixed fee as well for the registry lookup. like `ipld::block_stat`?
[...]

PR #1054, opened 2022-11-07
MEM-1051: syscall (re-)pricing stats

aakoshh:
Resolves https://github.com/filecoin-project/ref-fvm/issues/1051

The PR aims to put in place some machinery to collect time statistics for things we charge gas for. The goal is to be able to do an analysis as described in https://arxiv.org/pdf/1905.00553v1.pdf , notably to establish a _Time-to-Gas ratio_ for syscalls, where possible.

The PR does so by returning a new `GasTimer` from the `charge_gas` method available on the `CallManager` and the `Kernel` which the caller must use either by ignoring it (where measuring time doesn't seem to make sense) or by stopping it at a point where the scope of the elapsed time is germane to the gas charged. The `GasTimer` writes the measured duration into a new field on `GasCharge`, from where it can be collected later for analysis as part of the execution trace.

### DIFF
--- a/fvm/src/gas/charge.rs
+++ b/fvm/src/gas/charge.rs
@@ -2,6 +2,7 @@
 // SPDX-License-Identifier: Apache-2.0, MIT
 
 use std::borrow::Cow;
+use std::time::Duration;
 
 use super::Gas;
 
@@ -14,6 +15,8 @@ pub struct GasCharge {
     pub compute_gas: Gas,
     /// Storage costs
     pub storage_gas: Gas,
+    /// Execution time related to this charge, if measured.
+    pub elapsed: Option<Duration>,
 }
 
 impl GasCharge {
@@ -23,6 +26,7 @@ impl GasCharge {
             name,
             compute_gas,
             storage_gas,
+            elapsed: None,
         }
     }
 
--- a/fvm/src/gas/charge.rs
+++ b/fvm/src/gas/charge.rs
@@ -2,7 +2,10 @@
 // SPDX-License-Identifier: Apache-2.0, MIT
 
 use std::borrow::Cow;
+use std::sync::Arc;
 use std::time::Duration;
+
+use once_cell::sync::OnceCell;
 
 use super::Gas;
 
@@ -15,8 +18,8 @@ pub struct GasCharge {
     pub compute_gas: Gas,
     /// Storage costs
     pub storage_gas: Gas,
-    /// Execution time related to this charge, if measured.
-    pub elapsed: Option<Duration>,
+    /// Execution time related to this charge, if successfully measured.
+    pub elapsed: Arc<OnceCell<Duration>>,
 }
 
 impl GasCharge {
@@ -26,7 +29,7 @@ impl GasCharge {
             name,
             compute_gas,
             storage_gas,
-            elapsed: None,
+            elapsed: Default::default(),
         }
     }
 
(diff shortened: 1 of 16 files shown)


Stebalien:
@aakoshh If you need test vectors from mainnet, tvx is now fixed so we can work on extracting some skyr vectors.

aakoshh:
Are there examples somewhere of how the existing test vectors were extracted? I see an [command](https://github.com/filecoin-project/lotus/blob/master/cmd/tvx/main.go) to extract a single message by ID, or multiple messages from a CSV, but I don't know how one would obtain such a list in the first place and how they'd name the test vector.

raulk:
Hey, sorry for the delay. Here are some examples.

## Single vector

Extracts the specified message, which is known to be in the specified block (providing a block is optional, but prevents walking the chain and is much faster).

```sh
./tvx extract --cid bafy2bzacebjoc6hctsurbmneuq6fqbflduhnev7xbhpjxugxbtt5e463wvomy --block bafy2bzaced4lzugaptbe3y6o6ui3ho5vx7tgcnuiadh6c2g6hhkhblhs2gzkg --out vector.json
```

Most of the time this will do the right thing, which includes running precursor messages in the block to derive the exact precondition state _before_ the selected message was applied.

I believe it will first try an optimistic approach, where only messages in the inclusion tipset from the same sender and/or to the same recipient are applied first. If after applying the desired message the receipt does not match, it will fall back to a more conservative approach of applying **all** precursors.

(This is what the --precursor-select flag does, it switches between the available strategies for precursor selection).

## Batch extraction

See the docs at https://github.com/filecoin-project/test-vectors/blob/master/corpus/extracted/README.md.

Every extracted corpus has a README.md which details how the corpus was extracted, and also the csv file that was used. For example: https://github.com/filecoin-project/test-vectors/tree/master/corpus/extracted/0004-coverage-boost

aakoshh:
Thanks @raulk that is useful indeed. I got confused, how could I have missed all these readmes, and the batch numbers didn't look familiar either. I see now; https://github.com/filecoin-project/fvm-test-vectors doesn't have this info and I never checked out the original it came from. 

The [SQL query](https://github.com/filecoin-project/test-vectors/tree/master/corpus/extracted/0004-coverage-boost#sql-query-against-chainwatch-db) is super useful for provenance.

The docs say this:
> To maximise traceability and reproduceability, scripts and technical assets are welcome, e.g. Python scripts against the JSON-RPC API, or SQL queries against chainwatch/sentinel databases.

Are you aware of any script like that, and what instance they should connect to? Where would one access a Chainwatch DB instance to extract new IDs? How do I know what heights to use for V8 actors?

Maybe these are available somewhere in the more general description on https://github.com/filecoin-project/test-vectors but any further pointers on what to run are welcome :pray:

aakoshh:
@Stebalien I tried to pick commits after the merge of #1139 into master, but there have been too many changes on master and it won't compile on this branch. If you agree with what's happening in this PR then it would be easier for me to merge this here and then open another PR that takes it to master, so it's preserved, even if there would be no similar traces there currently.

Stebalien:
Hm. This is one of those times where I miss go's `defer`.

But yes, I agree this is the best approach overall.

fvm/src/gas/charge.rs
```diff
@@ -14,6 +18,8 @@ pub struct GasCharge {
     pub compute_gas: Gas,
     /// Storage costs
     pub storage_gas: Gas,
+    /// Execution time related to this charge, if successfully measured.
+    pub elapsed: Arc<OnceCell<Duration>>,
```
Stebalien, line 22: I'd prefer to hide the details here if possible (and maybe avoid the atomic op on drop?). E.g.:

```rust
pub struct GasDuration(Option<Arc<OnceCell<Duration>>>);

impl GasDuration {
    pub fn get(&self) -> Option<Duration> {
        self.0.as_ref().and_then(|d|d.get())
    }
}
```

PR #1058, opened 2022-11-08
MEM-851: Memory expansion gas (to feat)

aakoshh:
This is a rollup of all `MEM-*` requests that were individually a bit fragmented to review. 

The PR is still opened to a feature branch because eventually to go to master we'll have to take just the gas related code, not the conformance tests which won't work on master.


fvm/src/syscalls/mod.rs
```diff
@@ -41,6 +43,12 @@ pub struct InvocationData<K> {
     /// `last_milligas_available`.
     pub last_milligas_available: i64,
 
+    /// The total size of the memory used by the execution at the beginning of this call.
+    pub start_memory_bytes: usize,
+
+    /// The total size of the memory used by the execution the last time we charged gas for it.
+    pub last_memory_bytes: usize,
```
Stebalien, line 50: We need to charge on use. After thinking through this, I'm pretty sure the _only_ way to do this will be to charge through instrumentation.

Stebalien: Basically:

1. Right now, we charge after the fact. A user could rack up more gas charges than they can afford.
2. If we tried to charge inside the limiter, we'd need to a reference to some form of reference-counted gas counter, which I'd rather not have.
3. If we charge inside instrumentation (before every call to "reserve"), we'd be able to use the current wasm gas counter.

Stebalien:
I think we can make the limiter an implementation detail of the call stack if we fix https://github.com/filecoin-project/ref-fvm/pull/1058#discussion_r1017409282

1. We only need to use the limiter inside of `DefaultCallManager::send_resolved`.
2. We don't need to expose the limiter from the kernel, because we no longer use it when we enter/exit wasm.

fvm/src/syscalls/mod.rs
aakoshh: Have you seen my comments [here](https://github.com/filecoin-project/ref-fvm/pull/1060#discussion_r1017708956) and [here](https://github.com/filecoin-project/ref-fvm/pull/1039#issuecomment-1307025078)? I posit that we are doing things exactly the way you describe in 3). 

The user cannot rack up more charges than they can afford because [this injection](https://github.com/filecoin-project/fvm-wasm-instrument/blob/master/src/gas_metering/mod.rs#L469) will make it [call the gas function before growing](https://github.com/filecoin-project/fvm-wasm-instrument/blob/master/src/gas_metering/mod.rs#L504-L505).

aakoshh:
> I think we can make the limiter an implementation detail of the CallStackManager if we fix <use instrumentation>
> We only need to use the limiter inside of DefaultCallManager::send_resolved.
> We don't need to expose the limiter from the kernel, because we no longer use it when we enter/exit wasm.

I don't think that's true. 

As I understand you are suggesting that if we use instrumentation (which we do) to charge for gas when the memory is growing, then we don't need to [pass a limiter](https://github.com/filecoin-project/ref-fvm/pull/1058/files#diff-f6ca4a1c384bf9a1f911827c3974f3f2b25357d937ce1d529732c09404ed6c92R425) to the `Store`, which is the reason the limiter is accessible through the `Kernel`. We don't use it when we enter/exit as much as we use it from inside Wasm, by the virtue of the `Store` [asking](https://github.com/bytecodealliance/wasmtime/blob/main/crates/wasmtime/src/store.rs#L1813) if it's okay to grow. Once again at this point the usage you after is purely for tracing and can be removed.

The problem is that without the limiter keeping tab on each memory grow during execution, it would be impossible to enforce rules such as not using more than a maximum amount at any point in time. Say we want nobody to use more than 100MB in any Wasm instance and 1GB at any point, but the user has gas for 10GB. With the limiter, they can spend it on multiple recursive calls that each go up to 1GB and back to the minimum (e.g. 10*100MB on the stack); without the limiter they can just keep going right up to 10GB by doing a 100*100MB deep recursion.

fvm/src/syscalls/mod.rs
Stebalien: Ah, I'm sorry, there are just too many threads to keep track of here.

You're right, we're already charging. I hadn't realized that that code had already landed in the instrumentation logic. I saw you doing memory things in `charge_for_exec` and got confused.

Stebalien:
Ah... I saw `store.limiter(...)` and figured we could call that from _after_ creating the store (from the CallManger). But we can't because we're actually passing a closure that takes the kernel and borrows the limiter from it.

Stebalien (approved):

Issue #1055, opened 2022-11-07
Handle initial memory in memory limits/charges


Stebalien:
Ref: https://webassembly.github.io/spec/core/syntax/modules.html#memories

We need to look at the memory's _minimum_ before we construct an instance and charge for the memory we're about to allocate. This will likely involve charging gas inside of https://github.com/filecoin-project/ref-fvm/blob/cf5ba35f12c74cf425ed71ded568e506290e1db2/fvm/src/machine/engine.rs#L363

Stebalien:
Was fixed in #1060

PR #1060, opened 2022-11-08
MEM-1055: Charge gas for initializing memory

aakoshh:
Fixes https://github.com/filecoin-project/ref-fvm/issues/1055

The PR adds a `charge_for_init` function to the `syscall` package (similar to `charge_for_exec`) which is called from `Engine::get_instance` to charge gas for the minimum amount of memory the module will require to start. 

This simplified `charge_for_exec` as well because it doesn't have to worry about the discount we give for the first page any more, since that is never accounted for in the wasm gas. I verified this by running a test and logging calls to `ExecResourceLimiter::memory_growing`:

```
MEMORY GROWING current=0 desired=1179648 maximum=Some(4294967296)
MEMORY GROWING current=1179648 desired=1245184 maximum=Some(536870912)
MEMORY GROWING current=1245184 desired=1310720 maximum=Some(536870912)
```
The first call is coming from the `Memory::new_static` in wasmtime, passing the default limits of a 32bit wasm memory by the looks of it, and then the second is with the default limit we set in our config, 512MiB. If the wasm module asked to grow from 0 to the minimum, we would see two calls with current=0. Therefore it won't charge gas via instrumentation.

### DIFF
--- a/fvm/src/syscalls/mod.rs
+++ b/fvm/src/syscalls/mod.rs
@@ -116,16 +116,17 @@ pub fn charge_for_exec<K: Kernel>(
 /// only charges for growing the memory _beyond_ the initial amount. It's up to us to make sure
 /// the minimum memory is properly charged for.
 pub fn charge_for_init<K: Kernel>(
-    store: &mut wasmtime::Store<InvocationData<K>>,
+    ctx: &mut impl AsContextMut<Data = InvocationData<K>>,
     module: &Module,
 ) -> crate::kernel::Result<()> {
     let min_memory_bytes = min_memory_bytes(module)?;
-    let kernel = &mut store.data_mut().kernel;
+    let mut ctx = ctx.as_context_mut();
+    let kernel = &mut ctx.data_mut().kernel;
     let charge = kernel.price_list().init_memory_gas(min_memory_bytes);
     kernel.charge_gas("wasm_memory_init", charge)?;
 
     // Adjust `last_memory_bytes` so that we don't charge for it again in `charge_for_exec`.
-    store.data_mut().last_memory_bytes += min_memory_bytes;
+    ctx.data_mut().last_memory_bytes += min_memory_bytes;
 
     Ok(())
 }
--- a/fvm/src/syscalls/mod.rs
+++ b/fvm/src/syscalls/mod.rs
@@ -2,10 +2,11 @@ use std::mem;
 
 use anyhow::{anyhow, Context as _};
 use num_traits::Zero;
-use wasmtime::{AsContextMut, Global, Linker, Memory, Val};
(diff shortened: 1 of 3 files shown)


fvm/src/syscalls/mod.rs
```diff
@@ -91,42 +89,62 @@ pub fn charge_for_exec<K: Kernel>(
...
+    module: &Module,
+) -> crate::kernel::Result<()> {
+    let min_memory_bytes = min_memory_bytes(module)?;
+    let mut ctx = ctx.as_context_mut();
+    let kernel = &mut ctx.data_mut().kernel;
+    let charge = kernel.price_list().init_memory_gas(min_memory_bytes);
+    kernel.charge_gas("wasm_memory_init", charge)?;
+
+    // Adjust `last_memory_bytes` so that we don't charge for it again in `charge_for_exec`.
+    ctx.data_mut().last_memory_bytes += min_memory_bytes;
```
Stebalien, line 129: Switching over to instrumentation will let us get rid of this.

Stebalien (commented):
This generally looks like the right approach.

fvm/src/syscalls/mod.rs
aakoshh: I still don't understand. We _are_ using instrumentation because of [this rule](https://github.com/filecoin-project/ref-fvm/blob/d00a3ea4cebbd50e579d613f03deb0b7098202e9/fvm/src/gas/price_list.rs#L743-L750) triggering the [injection](https://github.com/filecoin-project/fvm-wasm-instrument/blob/master/src/gas_metering/mod.rs#L200) of the grow cost function. 

But my understanding is that this gas counting happens purely inside Wasm, and it's the `charge_for_exec` method which [reads the current value and compares to the last](https://github.com/filecoin-project/ref-fvm/blob/d00a3ea4cebbd50e579d613f03deb0b7098202e9/fvm/src/syscalls/mod.rs#L76-L88) to figure out what it actually has to [charge for](https://github.com/filecoin-project/ref-fvm/blob/d00a3ea4cebbd50e579d613f03deb0b7098202e9/fvm/src/syscalls/mod.rs#L100). 

Charging for memory is using the same accounting mechanism, so we have to charge for it the same way. 

We can get rid of `last_memory_bytes` even now if we don't want to have a separate `wasm_exec` and `wasm_memory_grow` item in the tracer, because that is the only reason it exists.

aakoshh: Just let me know if you thing that separation is not needed, I only put it in there to get feedback about it in the PR, present it as an option.

aakoshh:
Merged so I can prepare a PR against master. We can finish discussing the instrumentation and the cost trace splitting there.

PR #1070, opened 2022-11-11
Split `InvokeContext` into two

mriise:
Splits `InvokeContext` into `NetworkContext` and `MessageContext` for clearer separation of capability domains (network context must be able to inspect network values, same with message/message context).

`vm` module is where i kept the message context as it doesnt really fit anywhere else and I didnt want to have it in it's own module.

!!! a test was disabled till I can merge a fix into builtin actors for dependencies of `vm::context` !!!


fvm/src/syscalls/network.rs
```diff
@@ -1,5 +1,6 @@
 use anyhow::Context as _;
-use fvm_shared::sys;
+use fvm_shared::sys::out::network::SyscallNetworkContext;
+use fvm_shared::sys::{self};
```
Stebalien, line 3: ?

shared/src/sys/out.rs
```diff
@@ -57,30 +57,37 @@ pub mod crypto {
...
-    use crate::clock::ChainEpoch;
     use crate::sys::TokenAmount;
     use crate::{ActorID, MethodNum};
 
     #[derive(Debug, Copy, Clone)]
     #[repr(packed, C)]
-    pub struct InvocationContext {
-        /// The value that was received.
-        pub value_received: TokenAmount,
+    pub struct SyscallMessageContext {
```
Stebalien, line 65:
```suggestion
    pub struct MessageContext {
```

?

testing/integration/tests/address_test.rs
```diff
@@ -11,8 +11,9 @@ use fvm_shared::state::StateTreeVersion;
 use fvm_shared::version::NetworkVersion;
 use num_traits::Zero;
 
-#[test]
-fn basic_address_tests() {
+// XXX RE-ENABLE THIS AFTER MERGE WITH BUILTIN ACTORS
+// #[test]
+fn _basic_address_tests() {
```
Stebalien, line 16: `#[cfg(ignore)]`?

sdk/src/vm.rs
```diff
@@ -1,17 +1,17 @@
...
+use fvm_shared::sys::out::vm::SyscallMessageContext;
 
 use crate::sys;
 
 /// BlockID representing nil parameters or return data.
 pub const NO_DATA_BLOCK_ID: u32 = 0;
 
 lazy_static::lazy_static! {
-    pub(crate) static ref INVOCATION_CONTEXT: InvocationContext = {
+    pub(crate) static ref MESSAGE_CONTEXT: SyscallMessageContext = {
```
Stebalien, line 12: move to `message`.

Stebalien (commented):
mostly nits

shared/src/sys/out.rs
mriise: I would like to keep it consistent with `SyscallNetworkContext`

testing/integration/tests/address_test.rs
mriise: then it makes the imports all mad so... eh

shared/src/sys/out.rs
Stebalien: I mean, rename both.

testing/integration/tests/address_test.rs
Stebalien: Ah, ok, fine.

Stebalien (approved):

PR #1073, opened 2022-11-11
refactor: slightly rework wasm memory limit configuration

Stebalien:
- Avoid exporting `default_wastime_config` (we don't actually need to).
- Make it take the `EngineConfig`.
- Return an error if the requested memory size is not a multiple of the page size.


aakoshh (approved):
Great, thanks!

PR #1083, opened 2022-11-15
feat: allow constructing an AMT from an iterator or borrowed values.

Stebalien:
This works because we never have to read or deserialize when creating an AMT from scratch.

Most of the code changes are actually just re-arranging the implementation into three sections:

1. The first that has no type restrictions (useful for avoiding type bounds where we don't need them).
2. The second for the newly changed function that doesn't have the `V: DeserializeOwned` requirement.
3. The third for everything else.

### DIFF
--- a/ipld/amt/src/amt.rs
+++ b/ipld/amt/src/amt.rs
@@ -95,10 +95,21 @@ where
     BS: Blockstore,
     V: Serialize,
 {
-    /// Generates an AMT with block store and array of cbor marshallable objects and returns Cid
+    /// Generates an AMT from an array of serializable objects.
     ///
     /// This can be called with an iterator of _references_ to values to avoid copying.
     pub fn new_from_iter(block_store: BS, vals: impl IntoIterator<Item = V>) -> Result<Cid, Error> {
+        Self::new_from_iter_with_bit_width(block_store, DEFAULT_BIT_WIDTH, vals)
+    }
+
+    /// Generates an AMT with the requested bitwidth from an array of serializable objects.
+    ///
+    /// This can be called with an iterator of _references_ to values to avoid copying.
+    pub fn new_from_iter_with_bit_width(
+        block_store: BS,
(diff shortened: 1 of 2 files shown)


ipld/amt/src/amt.rs
```diff
@@ -103,15 +87,60 @@ where
...
+        #[serde(transparent)]
+        struct FakeDeserialize<V>(V);
+
+        impl<'de, V> Deserialize<'de> for FakeDeserialize<V> {
+            fn deserialize<D>(_: D) -> Result<Self, D::Error>
+            where
+                D: fvm_ipld_encoding::serde_bytes::Deserializer<'de>,
+            {
+                use serde::de::Error;
+                Err(D::Error::custom("can't deserialize"))
```
mriise, line 112: "FakeDeserialize can't be deserialized"

mriise (approved):
This ended up a lot cleaner than I expected!

ipld/amt/src/amt.rs
```diff
@@ -103,15 +87,60 @@ where
...
+impl<V, BS, Ver> AmtImpl<V, BS, Ver>
+where
+    Ver: AmtVersion,
+    BS: Blockstore,
+    V: Serialize,
+{
     /// Generates an AMT with block store and array of cbor marshallable objects and returns Cid
+    ///
+    /// This can be called with an iterator of _references_ to values to avoid copying.
     pub fn new_from_iter(block_store: BS, vals: impl IntoIterator<Item = V>) -> Result<Cid, Error> {
```
raulk, line 101: Need the bitwidth here; that's another reason I couldn't use this method. Because event HAMTs are not mutable, I think it makes sense to optimise for lower depth, but curious about your opinion. I think a bitwidth of 5 or 8 makes the most sense here?

Stebalien: Depends on the max serialized size of an event, really.

PR #1088, opened 2022-11-15
EVM-892: HAMT min_data_depth option

aakoshh:
This PR contains some refactorings on the HAMT tests and the `min_data_depth` option from https://github.com/filecoin-project/ref-fvm/pull/935 which is the only one identified as a common good that we want to keep independent of the EVM footprint optimisation, which we'll carry on with a separate data structure.

### DIFF
--- a/ipld/hamt/src/node.rs
+++ b/ipld/hamt/src/node.rs
@@ -260,7 +260,7 @@ where
 
         // No existing values at this point.
         if !self.bitfield.test_bit(idx) {
-            if conf.min_data_depth <= depth {
+            if depth >= conf.min_data_depth {
                 self.insert_child(idx, key, value);
             } else {
                 // Need to insert some empty nodes reserved for links.
--- a/ipld/hamt/src/node.rs
+++ b/ipld/hamt/src/node.rs
@@ -95,6 +95,7 @@ where
         self.modify_value(
             &mut HashBits::new(&hash),
             conf,
+            0,
             key,
             value,
             store,
@@ -129,7 +130,7 @@ where
         S: Blockstore,
     {
         let hash = H::hash(k);
-        self.rm_value(&mut HashBits::new(&hash), conf, k, store)
+        self.rm_value(&mut HashBits::new(&hash), conf, 0, k, store)
     }
 
     pub fn is_empty(&self) -> bool {
@@ -246,6 +247,7 @@ where
         &mut self,
         hashed_key: &mut HashBits,
         conf: &Config,
+        depth: u32,
         key: K,
         value: V,
         store: &S,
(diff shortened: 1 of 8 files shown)


Stebalien:
cc @anorth and @ZenGround0 this is a useful optimization that lets us keep the top few nodes in the HAMT empty of values to reduce churn. Thoughts?

Stebalien:
Note: the feature is _optional_, but the builtin-actors will be primary users of this optimization.

aakoshh:
@anorth they would be a bit awkward to separate at this point because of the roundabout way it has been added. 

The original PR for adding the minimum depth is quite small: https://github.com/filecoin-project/ref-fvm/pull/935 
but it was built on another feature which has been removed in the 3rd commit in this PR, leaving in all the other refactoring that allow adding more configuration to the HAMT, which the tests rely use. For example I moved the bucket size into the config in https://github.com/filecoin-project/ref-fvm/pull/1094 

There are no tests specific to this option, only property based and fuzz tests ensure that the invariants of the HAMT are maintained. The effect of having the top levels contain less data was tested indirectly in https://github.com/filecoin-project/builtin-actors/pull/740

I can try to undo this feature as well, have a PR with just test refactoring, and then undo the undo to add the option back.

aakoshh:
Done, https://github.com/filecoin-project/ref-fvm/pull/1123 only has the test refactoring and this PR is now agains that branch.

Stebalien:
@aakoshh this can be rebased/merged with master now.

ipld/hamt/src/node.rs
```diff
@@ -257,7 +259,14 @@ where
 
         // No existing values at this point.
         if !self.bitfield.test_bit(idx) {
-            self.insert_child(idx, key, value);
+            if conf.min_data_depth <= depth {
```
anorth, line 262: Nit: this would be more obviously correct to me if the test were written the other way round

anorth (commented):
This looks good, but I wish there were tests to demonstrate it was actually doing something to the structure.

I'll defer to @Stebalien for approval

ipld/hamt/src/node.rs
aakoshh: Interesting, thanks for sharing your preference. I have that same feeling that I want the variable to be on the left, but I also like to order GT and LT relations as if they were aligned on an axis, like `x < y && y < z`. 

I can reverse it, no problem.

aakoshh:
> I wish there were tests to demonstrate it was actually doing something to the structure.

@anorth agreed, I added a test to demonstrate there is an effect on the root size.

Stebalien (approved):

PR #1092, opened 2022-11-16
refactor: move base-fee & timestamp into the network context

Stebalien:
Now that we have a network context, there's no need to have separate syscalls for these. We've kept a separate syscall for the circulating supply, because we'd like to restrict that to special actors...

Also, rename network_curr_epoch to just epoch.

### DIFF
--- a/fvm/src/syscalls/network.rs
+++ b/fvm/src/syscalls/network.rs
@@ -5,16 +5,6 @@ use fvm_shared::sys::out::network::NetworkContext as SyscallNetworkContext;
 use super::Context;
 use crate::kernel::{ClassifyResult, Kernel, Result};
 
-/// Returns the base fee split as two u64 ordered in little endian.
-pub fn base_fee(context: Context<'_, impl Kernel>) -> Result<sys::TokenAmount> {
-    context
-        .kernel
-        .network_base_fee()
-        .try_into()
-        .context("base-fee exceeds u128 limit")
-        .or_fatal()
-}
-
 /// Returns the network circ supply split as two u64 ordered in little endian.
 pub fn total_fil_circ_supply(context: Context<'_, impl Kernel>) -> Result<sys::TokenAmount> {
     context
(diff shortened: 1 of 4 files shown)


Stebalien:
(should make validation a bit simpler)

Stebalien:
Er, abstract accounts.

fvm/src/syscalls/network.rs
```diff
@@ -27,15 +17,18 @@ pub fn total_fil_circ_supply(context: Context<'_, impl Kernel>) -> Result<sys::T
 
 pub fn context(context: Context<'_, impl Kernel>) -> crate::kernel::Result<SyscallNetworkContext> {
     Ok(SyscallNetworkContext {
-        network_curr_epoch: context.kernel.network_epoch(),
```
Stebalien, line 30: the function name is only called `network_epoch` because it's on the `NetworkOps` trait.

shared/src/sys/out.rs
```diff
@@ -82,12 +82,17 @@ pub mod vm {
...
     use crate::clock::ChainEpoch;
+    use crate::sys::TokenAmount;
 
     #[derive(Debug, Copy, Clone)]
     #[repr(packed, C)]
     pub struct NetworkContext {
         /// The current epoch.
-        pub network_curr_epoch: ChainEpoch,
+        pub epoch: ChainEpoch,
+        /// The current time (seconds since the unix epoch).
```
mriise, line 92: this currently is hardwired no?

mriise (approved):
yay more context things!

shared/src/sys/out.rs
Stebalien: For now, yes.

mriise:
huh.

shared/src/sys/out.rs
Stebalien: Ah, no, it's not hardwired. I need to fix that todo.

vyzo: the ts is wires in the fevm lotus branch, nothing to fix.

raulk (approved):
LGTM. But let's hold off from merging until we fix #1096 and #1093 and make a release for Topaz 🙏

Stebalien:
Yep, that;s the plan.

PR #1094, opened 2022-11-16
EVM-1085: Kamt

aakoshh:
Working on a data structure that has extension/skip support like the one I added to the `Hamt` but isn't the `Hamt`. 

Not happy about the massive amount of code duplication :cry:


ipld/kamt/src/bitfield.rs
```diff
@@ -0,0 +1,212 @@
+// Copyright 2019-2022 ChainSafe Systems
```
vyzo, line 1: ??? did chainsafe write this code?

ipld/kamt/fuzz/fuzz_targets/simple.rs
```diff
@@ -0,0 +1,18 @@
+// Copyright 2019-2022 ChainSafe Systems
```
vyzo, line 1: ???

ipld/kamt/fuzz/fuzz_targets/common.rs
```diff
@@ -0,0 +1,50 @@
+// Copyright 2019-2022 ChainSafe Systems
```
vyzo, line 1: ???

ipld/kamt/benches/kamt_benchmark.rs
```diff
@@ -0,0 +1,120 @@
+// Copyright 2019-2022 ChainSafe Systems
```
vyzo, line 1: ????

ipld/kamt/src/bitfield.rs
```diff
@@ -0,0 +1,212 @@
...
+
+use std::u64;
+
+use byteorder::{BigEndian, ByteOrder};
+use fvm_ipld_encoding::de::{Deserialize, Deserializer};
+use fvm_ipld_encoding::ser::{Serialize, Serializer};
+use fvm_ipld_encoding::strict_bytes;
+
+#[derive(Debug, Clone, PartialEq, Eq, Copy)]
+pub struct Bitfield([u64; 4]);
```
vyzo, line 12: Is this copied from somewhere?

vyzo: I mean, is it different from the hamt bitfield? If so, can we add some comments as to how? If not, can't we reuse the original?

ipld/kamt/src/error.rs
```diff
@@ -0,0 +1,57 @@
+// Copyright 2019-2022 ChainSafe Systems
```
vyzo, line 1: ???

ipld/kamt/src/ext.rs
```diff
@@ -0,0 +1,294 @@
+// Copyright 2019-2022 ChainSafe Systems
```
vyzo, line 1: ???

ipld/kamt/src/hash_bits.rs
```diff
@@ -0,0 +1,139 @@
+// Copyright 2019-2022 ChainSafe Systems
```
vyzo, line 1: ???

ipld/kamt/src/kamt.rs
```diff
@@ -0,0 +1,354 @@
+// Copyright 2019-2022 ChainSafe Systems
```
vyzo, line 1: ???

ipld/kamt/src/lib.rs
```diff
@@ -0,0 +1,97 @@
+// Copyright 2019-2022 ChainSafe Systems
```
vyzo, line 1: ???

ipld/kamt/src/node.rs
```diff
@@ -0,0 +1,697 @@
+// Copyright 2019-2022 ChainSafe Systems
```
vyzo, line 1: ???

ipld/kamt/src/pointer.rs
```diff
@@ -0,0 +1,296 @@
+// Copyright 2019-2022 ChainSafe Systems
```
vyzo, line 1: ???

vyzo (commented):
Just a first cursory pass to take in the lay of land; this is a lot of code, out of context for me, and quite subtle as a bug can be catastrophic.
It will take me a little while to review.

But first things first.... Is this new code? Is it pilfering chainsafe code? Why are you assigning copyright of Protocol Labs code to chainsafe?

ipld/kamt/benches/kamt_benchmark.rs
aakoshh: Copy of https://github.com/filecoin-project/ref-fvm/blob/master/ipld/hamt/benches/hamt_benchmark.rs
Similar one in https://github.com/filecoin-project/ref-fvm/blob/master/ipld/amt/benches/amt_benchmark.rs

ipld/kamt/fuzz/fuzz_targets/common.rs
aakoshh: Copy of https://github.com/filecoin-project/ref-fvm/blob/master/ipld/hamt/fuzz/fuzz_targets/common.rs

I created that file, but only as a refactoring of code that was already in `simple.rs` :man_shrugging:

ipld/kamt/src/bitfield.rs
aakoshh: This question is one of the TODO items I left there. I wanted to hear what @Stebalien wants in this case because the HAMT bitfield is private to the crate. Do we want to make it public and couple these two?

aakoshh: https://github.com/filecoin-project/ref-fvm/pull/1094/files#diff-6758c2ada424248b816b085b04035a3ee8fe7903fa44e237e11b26f99d6d9d4aR89 is new

aakoshh:
As I commented in the PR description, this code is 90% same as the Hamt and is a copy of it at the stage where I added the extra features that helped the EVM use case. I linked to those PRs in the description. The license headers were copied with the code. 

I see that in the project in general licenses are not maintained in the sense that many (new?) files don't have them, and modified ones don't update the header. 

If you want this situation fixed, I suggest using an automation [similar to Forest](https://github.com/ChainSafe/forest/blob/main/Makefile#L129) that checks that each modified file in the commit has a header that contains Protocol Labs, and replaces the header with one that contains both PL and ChainSafe if there already is a header, or adds one with only PL if there is no header.

ipld/kamt/src/bitfield.rs
aakoshh: This is easier to review in https://github.com/filecoin-project/ref-fvm/pull/930/files#diff-1b21066d1c876de01d801288f2064722e144b24ba2b715736307fea4362855e4

Stebalien (approved):
I fixed a couple of nits, otherwise lgtm.

Stebalien:
> Maybe I have tunnel vision, but I'm not sure if having to maintain the HAMT and the KAMT is better than just living with it in the HAMT.

I'd generally prefer to avoid the duplication, but I'm just concerned about making extensive changes to the HAMT itself at this point.

Let's revisit this after we ship M2.2. This is a reservable decision (we can likely just _replace_ the HAMT with the KAMT in many cases going forward).

ipld/kamt/src/pointer.rs
```diff
@@ -0,0 +1,297 @@
...
+fn unsplit_ext(
+    conf: &Config,
+    bf: &Bitfield,
+    parent_ext: &Extension,
+    child_ext: &Extension,
+) -> Result<Extension, Error> {
+    // Figure out which bucket contains the pointer.
+    let idx = bf
+        .last_one_idx()
+        .expect("There is supposed to be exactly one pointer") as u8;
```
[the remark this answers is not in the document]
aakoshh: I mean this method should only ever be called if there's just one pointer in one bucket. It would be a programming error to do otherwise. If there were multiple pointers, then we could not unsplit a node. 

We want to turn `A -> B -> C` into `A -> C`, but if we have `A -> B -> [C, D]`, with `B` pointing at `C` and `D`, then it's okay to leave it like that.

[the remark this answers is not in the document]
aakoshh: So basically, this helper should not be called outside [this](https://github.com/filecoin-project/ref-fvm/pull/1094/files/20a3edacedafa5895f3de39d1bdb8f40a32e4969#diff-7b99b75f2612ac6832d2903486f74ce43fe930def46d807e728b44d698c57e17R174) branch.

ipld/kamt/src/bitfield.rs
```diff
@@ -0,0 +1,213 @@
...
+        self.0.iter().map(|a| a.count_ones() as usize).sum()
+    }
+
+    pub fn last_one_idx(&self) -> Option<usize> {
+        [192, 128, 64, 0]
+            .iter()
+            .enumerate()
+            .find_map(|(i, offset)| match self.0[3 - i] {
+                0 => None,
+                n => Some((n as f64).log2().floor() as usize + offset),
```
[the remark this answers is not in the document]
aakoshh: You are right, this is ugly. I just wanted to avoid having to loop through all of the indexes to find a non-zero value, I was hoping the logarithm might be faster. I didn't benchmark it, though. 

It doesn't need to be precise since I'm just looking at the highest non-zero bit, and throw away the parts where the conversion from `u64` to `f64` loses precision.

Do let me know if you have a better idea :pray:

PR #1095, opened 2022-11-16
Revert "FIX: HashBits::next when bit_width does not divide 256 and the full hash is consumed"

vyzo:
Reverts filecoin-project/ref-fvm#945

Suspected of causing a nasty regression, so testing with it reverted.
Do not merge before we have verified its the origin of the problem.


raulk:
This was just a red herring. The problem was: #1096.

PR #1097, opened 2022-11-16
fix: patch funty for now

Stebalien:
Upstream broke the world by yanking a version.


raulk:
See: https://github.com/bitvecto-rs/funty/issues/7

raulk (approved):

PR #1103, opened 2022-11-17
fix: allow f0 addresses to be used to generate new actor addresses

Stebalien:
This is a stop-gap to ensure that we can deploy actors when the "origin" doesn't have a predictable address.

Future work: https://github.com/filecoin-project/FIPs/discussions/388#discussioncomment-4163159


Stebalien:
Note: this _wasn't_ a problem for the system actor, because we treat the system actor's ID address as a key address (by pure luck).

vyzo (approved):

PR #1104, opened 2022-11-17
chore: update wasmtime to 2.0.2

Stebalien:


Stebalien:
Note: due to linking issues, we'll need to apply this change to v2 as well at the same time, and make sure it's backwards compatible.

PR #1105, opened 2022-11-17
chore: update wasmtime

Stebalien:
Includes some important bug fixes (that likely don't actually affect us, but still).


vyzo (approved):
needs backport to rel/v2 or pain in ffi.

PR #1106, opened 2022-11-17
fix: avoid treating some instantiation errors as fatal

### LINKED ISSUE #1102: In a deep recursive call, we hit the wasmtime memory limit before our memory limit
Specifically, we're failing with:

> [FATAL] Error: [from=f3qzzyjciy4dqg7b45ajfur6jhwb32wq5fovnu2ydaj4jqpihznjxj3ejujs5maa6sr43rpqlueuw5dhvx3xfa, to=f01001, seq=2, m=2, h=19]: Insufficient resources: memory minimum size of 17 pages exceeds memory limits, Backtrace:

After 455 recursive calls in the multisig in the multisig itest (lotus).

https://app.circleci.com/pipelines/github/filecoin-project/lotus/24424/workflows/d35000d5-1a5c-4314-8e28-ebd401fb16c0/jobs/676915/parallel-runs/0/steps/0-114

This means two things:

1. ~I don't think we actually need to handle initial memory, because it looks like wasmtime does this for us.~
2. We need to _not_ treat this as a fatal error.

Stebalien:
Specifically, OOM and TRAPs during instantiation are _not_ fatal.

fixes #1102

### DIFF
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -432,11 +432,33 @@ where
 
             // From this point on, there are no more syscall errors, only aborts.
             let result: std::result::Result<BlockId, Abort> = (|| {
+                use wasmtime_runtime::InstantiationError;
                 // Instantiate the module.
                 let instance = engine
                     .get_instance(&mut store, &state.code)
                     .and_then(|i| i.context("actor code not found"))
-                    .map_err(Abort::Fatal)?;
+                    .map_err(|e| match e.downcast::<InstantiationError>() {
+                        Ok(e) => match e {
+                            // This will be handled in validation.
+                            InstantiationError::Link(e) => Abort::Fatal(anyhow!(e)),
+                            // TODO: We may want a separate OOM exit code? However, normal ooms will usually exit with SYS_ILLEGAL_INSTRUCTION.
+                            InstantiationError::Resource(e) => {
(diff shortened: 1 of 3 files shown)


Stebalien:
(this definitely needs a test)

fvm/src/call_manager/default.rs
```diff
@@ -432,11 +432,33 @@ where
...
                 // Instantiate the module.
                 let instance = engine
                     .get_instance(&mut store, &state.code)
                     .and_then(|i| i.context("actor code not found"))
-                    .map_err(Abort::Fatal)?;
+                    .map_err(|e| match e.downcast::<InstantiationError>() {
+                        Ok(e) => match e {
+                            // This will be handled in validation.
+                            InstantiationError::Link(e) => Abort::Fatal(anyhow!(e)),
+                            // TODO: We may want a separate OOM exit code? However, normal ooms will usually exit with SYS_ILLEGAL_INSTRUCTION.
```
vyzo, line 444: precedent in UNIX says we probably dont.
Would it help anyone?

We should definitely log this loudly however.

vyzo (approved):

fvm/src/call_manager/default.rs
Stebalien: >  Would it help anyone?

It may help with some debugging. Let's just leave these TODOs here and move on for now.

Issue #1116, opened 2022-11-17
FIP: Actor creation changes


Stebalien:
1. Replace new_actor_address with next_actor_address https://github.com/filecoin-project/ref-fvm/issues/941.
2. Use the origin as written in the message when creating new f2 addresses (don't resolve anything). This is necessary for account abstraction as abstract accounts don't have key addresses.

(maybe some more changes?)
(that second change may get bundled with the account abstraction fip)

Stebalien:
> Use the origin as written in the message when creating new f2 addresses (don't resolve anything). This is necessary for account abstraction as abstract accounts don't have key addresses.

Will be a part of the EVM FIP.

> Replace new_actor_address with next_actor_address Replace the new_actor_address with a next_actor_address syscall #941.

Already done.

PR #1117, opened 2022-11-17
feat: refactor actor creation logic

### LINKED ISSUE #941: Replace the new_actor_address with a next_actor_address syscall
We've discussed two proposals in the past:

1. Combining the `create_actor` and the `new_actor_address` syscalls.
2. Converting the `Constructor` method to a `constructor` wasm endpoint. 

Unfortunately, these two proposals are incompatible as we need to register actor's stable address in the init actor _before_ calling the constructor.

So instead I'd like to propose the following: change `new_actor_address` to `next_actor_address` where `next_actor_address` returns the stable address of the next deployed actor, but doesn't actually update the number of actors created. Instead, we'd leave that to `create_actor`.

Concretely: increment `num_actors_created` inside `create_actor` instead of inside `next_actor_address`.

Stebalien:
1. Use the origin as specified in the top-level message without trying to resolve it. This is necessary for account abstraction because, per-finality, we should use the _stable_ address (f1-f3) to compute new `f2` addresses. Unfortunately, we have no way to lookup an actor's "stable" (`f2`) address. Unfortunately, we can't just use the actor's `f4` address because not all abstract accounts will have one (thinking beyond the EVM here).
2. Replace the `new_actor_address` syscall with `next_actor_address`. We now update the "num actors created" _after_ creating the next actor (see https://github.com/filecoin-project/ref-fvm/issues/941).

All these changes are a bit intertwined because I didn't want to expose a new `origin_address` method on the CallManager, only to remove it later.

fixes #941

### DIFF
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -1,4 +1,5 @@
 use anyhow::{anyhow, Context};
+use cid::Cid;
 use derive_more::{Deref, DerefMut};
 use fvm_ipld_encoding::{to_vec, RawBytes, DAG_CBOR};
 use fvm_shared::address::{Address, Payload};
@@ -38,6 +39,8 @@ pub struct InnerDefaultCallManager<M: Machine> {
     /// The ActorID and the address of the original sender of the chain message that initiated
     /// this call stack.
     origin: ActorID,
+    /// The origin address as specified in the message (used to derive new f2 addresses).
+    origin_address: Address,
     /// The nonce of the chain message that initiated this call stack.
     nonce: u64,
     /// Number of actors created in this call stack.
@@ -82,6 +85,7 @@ where
         machine: M,
         gas_limit: i64,
         origin: ActorID,
+        origin_address: Address,
         nonce: u64,
         gas_premium: TokenAmount,
     ) -> Self {
@@ -96,6 +100,7 @@ where
             machine,
             gas_tracker,
             origin,
+            origin_address,
             nonce,
             num_actors_created: 0,
             call_stack_depth: 0,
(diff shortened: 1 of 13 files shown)


Stebalien:
So we don't forget the FIP: https://github.com/filecoin-project/ref-fvm/issues/1116

fvm/src/call_manager/default.rs
```diff
@@ -242,21 +247,83 @@ where
...
+        // an f2 address (before finality).
+        let mut b = to_vec(&self.origin_address).expect("failed to serialize address");
+        b.extend_from_slice(&self.nonce.to_be_bytes());
+        b.extend_from_slice(&self.num_actors_created.to_be_bytes());
+        Address::new_actor(&b)
     }
 
-    fn invocation_count(&self) -> u64 {
-        self.invocation_count
+    fn create_actor(
```
Stebalien, line 267: I moved this here because:

1. It doesn't actually care about the kernel.
2. We'll eventually need to change this to do an internal "send" to actually construct the actor (which is in the domain of the CallManager anyways).

Stebalien: Note: this hasn't changed, just moved.

fvm/src/call_manager/default.rs
```diff
@@ -242,21 +247,83 @@ where
...
+        // e.g., an f2 address even if we can't look it up anywhere.
+        //
+        // Of course, if the user decides to send from an f0 address without waiting for finality,
+        // their "stable" address may not be as stable as they'd like. But that's their problem.
+        //
+        // In case you're wondering: but what if someone _else_ is relying on the stability of this
+        // address? They shouldn't be. The sender can always _replace_ a message with a new message,
+        // and completely change how f2 addresses are assigned. Only the message sender can rely on
+        // an f2 address (before finality).
+        let mut b = to_vec(&self.origin_address).expect("failed to serialize address");
```
Stebalien, line 261: This is the original motivation for this PR. It's the correct fix for https://github.com/filecoin-project/ref-fvm/pull/1103.

arajasek (approved):
This looks good, we definitely need to publicize this in case tooling / users are affected.

I think you can drop the `origin` field and its getter from the call manager, I'd be surprised if it had any other usages. I'd be in favour of not having it.

Issue #913, opened 2022-09-30
refactor `NetworkConfig` and `NetworkContext`


raulk:
There's more refactor/consolidation we'll need to do here between `NetworkConfig` and `NetworkContext`.

`NetworkConfig` is confusing because it actually contains **machine** configuration:

- max call depth
- max wasm stack
- actor debugging
- actor redirects
- builtin overrides

As well as some network attributes which should live in `NetworkContext`:

- network version
- price list

But we don't need to do it _now_.

_Originally posted by @raulk in https://github.com/filecoin-project/ref-fvm/pull/905#discussion_r984526392_

Stebalien:
`NetworkConfig` defines the current network. The network version and the price list are part of the current network config. Basically, the network config is the thing that changes on network upgrades.

Issue #933, opened 2022-10-07
Plumb tipset CIDs and timestamps through the FFI and/or externs


Stebalien:
Currently, we just hard-code them. But we can't exactly forget to fix this before we cut a release.

PR #1119, opened 2022-11-17
feat: use externs for tipset cid lookups

Stebalien:
So, @vyzo was right. This is the best way to do this.

It also means we don't need the whole network context thing.

part of https://github.com/filecoin-project/ref-fvm/issues/933
fixes https://github.com/filecoin-project/ref-fvm/issues/913 (by getting rid of the NetworkContext).

### DIFF
--- a/fvm/src/kernel/default.rs
+++ b/fvm/src/kernel/default.rs
@@ -28,7 +28,7 @@ use super::error::Result;
 use super::hash::SupportedHashes;
 use super::*;
 use crate::call_manager::{CallManager, InvocationResult, NO_DATA_BLOCK_ID};
-use crate::externs::{Consensus, Rand};
+use crate::externs::{Chain, Consensus, Rand};
 use crate::state_tree::ActorState;
 use crate::syscall_error;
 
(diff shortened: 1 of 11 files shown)


fvm/src/kernel/default.rs
```diff
@@ -40,6 +40,7 @@ lazy_static! {
 const BLAKE2B_256: u64 = 0xb220;
 const ENV_ARTIFACT_DIR: &str = "FVM_STORE_ARTIFACT_DIR";
 const MAX_ARTIFACT_NAME_LEN: usize = 256;
+const FINALITY: i64 = 900;
```
mriise, line 43: TODO bikeshed

testing/conformance/src/vm.rs
```diff
@@ -107,7 +107,7 @@ impl TestMachine<Box<DefaultMachine<MemoryBlockstore, TestExterns>>> {
 
         let mut nc = NetworkConfig::new(network_version);
         nc.override_actors(builtin_actors);
-        let mut mc = nc.for_epoch(epoch, state_root);
+        let mut mc = nc.for_epoch(epoch, (epoch * 30) as u64, state_root);
```
mriise, line 110: timestamp = 30 * chain epoch num?

mriise (approved):

fvm/src/kernel/default.rs
Stebalien: https://github.com/filecoin-project/ref-fvm/issues/1120

testing/conformance/src/vm.rs
Stebalien: Yes. This is for testing only because we need a number. We'll probably need a better number later, but we can fix that later.

Issue #987, opened 2022-10-18
Non-"dag-cbor" IPLD


Stebalien:
# Background

Currently, we have quite a few places that simply assume that all IPLD data is dag-cbor.

In the sdk, `send` assumes `dag-cbor`:

- https://github.com/filecoin-project/ref-fvm/blob/f7b51ba8e3a779bd3e5310cb3e2305ef59075d92/sdk/src/send.rs#L29
- https://github.com/filecoin-project/ref-fvm/blob/f7b51ba8e3a779bd3e5310cb3e2305ef59075d92/sdk/src/send.rs#L38

In the builtin actors, the "trampoline" assumes `dag-cbor`:

- https://github.com/filecoin-project/builtin-actors/blob/91ff9fa5c057359174424eb59b34965b10d19e90/runtime/src/runtime/fvm.rs#L556
- https://github.com/filecoin-project/builtin-actors/blob/91ff9fa5c057359174424eb59b34965b10d19e90/runtime/src/runtime/fvm.rs#L577

On-chain, the `Message` and `Receipt` types assume `dag-cbor` (simply because they don't allow us to specify anything else):

- https://github.com/filecoin-project/ref-fvm/blob/f7b51ba8e3a779bd3e5310cb3e2305ef59075d92/shared/src/message.rs#L23
- https://github.com/filecoin-project/ref-fvm/blob/f7b51ba8e3a779bd3e5310cb3e2305ef59075d92/shared/src/receipt.rs#L13

However:

- For FEVM, we'd like to support "raw" IPLD blocks because FEVM calls use untyped byte arrays.
- In M2.2, wasm actors will be raw blocks as well (most likely, at least).
- We should probably add support for bare "cbor" (not "dag-cbor") for message parameters as message parameters generally aren't allowed to link to anything.
- The system _cares_ about the codec:
    - It needs to know the codec so it can perform reachability analysis.
    - Actors themselves need to know how to decode their parameters. Right now they can just assume "cbor", but we'd _like_ to allow raw as well to avoid having to wrap _every_ object in a bit of CBOR.

# Changes

We have three categories of changes:

- On-chain messages/receipts.
- FVM SDK/APIs.
- "Embedded" blocks.

## Messages/Receipts

Currently, on-chain messages don't have a field to specify the codec. So the simplest solution here is to just say that all on-chain message parameters must be "cbor" (not "dag-cbor", just "cbor", because message parameters can't link to anything).

[...]

PR #1121, opened 2022-11-18
Generalize Abort to Exit with data

### LINKED ISSUE #986: Change vm::abort to vm::exit
Replace `abort` with an `exit` syscall.

```rust
fn exit(code: u32,       // Exit code.
        value: u32,      // Block ID (return value).
        msg: *const u8,  // Optional message.
        msg_len: u32);
```

- If `code` is non-zero, revert.
- Always return the value where `0` identifies the "empty" value.
- The message is optional and only for tracing (it doesn't affect consensus, doesn't cost gas, and will usually just be discarded).

vyzo:
Implements generalized aborts as exits, which can carry data and also allow non-local returns (exit(0)).

Closes #986.

### DIFF
--- a/fvm/src/syscalls/error.rs
+++ b/fvm/src/syscalls/error.rs
@@ -6,13 +6,14 @@ use derive_more::Display;
 use fvm_shared::error::ExitCode;
 use wasmtime::Trap;
 
-use crate::kernel::ExecutionError;
+use crate::kernel::{BlockId, ExecutionError};
+use crate::call_manager::{NO_DATA_BLOCK_ID};
 
 /// Represents an actor "abort".
 #[derive(Debug)]
 pub enum Abort {
     /// The actor explicitly aborted with the given exit code (or panicked).
-    Exit(ExitCode, String),
+    Exit(ExitCode, String, BlockId),
     /// The actor ran out of gas.
     OutOfGas,
     /// The system failed with a fatal error.
@@ -30,6 +31,7 @@ impl Abort {
                     "actor aborted with an invalid message: {} (code={:?})",
                     e.0, e.1
                 ),
+                0,
             ),
             ExecutionError::OutOfGas => Abort::OutOfGas,
             ExecutionError::Fatal(err) => Abort::Fatal(err),
@@ -43,6 +45,11 @@ impl Abort {
             ExecutionError::Fatal(e) => Abort::Fatal(e),
             ExecutionError::Syscall(e) => Abort::Fatal(anyhow!("unexpected syscall error: {}", e)),
         }
+    }
+
+    /// Convert a generalized exit "jump" into an "abort".
+    pub fn from_exit(code: ExitCode, msg: String, blk: BlockId) -> Self {
+        Abort::Exit(code, msg, blk)
     }
 }
 
@@ -60,7 +67,7 @@ impl From<Trap> for Abort {
 
         // Actor panic/wasm error.
         if let Some(code) = t.trap_code() {
-            return Abort::Exit(ExitCode::SYS_ILLEGAL_INSTRUCTION, code.to_string());
+            return Abort::Exit(ExitCode::SYS_ILLEGAL_INSTRUCTION, code.to_string(), NO_DATA_BLOCK_ID);
         }
 
         // Try to get a smuggled error back.
--- a/fvm/src/syscalls/error.rs
+++ b/fvm/src/syscalls/error.rs
@@ -6,8 +6,8 @@ use derive_more::Display;
 use fvm_shared::error::ExitCode;
 use wasmtime::Trap;
 
+use crate::call_manager::NO_DATA_BLOCK_ID;
 use crate::kernel::{BlockId, ExecutionError};
-use crate::call_manager::{NO_DATA_BLOCK_ID};
 
 /// Represents an actor "abort".
 #[derive(Debug)]
@@ -67,7 +67,11 @@ impl From<Trap> for Abort {
 
         // Actor panic/wasm error.
         if let Some(code) = t.trap_code() {
-            return Abort::Exit(ExitCode::SYS_ILLEGAL_INSTRUCTION, code.to_string(), NO_DATA_BLOCK_ID);
+            return Abort::Exit(
+                ExitCode::SYS_ILLEGAL_INSTRUCTION,
+                code.to_string(),
+                NO_DATA_BLOCK_ID,
+            );
         }
 
         // Try to get a smuggled error back.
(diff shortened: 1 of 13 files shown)


Stebalien:
Approach LGTM. Now we just need to fix the exit syscall.

vyzo:
The audit CI job is barfing because of the rogue dependency -- haven't we fixed that?

Stebalien:
> The audit CI job is barfing because of the rogue dependency -- haven't we fixed that?

Not for that one CI job. For everything else, we're relying on the lockfile.

fvm/src/syscalls/error.rs
```diff
@@ -44,6 +46,11 @@ impl Abort {
             ExecutionError::Syscall(e) => Abort::Fatal(anyhow!("unexpected syscall error: {}", e)),
         }
     }
+
+    /// Convert a generalized exit "jump" into an "abort".
+    pub fn from_exit(code: ExitCode, msg: String, blk: BlockId) -> Self {
+        Abort::Exit(code, msg, blk)
+    }
```
Stebalien, line 53: This isn't really necessary. I'd just call `Abort::Exit(code, msg, blk)` directly.

fvm/src/syscalls/vm.rs
```diff
@@ -14,11 +14,12 @@ pub enum Never {}
 unsafe impl SyscallSafe for Never {}
 
 // NOTE: this won't clobber the last syscall error because it directly returns a "trap".
-pub fn abort(
+pub fn exit(
     context: Context<'_, impl Kernel>,
     code: u32,
     message_off: u32,
     message_len: u32,
+    blk: u32,
```
Stebalien, line 22: nit: let's switch the ordering. It's cleaner if the optional error message goes last.

sdk/src/sys/vm.rs
```diff
@@ -6,23 +6,27 @@ pub use fvm_shared::sys::out::vm::MessageContext;
 super::fvm_syscalls! {
     module = "vm";
 
-    /// Abort execution with the given code and message. The code is recorded in the receipt, the
-    /// message is for debugging only.
+    /// Abort execution with the given code and optional message and data.
+    /// The code is recorded in the receipt, the message is for debugging only.
```
Stebalien, line 10: The code and return value are recorded in the receipt.

sdk/src/vm.rs
```diff
@@ -9,14 +10,31 @@ pub const NO_DATA_BLOCK_ID: u32 = 0;
 
 /// Abort execution.
 pub fn abort(code: u32, message: Option<&str>) -> ! {
+    exit(code, message, None)
```
Stebalien, line 13: Check that the code is non-zero, and convert it into something else if it is (assertion failed?).

sdk/src/vm.rs
```diff
@@ -9,14 +10,31 @@ pub const NO_DATA_BLOCK_ID: u32 = 0;
...
+}
+
+/// Exit from current message execution, with the specified code and an optional message and data.
+pub fn exit(code: u32, message: Option<&str>, data: Option<RawBytes>) -> ! {
```
Stebalien, line 17: We can probably just pass `RawBytes`, no need for the option here.

We always use an empty "raw bytes" to indicate "nothing" (because that's not valid cbor anyways).

sdk/src/vm.rs
```diff
@@ -9,14 +10,31 @@ pub const NO_DATA_BLOCK_ID: u32 = 0;
...
+        // but the compiler doexn't accept it.
+        let blk_id = if let Some(bytes) = data {
+            if bytes.len() > 0 {
+                sys::ipld::block_create(DAG_CBOR, bytes.as_ptr(), bytes.len() as u32).unwrap()
+            } else {
+                NO_DATA_BLOCK_ID
+            }
+        } else {
+            NO_DATA_BLOCK_ID
+        };
```
Stebalien, line 35: FYI

```suggestion
        let blk_id = match data {
            Some(bytes) if !bytes.is_empty() => {
                sys::ipld::block_create(DAG_CBOR, bytes.as_ptr(), bytes.len() as u32).unwrap()
            },
            _ => NO_DATA_BLOCK_ID,
        };
```

Stebalien:
Note: the integration test failure looks real.

sdk/src/vm.rs
Stebalien: E.g., see `send`.

(all this will change with #987 anyways)

fvm/src/syscalls/vm.rs
vyzo: ok.

sdk/src/sys/vm.rs
vyzo: ack.

sdk/src/vm.rs
vyzo: Ok, fair enough, that's a rust thing :)

sdk/src/vm.rs
vyzo: ah, if let doesn't work but there is always plan match; ok, good.
Too bad it will be written to not use the option :)

vyzo:
It's because the bundles try to use vm::abort syscall which is gone...
It is very real, an artifact of the circular dependency. 
This can't be fixed until we have build new bundles that use the new syscall, so the only way forward is to define it again until we can remove it... sigh.

sdk/src/vm.rs
Stebalien: Oh, no, this isn't a rust thing. This is just a legacy FVM thing that needs to be fixed.

sdk/src/vm.rs
Stebalien: Yep. I just wanted to make sure you've seen match guards.

fvm/src/syscalls/error.rs
vyzo: ok, will drop.

Stebalien:
Ah... it's because our replaced deps don't affect downstream. They usually _should_, but our bundling logic downstream _intentionally_ ignores that because I was an idiot at the time (and lazy).

Ok, I guess we're doing a dance.

fvm/src/syscalls/error.rs
vyzo: removed.

sdk/src/sys/vm.rs
vyzo: fixed

sdk/src/vm.rs
vyzo: done.

fvm/src/syscalls/vm.rs
vyzo: done.

Stebalien (approved):

PR #1122, opened 2022-11-18
refactor: merge "context" kernel methods

Stebalien:
Remove the per-item functions and collapse them into single "context" functions. This will let us:

1. Enforce restrictions necessary for account abstraction.
2. Correctly charge gas.

Also: move the gas premium into the call manager and out of the gas tracker. The gas tracker is for tracking gas available/spent and doesn't care about price.

### DIFF
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -36,6 +36,8 @@ pub struct InnerDefaultCallManager<M: Machine> {
     machine: M,
     /// The gas tracker.
     gas_tracker: GasTracker,
+    /// The gas premium paid by this message.
+    gas_premium: TokenAmount,
     /// The ActorID and the address of the original sender of the chain message that initiated
     /// this call stack.
     origin: ActorID,
@@ -90,7 +92,7 @@ where
         gas_premium: TokenAmount,
     ) -> Self {
         let limits = machine.new_limiter();
-        let mut gas_tracker = GasTracker::new(Gas::new(gas_limit), Gas::zero(), gas_premium);
+        let mut gas_tracker = GasTracker::new(Gas::new(gas_limit), Gas::zero());
 
         if machine.context().tracing {
             gas_tracker.enable_tracing()
@@ -99,6 +101,7 @@ where
         DefaultCallManager(Some(Box::new(InnerDefaultCallManager {
             machine,
             gas_tracker,
+            gas_premium,
             origin,
             origin_address,
             nonce,
@@ -235,6 +238,10 @@ where
 
     fn gas_tracker_mut(&mut self) -> &mut GasTracker {
         &mut self.gas_tracker
+    }
+
+    fn gas_premium(&self) -> &TokenAmount {
+        &self.gas_premium
     }
 
     // Other accessor methods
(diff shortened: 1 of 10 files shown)


fvm/src/kernel/default.rs
```diff
@@ -616,20 +609,26 @@ impl<C> NetworkOps for DefaultKernel<C>
...
+            ..
+        } = self.call_manager.context();
+        Ok(NetworkContext {
+            epoch: *epoch,
+            network_version: *network_version as u32,
+            timestamp: *timestamp,
+            base_fee: base_fee
+                .try_into()
+                .or_fatal()
+                .context("base-fee exceeds u128 limit")?,
```
Stebalien, line 630: I'd _prefer_ to do this conversion in the syscall itself, but then we'd need to define _two_ different `NetworkContext` types. It just isn't worth it.

fvm/src/kernel/default.rs
```diff
@@ -304,32 +304,25 @@ impl<C> MessageOps for DefaultKernel<C>
...
-
-    fn msg_gas_premium(&self) -> TokenAmount {
-        self.call_manager.gas_tracker().gas_premium()
-    }
-
-    fn msg_gas_limit(&self) -> u64 {
-        self.call_manager.gas_tracker().gas_limit().round_down() as u64
+    fn msg_context(&self) -> Result<MessageContext> {
+        // TODO: charge gas.
+        Ok(MessageContext {
```
Stebalien, line 309: We could consider creating this structure once on initialization, but that may not be worth it.

Stebalien:
cc @aakoshh

Stebalien:
> Enforce restrictions necessary for account abstraction.

Although... after doing this, I think we may be able to punt on that particular part of account abstraction for now.

fvm/src/call_manager/default.rs
```diff
@@ -36,6 +36,8 @@ pub struct InnerDefaultCallManager<M: Machine> {
     machine: M,
     /// The gas tracker.
     gas_tracker: GasTracker,
+    /// The gas premium paid by this message.
+    gas_premium: TokenAmount,
```
vyzo, line 40: Maybe put this in the gas tracker?
Looks kinda out of place.

vyzo: oh it was and you moved it out.
Any particular reason?

fvm/src/kernel/default.rs
vyzo: Depends on how much it costs!

vyzo: We could cache in first access and reuse and reduce gas charge for subsequent accesses.

fvm/src/kernel/default.rs
vyzo: Similar caching considerations here.

fvm/src/call_manager/default.rs
vyzo: Note i am fine with it, i shall blame raul's insistence to hide it on how it ended there in the first place (git will blame me).

vyzo (approved):
Overall LGTM, just some caching considerations.

fvm/src/call_manager/default.rs
Stebalien: Yeah, basically, it has nothing to do with gas tracking. It's used for gas _charging_ (which is why the gas tracker never even looked at it). Same with the base fee.

fvm/src/kernel/default.rs
Stebalien: Yep, you're right. Profile first.

Issue #859, opened 2022-09-09
Optimize EVM runtime storage footprint


raulk:
Currently, the built-in EVM actor is using a HAMT as the physical data structure to represent Ethereum smart contract storage. We know this is an inefficient data structure to map, store, and query highly atomized/granular U256 => U256. This epich is about:

1. Modeling different Ethereum native workloads to benchmark the HAMT in terms of gas performance (obtain a baseline).
2. Investigating and proposing alternative data structures that would lead to more efficient patterns of access and writing.
3. Discussing and agreeing on one or two approaches to prototype.
4. Prototyping and benchmarking them.
5. Taking the implementation of the selected approach to completion.

raulk:
For point 1, I think the first thing we'll want to do is write several Solidity contracts that are write and read intensive in terms of data volume and dispersion of keys. We can use that suite as a benchmark harness, and to establish an initial baseline.

We can also add real typical Ethereum contracts into the benchmarking suite — that way we can make sure we are not overfitting to read/write intensive scenarios.

raulk:
@Stebalien — could you review this issue for correctness of approach, and provide thoughts and guidance to guide @aakoshh's work?

Stebalien:
This looks correct. To extend it a bit:

- We should fix https://github.com/filecoin-project/ref-fvm/issues/636 first.
- We should test the HAMT with different parameters. HAMTs may actually work, but we'll probably need to tune the width and bucket size at a minimum.

Beyond that, we should try:

- Optimizing for sequential storage (e.g., with some form of sparse array).
- Optimizing the storage of small(er) numbers. We don't _have_ to treat U256 values as byte arrays. We can treat them as integers when small.
- Optimize the storage of keys. Currently, our HAMTs aren't "proper" tries and store the full keys in the leaves. Given our small values, we may want to optimize that.

Stebalien:
As much as it pains me, we may also want to consider a custom "packed" format for the leaves of our merkletree given our highly regular data.

raulk:
^^ Solidity calculates storage slots by hashing, so unfortunately optimizing for sequential keys will probably not help much: https://docs.soliditylang.org/en/v0.8.13/internals/layout_in_storage.html

Stebalien:
I'm actually not so sure:

> Except for dynamically-sized arrays and mappings ... data is stored contiguously item after item starting with the first state variable, which is stored in slot 0

In terms of dynamically sized arrays/maps:

[...]

PR #1123, opened 2022-11-18
EVM-892: Hamt test and config refactoring

aakoshh:
The PR adds a `Config` to the `hamt` which only contains the `bit_width` but can be extended with further options. Also refactors the unit tests and fuzzing for reuse based on these config. 

The PR has been a cutting back of options added during the EVM footprint effort under https://github.com/filecoin-project/ref-fvm/issues/859


ipld/hamt/src/hash_bits.rs
```diff
@@ -5,15 +5,18 @@ use std::cmp::Ordering;
 
 use crate::{Error, HashedKey};
 
+/// Part of a `HashedKey`.
+pub type HashedKeySlice = [u8];
```
Stebalien, line 9: Can we keep this private?

ipld/hamt/src/node.rs
```diff
@@ -302,16 +306,18 @@ where
 
                 // If the array is full, create a subshard and insert everything
                 if vals.len() >= MAX_ARRAY_WIDTH {
-                    let mut sub = Node::<K, V, H>::default();
+                    let kvs = std::mem::take(vals);
+                    let hashes = kvs.iter().map(|kv| H::hash(kv.key())).collect::<Vec<_>>();
```
Stebalien, line 310: I _think_ I can drop the `collect` here, right?

ipld/hamt/src/node.rs
```diff
@@ -141,7 +142,7 @@ where
     {
         for p in &self.pointers {
             match p {
-                Pointer::Link { cid, cache } => {
+                Pointer::Link { cid, cache, .. } => {
```
Stebalien, line 145: leftover?

Stebalien (approved):
Couple of nits, otherwise lgtm! Thanks!

ipld/hamt/src/node.rs
Stebalien: (fixed because I want to merge before cutting a release)

ipld/hamt/src/hash_bits.rs
Stebalien: Also fixed.

PR #1129, opened 2022-11-18
feat: replace wat2wasm with the pure-rust wat crate

Stebalien:
`wat` is built and maintained by the bytecode alliance, and is pure rust. Importantly, it doesn't invoke cmake.


vyzo (approved):
yay!

mriise:
wat?

PR #1133, opened 2022-11-19
feat: limit the size of backtrace messages to 1k

Stebalien:
This limits the overall size of a backtrace to ~1MiB of user data.

### DIFF
--- a/fvm/src/syscalls/vm.rs
+++ b/fvm/src/syscalls/vm.rs
@@ -13,6 +13,10 @@ pub enum Never {}
 
 unsafe impl SyscallSafe for Never {}
 
+/// The maximum message length included in the backtrace. Given 1025 levels, this gives us a total
+/// maximum of around 1MiB for debugging.
+const MAX_MESSAGE_LEN: usize = 1024;
+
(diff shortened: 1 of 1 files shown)


vyzo (approved):

fvm/src/syscalls/vm.rs
```diff
@@ -34,7 +38,20 @@ pub fn exit(
...
         match context.memory.try_slice(message_off, message_len) {
-            Ok(bytes) => String::from_utf8_lossy(bytes).into_owned(),
+            Ok(bytes) => {
+                if bytes.len() > MAX_MESSAGE_LEN {
+                    let prefix = &bytes[..(MAX_MESSAGE_LEN / 2)];
+                    let suffix = &bytes[bytes.len() - (MAX_MESSAGE_LEN / 2)..];
+                    format!(
+                        "{} ... (skipped {} bytes) ... {}",
+                        String::from_utf8_lossy(prefix),
+                        MAX_MESSAGE_LEN - bytes.len(),
```
aakoshh, line 48: Shouldn't this be the other way around? 

Say `bytes.len()` is 2345, then it will say it skipped (1024-2345) = -1321 bytes, with a negative sign.

Issue #1075, opened 2022-11-11
Charge for table growing


Stebalien:
We don't curently charge for table growing, but we should. According to the wasmtime docs, each table element requires one usize of memory (host memory).

Proposal: Charge 8 bytes per table element.

1. Instrument and charge for zeroing in wasm.
2. Account for the expanded table memory in wasm.

aakoshh:
Just looking at where this could be implemented.

When an instance is allocated by the pooling allocator, it already [gets the maximum number of table elements](https://github.com/bytecodealliance/wasmtime/blob/v2.0.1/crates/runtime/src/instance/allocator/pooling.rs#L484-L488) writeable to it. The table will be static; it will [start from the minimum](https://github.com/bytecodealliance/wasmtime/blob/v2.0.1/crates/runtime/src/table.rs#L196) and account for growth, but this for a static table [is just a formality](https://github.com/bytecodealliance/wasmtime/blob/v2.0.1/crates/runtime/src/table.rs#L360-L364) of checking the limits. 

I'm not sure at the moment how to figure out the minimum table size for the module. Wasmtime doesn't expose the memory plans, and I resorted to calling `inst.get_memory("memory")`, which was an established well known name. I don't see us using `inst.get_table` at the moment. The instance limits only allow 1 table per module by default, so I'm assuming this is the function table (because all elements in it have to have the same type). Does that have some well-known name?

Something I don't understand is that in wasmtime we can see [table grow code being translated](https://github.com/bytecodealliance/wasmtime/blob/v2.0.1/cranelift/wasm/src/code_translator.rs#L1352-L1364), but when I follow [how we instrument memory growth](https://github.com/filecoin-project/fvm-wasm-instrument/blob/master/src/gas_metering/mod.rs#L468), there is [no such thing](https://github.com/paritytech/parity-wasm/blob/master/src/elements/ops.rs#L650) as a grow table instruction. At the moment I'm not sure where to inject any charge for growing the table. Also not sure where it would grow if it's only for functions - can more of them be added during runtime?

Stebalien:
Status: We now charge for table init, but not for growing.

We now need to charge for _growing_, but should do this when instrumenting instructions.

PR #1138, opened 2022-11-21
MEM-1075: Charge for table init

aakoshh:
Addresses https://github.com/filecoin-project/ref-fvm/issues/1075

The PR changes `charge_for_init` to also charge 8 bytes worth of memory expansion for the minimum number of table elements. 

See notes in the issue for why I don't think we can charge for growing the table right now. Here's the TL;DR version:

We are using `parity-wasm` to parse Wasm modules and inject our code e.g. for gas accounting, which then we write back into a modified Wasm module (binary) that Wasmtime compiles to native code. The `parity-wasm` project is essentially deprecated, archived, read-only - it's not being updated with newer Wasm 2.0 features. The `table-grow` instruction is newer than `parity-wasm` and would not be recognised by the parser if it was part of a smart contract code. Therefore we can't inject code that charges for it either. We'd either have to bring `parity-wasm` up to speed with newer op codes, or find a different library to use.

All we can do at the moment is allow modules to state their minimum number of table elements and charge for those.


fvm/src/gas/price_list.rs
```diff
@@ -860,6 +860,15 @@ impl PriceList {
         self.wasm_rules.memory_expansion_per_byte_cost * grow_memory_bytes as i64
     }
 
+    /// Returns the gas required for initializing tables.
+    pub fn init_table_gas(&self, min_table_elements: u32) -> Gas {
+        // Each element reserves a `usize` in the table, so we charge 8 bytes per pointer.
+        // https://docs.rs/wasmtime/2.0.2/wasmtime/struct.InstanceLimits.html#structfield.table_elements
+        self.wasm_rules.memory_expansion_per_byte_cost
+            * (min_table_elements as i32)
```
Stebalien, line 868: Let's deal with overflow here.

fvm/src/gas/price_list.rs
```diff
@@ -860,6 +860,15 @@ impl PriceList {
...
+            * (core::mem::size_of::<*const ()>() as i32)
```
Stebalien, line 869: We need this charge to be architecture independent, so let's just charge for 8 bytes (64 bits).

fvm/src/syscalls/mod.rs
```diff
@@ -149,6 +154,19 @@ fn min_memory_bytes(module: &Module) -> crate::kernel::Result<usize> {
     }
 }
 
+/// Get the minimum number of table elements a module will use.
+///
+/// This relies on a few assumptions:
+///     * That we use the default value for `InstanceLimits::tables` and only allow 1 table.
+///     * That `Linker::command` will only allow them to be exported with the name "table".
```
Stebalien, line 161: We'll have to re-evaluate in m2.2, but this is probably good enough.

Stebalien (commented):
Given that this is a fixed cost that can't be controlled by the user and that we can't quite implement it correctly, I'd like to punt this to M2.2.

Stebalien:
I can also just merge this (once we've fixed the platform-dependent issue) but that's something we can talk about.

fvm/src/gas/price_list.rs
aakoshh: Thanks. I had the same thought but I wasn't sure if it's possible to run this with 32 bit systems.

fvm/src/gas/price_list.rs
aakoshh: Doesn't the way multiplication is implemented deal with it? 

https://github.com/filecoin-project/ref-fvm/blob/master/fvm/src/gas/mod.rs#L141

fvm/src/gas/price_list.rs
Stebalien: Well... Ok, it's not. But I'd still prefer to just hard-code this.

fvm/src/gas/price_list.rs
Stebalien: Ah, I mean wrap around when casting to an i32.

aakoshh: Oh, good catch, that could have been a nasty vulnerability :facepalm: 

I added variants of `Gas::mul` to handle these so we don't have to keep casting.

Stebalien (approved):

PR #1139, opened 2022-11-22
MEM-1051: Charge gas for more syscalls

aakoshh:
Adds gas charge for syscalls that have none, mentioned in https://github.com/filecoin-project/ref-fvm/issues/1051#issuecomment-1313971723

- [ ] `sself::root`
- [ ] `sself::set_root`
- [ ] `sself::current_balance`
- [ ] ...


fvm/src/gas/price_list.rs
```diff
@@ -966,6 +972,16 @@ impl PriceList {
         GasCharge::new("OnMessageContext", self.context_base, Zero::zero())
     }
 
+    /// Returns the gas required for accessing the message context.
```
maciejwitowski, line 975: Incorrect comment.

fvm/src/kernel/default.rs
```diff
@@ -807,15 +845,23 @@ where
 
     #[cfg(feature = "m2-native")]
     fn install_actor(&mut self, code_id: Cid) -> Result<()> {
-        // TODO figure out gas
-        self.call_manager
+        let size = self
```
Stebalien, line 848: We'll likely need something significantly more complicated, but this is a good start I guess.

Stebalien (approved):
Good to merge.

fvm/src/call_manager/default.rs
```diff
@@ -298,19 +298,16 @@ where
...
             }
             // Create a new actor.
-            None => {
-                self.charge_gas(self.price_list().on_create_actor())?;
-                ActorState::new_empty(code_id, predictable_address)
-            }
+            None => (ActorState::new_empty(code_id, predictable_address), true),
         };
-
+        self.charge_gas(self.price_list().on_create_actor(is_new))?;
```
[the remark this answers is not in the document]
aakoshh: Thanks for the suggestion. I don't really know what an embryo actor is, so I thought we're still "creating an actor" but by upgrading an embryo. If you want to reuse updating for other things that touch the state tree, that would make sense to me as well. 

There is a tension between having to call everything a different name even if their cost comes from essentially doing the same thing, like updating the actor state CID or code CID or balance are accessing the same record in the state tree, at least _now_ they do. Should they charge `OnStateTreeRead`? Or should they keep charging `OnThis` and and `OnThat` so we can attribute the cost to them, but reuse the same price underneat - this is what I'm doing now.

PR #1140, opened 2022-11-22
MEM-1051: Refactor CallManager to use internal mutability for GasTracker

aakoshh:
Part of https://github.com/filecoin-project/ref-fvm/issues/1051

## What

The PR changes `CallManager` to keep the `GasTracker` in a `RefCell`, and allow mutable access to it even via non-mutable reference to itself. 

## Why

While working on https://github.com/filecoin-project/ref-fvm/pull/1139 I realised that with the way the `GasTracker` is exposed by the `CallManager` in `gas_tracker(&self)` and `gas_tracker_mut(&mut self)`, I'd have to change all methods on the `Kernel` to take `&mut self` to be able to charge gas for them. Many of these methods are read-only and take `&self` at the moment, for example `SelfOps::current_balance`. 

I think having them take `&self` is more expressive of their nature, ie. whether they mutate the state of the ledger, and worth keeping. Taking `&mut self` just for charging gas would make this distinction between operations disappear and would be a shame IMO.

Arguably charging for gas is orthogonal to the ledger semantics, since we have to charge for _everything_ - having to have a mutable reference for charging gas seems questionable.

## How

The PR changes `CallManager` to use a `RefCell` to hold the `GasTracker`, and take closures to apply on the tracker in either immutable or mutable fashion. I chose this approach over others I tried because:
* I don't want to return a `RefMut` from `gas_tracker_mut(&self)` for example. That would be rather unsafe as we'd lose the compiler not allowing simultaneous references and might result in a runtime panic. 
* There weren't that many places actually accessing these methods, so the syntactic churn turned out to be quite limited.
* I also decided against moving the internal mutability to the `GasTracker`, because there it felt more expressive the way it was.

### DIFF
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -1,3 +1,5 @@
+use std::cell::RefCell;
+
 use anyhow::{anyhow, Context};
 use cid::Cid;
 use derive_more::{Deref, DerefMut};
@@ -35,7 +37,7 @@ pub struct InnerDefaultCallManager<M: Machine> {
     #[deref_mut]
     machine: M,
     /// The gas tracker.
-    gas_tracker: GasTracker,
+    gas_tracker: RefCell<GasTracker>,
     /// The gas premium paid by this message.
     gas_premium: TokenAmount,
     /// The ActorID and the address of the original sender of the chain message that initiated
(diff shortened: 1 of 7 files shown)


Stebalien:
I'm really not a fan of having to pass closures everywhere. I'd either:

1. Add interior mutability to the gas tracker itself. This kind of thing is pretty common in rust and you can even use a `Cell` instead of a `RefCell` to 
2. Just make everything mutable.

Here's an implementation of 1: https://github.com/filecoin-project/ref-fvm/pull/1141

aakoshh:
Fine with me, thanks for making that PR!

PR #1141, opened 2022-11-23
feat: add interior mutability to the gas tracker

Stebalien:
see https://github.com/filecoin-project/ref-fvm/pull/1140

### DIFF
--- a/fvm/src/gas/mod.rs
+++ b/fvm/src/gas/mod.rs
@@ -248,7 +248,7 @@ mod tests {
     #[test]
     #[allow(clippy::identity_op)]
     fn basic_gas_tracker() -> Result<()> {
-        let mut t = GasTracker::new(Gas::new(20), Gas::new(10));
+        let t = GasTracker::new(Gas::new(20), Gas::new(10), false);
         t.apply_charge(GasCharge::new("", Gas::new(5), Gas::zero()))?;
         assert_eq!(t.gas_used(), Gas::new(15));
         t.apply_charge(GasCharge::new("", Gas::new(5), Gas::zero()))?;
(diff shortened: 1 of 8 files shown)


fvm/src/gas/mod.rs
```diff
@@ -152,53 +153,52 @@ impl Mul<i32> for Gas {
...
-        self.trace = Some(vec![]);
-    }
-
-    fn charge_gas_inner(&mut self, name: &str, to_use: Gas) -> Result<()> {
+    fn charge_gas_inner(&self, name: &str, to_use: Gas) -> Result<()> {
         log::trace!("charging gas: {} {}", name, to_use);
         // The gas type uses saturating math.
-        self.gas_used += to_use;
-        if self.gas_used > self.gas_limit {
+        let gas_used = self.gas_used.get() + to_use;
```
Stebalien, line 174: get/set are "free".

fvm/src/gas/mod.rs
```diff
@@ -152,53 +153,52 @@ impl Mul<i32> for Gas {
 
 pub struct GasTracker {
     gas_limit: Gas,
-    gas_used: Gas,
-    trace: Option<Vec<GasCharge>>,
+    gas_used: Cell<Gas>,
+    trace: Option<RefCell<Vec<GasCharge>>>,
```
Stebalien, line 157: Putting the option on the outside means we don't have to pay for the dynamic refcell check unless we're actually tracing. Ideally we'd just use a cell, but `Cell::update` is still unstable and this isn't performance critical.

aakoshh (approved):
Looks good to me, thanks :+1:  I cherry picked the commit that makes kernel methods immutable where the only reason for mutability was to charge for gas.

PR #1146, opened 2022-11-23
feat: use actor IDs and remove some fatal errors

Stebalien:
This is in preparation for "readonly" mode. Basically, I'm:

- Changing all the actor get/set/modify methods to take actor IDs (removing any internal resolution). That way the caller has to handle all of the error cases and can chose what to do.
- Avoid marking failures in some state-tree modification functions as "fatal errors". This will let us enforce the readonly constraint at a very low-level (inside the state-tree).

Why didn't we do this in the first place? Basically, legacy code inherited from Forest (likely trying to be compatible with the lotus VM).

### DIFF
--- a/fvm/src/machine/default.rs
+++ b/fvm/src/machine/default.rs
@@ -18,7 +18,7 @@ use crate::blockstore::BufferedBlockstore;
 use crate::externs::Externs;
 #[cfg(feature = "m2-native")]
 use crate::init_actor::State as InitActorState;
-use crate::kernel::{ClassifyResult, Context as _, Result};
+use crate::kernel::{ClassifyResult, Result};
 use crate::machine::limiter::ExecResourceLimiter;
 use crate::machine::Manifest;
 use crate::state_tree::{ActorState, StateTree};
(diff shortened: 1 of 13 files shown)


fvm/src/machine/default.rs
```diff
@@ -211,15 +211,9 @@ where
...
-            .register_new_address(addr)
-            .context("failed to register new address")
-            .or_fatal()?;
+        let addr_id = state_tree.register_new_address(addr)?;
 
-        state_tree
-            .set_actor(&Address::new_id(addr_id), act)
-            .context("failed to set actor")
-            .or_fatal()?;
+        state_tree.set_actor(addr_id, act)?;
```
Stebalien, line 216: These now internally decide if the error should be fatal, and don't treat the error as fatal if, e.g., we're in "readonly" mode.

fvm/src/state_tree.rs
```diff
@@ -348,11 +339,7 @@ where
 
         let (state, _) = InitActorState::load(self)?;
 
-        let a = match state
-            .resolve_address(self.store(), addr)
-            .context("Could not resolve address")
-            .or_fatal()?
```
Stebalien, line 354: `resolve_address` already decides whether or not this should be fatal.

vyzo (approved):
Looks the right, unixy thing to do.

Issue #966, opened 2022-10-13
Allow limiting gas for internal sends


Stebalien:
Currently, there's no way to limit the amount of gas spent on internal sends. We need to add a new field to the `send` syscall to allow limiting the gas spent.

maciejwitowski:
@Stebalien I added this to [the checklist](https://github.com/filecoin-project/ref-fvm/issues/936) based on Quartz milestone assigned.

PR #1149, opened 2022-11-25
send syscall: take an optional gas limit.

raulk:
Closes https://github.com/filecoin-project/ref-fvm/issues/966.

## TODO

- [x] Actually enforce the gas limit.
- [x] Integration tests.

### DIFF
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -143,9 +143,32 @@ where
             });
         }
 
-        let result = self.with_stack_frame(|s| {
-            s.send_unchecked::<K>(from, to, method, params, value, gas_limit)
-        });
+        // If a specific gas limit has been requested, create a child GasTracker and use that
+        // one hereon.
+        let prev_gas_tracker = gas_limit
+            .and_then(|limit| self.gas_tracker.new_child(limit))
+            .map(|new| mem::replace(&mut self.gas_tracker, new));
+
+        let mut result =
+            self.with_stack_frame(|s| s.send_unchecked::<K>(from, to, method, params, value));
+
+        // Restore the original gas tracker and absorb the child's gas usage and traces into it.
+        if let Some(prev) = prev_gas_tracker {
+            let other = mem::replace(&mut self.gas_tracker, prev);
+            // This is capable of raising an OutOfGas, but it is redundant since send_resolved
+            // would've already raised it, so we ignore it here. We could check and assert that's
+            // true, but send_resolved could also error _fatally_ and mask the OutOfGas, so it's
+            // not safe to do so.
+            let _ = self.gas_tracker.absorb(&other);
+
+            // If we were limiting gas, convert the execution error to an exit.
+            if matches!(result, Err(ExecutionError::OutOfGas)) {
+                result = Ok(InvocationResult {
+                    exit_code: ExitCode::SYS_OUT_OF_GAS,
+                    value: None,
+                })
+            }
+        }
 
         if self.machine.context().tracing {
             self.trace(match &result {
@@ -409,7 +432,6 @@ where
         method: MethodNum,
         params: Option<Block>,
         value: &TokenAmount,
-        gas_limit: Option<Gas>,
     ) -> Result<InvocationResult>
     where
         K: Kernel<CallManager = Self>,
@@ -433,32 +455,7 @@ where
             },
         };
 
-        // If a specific gas limit has been requested, create a child GasTracker and use that
-        // one hereon.
-        let prev_gas_tracker = gas_limit
-            .and_then(|limit| self.gas_tracker.new_child(limit))
-            .map(|new| mem::replace(&mut self.gas_tracker, new));
-
-        let mut ret = self.send_resolved::<K>(from, to, method, params, value);
-
-        // Restore the original gas tracker and absorb the child's gas usage and traces into it.
-        if let Some(prev) = prev_gas_tracker {
-            let other = mem::replace(&mut self.gas_tracker, prev);
-            // This is capable of raising an OutOfGas, but it is redundant since send_resolved
-            // would've already raised it, so we ignore it here. We could check and assert that's
-            // true, but send_resolved could also error _fatally_ and mask the OutOfGas, so it's
-            // not safe to do so.
-            let _ = self.gas_tracker.absorb(&other);
-
-            // If we were limiting gas, convert the execution error to an exit.
-            if matches!(ret, Err(ExecutionError::OutOfGas)) {
-                ret = Ok(InvocationResult {
-                    exit_code: ExitCode::SYS_OUT_OF_GAS,
-                    value: None,
-                })
-            }
-        }
-        ret
+        self.send_resolved::<K>(from, to, method, params, value)
     }
 
     /// Send with resolved addresses.
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -432,20 +432,18 @@ where
             },
         };
 
-        // TODO should we deduct the method invocation gas from the supplied gas_limit?
-        //  Assuming yes -- this happens inside send_resolved.
-
         // If a specific gas limit has been requested, create a child GasTracker and use that
         // one hereon.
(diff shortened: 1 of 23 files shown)


raulk:
The current implementation doesn't lower out of gas errors from children to syscall error numbers (resuming execution at the caller) -- that's the next step to figure out.

fvm/src/call_manager/default.rs
```diff
@@ -427,9 +432,35 @@ where
             },
         };
 
+        // TODO should we deduct the method invocation gas from the supplied gas_limit?
+        //  Assuming yes -- this happens inside send_resolved.
```
Stebalien, line 436: Ideally, yes.

fvm/src/call_manager/default.rs
```diff
@@ -427,9 +432,35 @@ where
...
+
+        // If a specific gas limit has been requested, create a child GasTracker and use that
+        // one hereon.
+        let prev_gas_tracker = match gas_limit {
+            Some(gas_limit) if gas_limit < self.gas_tracker.gas_available() => {
+                let gas_used = self.gas_tracker.gas_used();
+                let tracing = self.machine.context().tracing;
+                let new = GasTracker::new(gas_limit, gas_used, tracing);
+                Some(mem::replace(&mut self.gas_tracker, new))
```
Stebalien, line 445: We could also consider just replacing the limit, then restoring it?

fvm/src/call_manager/default.rs
```diff
@@ -427,9 +432,35 @@ where
...
+        let ret = self.send_resolved::<K>(from, to, method, params, value);
+
+        // Restore the original gas tracker and absorb the child's gas usage and traces into it.
+        if let Some(prev) = prev_gas_tracker {
+            let other = mem::replace(&mut self.gas_tracker, prev);
+            // This is capable of raising an OutOfGas, but it is redundant since send_resolved
+            // would've already raised it, so we ignore it here. We could check and assert that's
+            // true, but send_resolved could also error _fatally_ and mask the OutOfGas, so it's
+            // not safe to do so.
+            let _ = self.gas_tracker.absorb(&other);
```
Stebalien, line 460: If we made the limit updatable, we wouldn't have to do this.

raulk: Could do, although that kind of mutation moves the accounting outside the component? Seems cleaner this way (but also less performant).

Stebalien: I think a `lower_limit` and `raise_limit` operation makes sense.

Stebalien:
The current distinction is:

- Syscall error if the send syscall fails.
- Exit code if the send syscall _succeeds_, but the target actor fails.

We do already "cook" up a new exit code in a few cases:

- `SYS_ILLEGAL_INSTRUCTION` on panic/illegal instruction.
- `SYS_ILLEGAL_EXIT_CODE` if the actor exits with a reserved exit code.
- `SYS_MISSING_RETURN` if the actor returns a non-existent block.

IMO, exiting with `SYS_OUT_OF_GAS` makes sense in this case (although it's a _bit_ funky because we might not even get to the point where the other actor runs...).

raulk:
This feature works now. There are two items raised above which haven't made it into the PR.

1. Patching the `GasTracker` instead of creating a new instance and then absorbing.
2. Deciding about exit code and error number. This patch introduces an error number, but switching to an exit code should be trivial.

I'm fine using an exit code for (2). The reason being that all other conditions equal, sending an explicit message from off chain with the same gas limit (and removing chain inclusion costs and other costs) would also raise an `SYS_OUT_OF_GAS` exit code.

Can you think of cases where returning a `SYS_OUT_OF_GAS` here would occlude actual system-generated out of gas errors? For example, if an OOG happens after the syscall ends and before we return? (e.g. result memcpy cost?)

Stebalien:
> Can you think of cases where returning a SYS_OUT_OF_GAS here would occlude actual system-generated out of gas errors? For example, if an OOG happens after the syscall ends and before we return? (e.g. result memcpy cost?)

We only convert an `ExecutionError::OutOfGas` to an exit _if_ the caller limited the gas, so we should be safe as long as we never use that field from the executor itself.

fvm/src/call_manager/default.rs
```diff
@@ -433,30 +431,17 @@ where
...
+        // limit is strictly less than the current limit).
+        let prev_gas_limit =
+            gas_limit.and_then(|gas_limit| self.gas_tracker.reduce_gas_remaining(gas_limit));
 
         // Do the actual send.
         let mut ret = self.send_resolved::<K>(from, to, method, params, value);
 
-        // Restore the original gas tracker and absorb the child's gas usage and traces into it.
-        if let Some(prev) = prev_gas_tracker {
-            let other = mem::replace(&mut self.gas_tracker, prev);
```
Stebalien: Note: this didn't actually copy the gas _used_ (one of the reasons I prefer to keep the gas tracker and just edit the limits).

fvm/src/gas/mod.rs
```diff
@@ -193,6 +194,27 @@ impl GasTracker {
         self.charge_gas_inner(to_use)
     }
 
+    /// Reduce the current the gas remaining to `sub_limit`. If there is less than `sub_limit` gas
+    /// remaining, does nothing and returns `None`. Otherwise, returns the old gas limit.
+    pub fn reduce_gas_remaining(&self, sub_limit: Gas) -> Option<Gas> {
```
Stebalien, line 199: I don't _particularly_ like this method, but it's the best I can come up with.

Stebalien: Hm. No, no, I'm not doing it this way...

fvm/src/call_manager/default.rs
Stebalien: Ah, nope. This was doing the right thing.

Stebalien:
Changes:

1. Apply the new gas limit _before_ attempting to automatically create new actors.
2. Start the new "gas used" value at 0 (bug fix).
3. Test fixes.
4. Use exit codes instead of syscall errors.
5. Handle overflow.

fvm/src/gas/mod.rs
```diff
@@ -185,22 +184,39 @@ impl GasTracker {
...
+    pub fn absorb(&self, other: &GasTracker) -> Result<()> {
+        if let Some(trace) = &self.trace {
+            trace.borrow_mut().extend(other.drain_trace());
+        }
+        self.charge_gas_inner(other.gas_used())
+    }
+
+    /// Make a "child" gas-tracker with a new limit, if and only if the new limit is less than the
+    /// available gas.
+    pub fn new_child(&self, new_limit: Gas) -> Option<GasTracker> {
```
raulk, line 217: Nice.

fvm/src/syscalls/bind.rs
```diff
@@ -199,3 +199,4 @@ impl_bind_syscalls!(A B C D);
 impl_bind_syscalls!(A B C D E);
 impl_bind_syscalls!(A B C D E F);
 impl_bind_syscalls!(A B C D E F G);
+impl_bind_syscalls!(A B C D E F G H);
```
raulk, line 202: Interesting. I don't think I had added this binding pattern here and nothing complained. Why would that be?

raulk (commented):
LGTM. I can't officially approve because I'm the PR author, but feel free to merge.

fvm/src/gas/mod.rs
```diff
@@ -185,22 +184,39 @@ impl GasTracker {
...
         if let Some(trace) = &self.trace {
             trace.borrow_mut().push(charge);
         }
-        res
+        self.charge_gas_inner(to_use)
+    }
+
+    /// Absorbs another GasTracker (usually a nested one) into this one, charging for gas
+    /// used and appending all traces.
+    pub fn absorb(&self, other: &GasTracker) -> Result<()> {
```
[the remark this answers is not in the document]
Stebalien: I don't think it'll make a difference, but that's better API.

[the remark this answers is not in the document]
Stebalien: Eh, won't really matter. We can change it later.

Stebalien (approved):

PR #1150, opened 2022-11-26
feat: add a read-only mode to sends

Stebalien:
Adds a read-only mode for sends.

1. Adds "flags" to the sends syscall. We can extend it in the future to support additional flags.
2. Adds the concept of "read-only" layers to both the event accumulator and the state tree. This means we can track the read-only state at the _lowest_ level.

The general approach here is to reject mutations _at_ the mutation site instead of early. That means we'll charge gas first. The alternative would be to check up-front at the top of the syscall, but that's non-trivial because a send may automatically create an actor in some cases.

### DIFF
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -168,10 +168,11 @@ where
 
     fn with_transaction(
         &mut self,
+        read_only: bool,
         f: impl FnOnce(&mut Self) -> Result<InvocationResult>,
     ) -> Result<InvocationResult> {
-        self.state_tree_mut().begin_transaction();
-        self.events.create_layer();
+        self.state_tree_mut().begin_transaction(read_only);
+        self.events.create_layer(read_only);
 
         let (revert, res) = match f(self) {
             Ok(v) => (!v.exit_code.is_success(), Ok(v)),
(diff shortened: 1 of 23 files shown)


sdk/src/vm.rs
```diff
@@ -8,6 +8,16 @@ use crate::sys;
...
 pub const NO_DATA_BLOCK_ID: u32 = 0;
 
+/// Returns true if the invocation context is read-only. In read-only mode:
+///
+/// - State-tree updates `sself::set_root` are forbidden.
+/// - Actor creation is forbidden.
+/// - Value transfers are forbidden.
+/// - Events are discarded.
+pub fn read_only() -> bool {
+    super::message::MESSAGE_CONTEXT.flags.read_only()
```
Stebalien, line 18: See #1151. This really doesn't belong in `fvm_sdk::message`.

shared/src/error/mod.rs
```diff
@@ -147,6 +147,8 @@ pub enum ErrorNumber {
     Forbidden = 11,
     /// The passed buffer is too small.
     BufferTooSmall = 12,
+    /// The actor is executing in a read-only context.
+    ReadOnly = 13,
```
Stebalien, line 151: I considered using an existing error but:

- IllegalOperation is ambiguous on `set_root` (returned both if the actor has already been deleted, or is in read-only mode).
- Forbidden is ambiguous on `self_destruct` if the beneficiary is self.

In general, I'm starting to regret having "common" errors and am becoming convinced we want more specific ones. We can still share errors between syscalls, but errors like `ActorDeleted` is probably better than `IllegalOperation`.

vyzo:
Shouldnt events be allowed in readonly mode?

fvm/src/call_manager/default.rs
```diff
@@ -168,10 +168,11 @@ where
 
     fn with_transaction(
         &mut self,
+        read_only: bool,
```
vyzo, line 171: What's a reaonly transaction?

Kind of nonsensical, but i guess convenient in not having to change existing code.

fvm/src/call_manager/default.rs
```diff
@@ -715,32 +716,52 @@ where
...
 }
 
 impl EventsAccumulator {
+    fn is_read_only(&self) -> bool {
+        self.read_only_layers > 0
+    }
+
     fn append_event(&mut self, evt: StampedEvent) {
-        self.events.push(evt)
+        if !self.is_read_only() {
```
vyzo, line 728: So they are allowed, just not emitted.

fvm/src/state_tree.rs
```diff
@@ -164,34 +196,19 @@ impl StateSnapshots {
...
-        self.layers
-            .last()
-            .with_context(|| {
-                format!(
-                    "set actor failed to index snapshot layer at index: {}",
-                    &self.layers.len() - 1
-                )
-            })
-            .or_fatal()?
+        self.assert_writable()?;
```
vyzo, line 199: subtle.

vyzo (commented):
first pass on the phone, will do another one on the computer.

Stebalien:
> Shouldnt events be allowed in readonly mode?

They are, they're just dropped.

- This matches EVM behavior.
- This avoids a potential security issue where an attacker might be able to cause another actor to emit an event even if the change won't be committed.

Eventually, I'd like to split "read-only" into multiple limits, but this is a good place to start.

fvm/src/call_manager/default.rs
Stebalien: Yeah, it is kind of subtle... I'm effectively just counting how many times I've entered "read-only" mode so I can exit when I get back to the top.

vyzo (approved):
ship it!

PR #1152, opened 2022-11-27
v3: Change SDK::send's params to take new type IpldBlock

arajasek:
Towards #987, but doesn't close it.

Introduces a new IpldBlock type that can support various codecs (currently DAG_CBOR and IPLD_RAW), and switches the SDK::send method to use `Option<IpldBlock>` for params.

### DIFF
--- a/testing/integration/tests/fil-readonly-actor/src/lib.rs
+++ b/testing/integration/tests/fil-readonly-actor/src/lib.rs
@@ -5,6 +5,7 @@ include!(concat!(env!("OUT_DIR"), "/wasm_binary.rs"));
 
 use cid::multihash::{Code, MultihashDigest};
 use cid::Cid;
+use fvm_ipld_encoding::ipld_block::IpldBlock;
 use fvm_ipld_encoding::{to_vec, RawBytes, DAG_CBOR};
 use fvm_sdk as sdk;
 use fvm_shared::address::{Address, SECP_PUB_LEN};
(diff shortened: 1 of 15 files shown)


shared/src/ipld_block.rs
```diff
@@ -0,0 +1,45 @@
...
+
+impl IpldBlock {
+    pub fn deserialize<'de, T>(&'de self) -> Result<T, Error>
+    where
+        T: serde::Deserialize<'de>,
+    {
+        match self.codec {
+            // IPLD_RAW => BytesDeserializer::new(self.data.as_slice())
+            //     .deser()
+            //     .map_err(Into::into),
```
Stebalien, line 22: Does this not work? (any reason not to just enable it?)

shared/src/ipld_block.rs
```diff
@@ -0,0 +1,45 @@
+use fvm_ipld_encoding::CodecProtocol::Cbor;
+// TODO: We'll probably need our own error type here
+use fvm_ipld_encoding::Error;
+use fvm_ipld_encoding::DAG_CBOR;
+use {serde, serde_ipld_dagcbor};
+
+// TODO: Slapped the Serialize derivations on for some actors testing, not clear to me it should stay
```
Stebalien, line 7: Yep, they shouldn't. Otherwise, we'll serialize twice.

vyzo (commented):
this looks useful.

testing/integration/tests/fil-readonly-actor/src/lib.rs
```diff
@@ -135,7 +139,10 @@ fn invoke_method(blk: u32, method: u64) -> u32 {
             let output = sdk::send::send(
                 &Address::new_id(sdk::message::receiver()),
                 4,
-                RawBytes::new("input".into()),
+                Some(IpldBlock {
+                    codec: DAG_CBOR,
```
Stebalien, line 143: nit: now that we can, consider using DAG_RAW? (if it works)

Stebalien (approved):

PR #1157, opened 2022-11-28
fix: exit code wording

Stebalien:
The grammar was bugging me ("indicates a" instead of "indicates that a"), so I just removed "indicates" and cleaned up the wording a bit.

### DIFF
--- a/shared/src/error/mod.rs
+++ b/shared/src/error/mod.rs
@@ -51,28 +51,29 @@ impl ExitCode {
 
     /// The code indicating successful execution.
     pub const OK: ExitCode = ExitCode::new(0);
-    /// Indicates the message sender doesn't exist.
+    /// The message sender doesn't exist.
     pub const SYS_SENDER_INVALID: ExitCode = ExitCode::new(1);
-    /// Indicates that the message sender was not in a valid state to send this message.
+    /// The message sender was not in a valid state to send this message.
+    ///
     /// Either:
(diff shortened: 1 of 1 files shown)


mriise:
Frankly, removing `The` as well would sound better IMO.

vyzo:
you guys are pathological...

Stebalien:
> Frankly, removing The as well would sound better IMO.

Then it wouldn't be valid english.

mriise:
That would be valid english though, right??

Stebalien:
> edit: Actually this was legitimately something my english teacher encouraged us to avoid.

Using articles? No, no your english teacher did not tell you to avoid using articles.

mriise:
They did! Granted they were fairly opinionated about writing in general and encouraged us to _avoid_ not disregard completely. Wikipeda has some good discussion of this https://en.wikipedia.org/wiki/English_articles, but I'm fairly confident that this is the case where there is no determiner required.

Stebalien:
No, this definitely isn't the same thing.

- You _can_ drop the article at the beginning of a title, but these aren't titles.
- You _can_ drop the article for "contiguous" things. See "Message execution (including subcalls) used more gas than the specified limit."

mriise (approved):
Ok. As a personal preference I would omit "the" in a few places here, but this wouldn't follow the format of other documentation in the FVM. Ultimately this isn't an important change anyways.

PR #1163, opened 2022-11-29
Release FVM, SDK, and shared

arajasek:
As in the title.


Cargo.lock
```diff
@@ -1585,7 +1585,7 @@ name = "fil_actor_embryo"
 version = "10.0.0-alpha.1"
 source = "git+https://github.com/filecoin-project/builtin-actors?branch=next#1b11df4b399550753a4105f45f58bc07015af2a3"
 dependencies = [
- "fvm_sdk 3.0.0-alpha.14 (registry+https://github.com/rust-lang/crates.io-index)",
+ "fvm_sdk 3.0.0-alpha.14",
```
arajasek, line 1588: Should I be concerned that this isn't bumping to 15?

Stebalien (approved):

Cargo.lock
Stebalien: No, we'll bump that _later_ when we update the actors.

PR #1165, opened 2022-11-29
fix: only push backtrace frames on error.

Stebalien:


fvm/src/call_manager/default.rs
```diff
@@ -611,10 +611,6 @@ where
                     value: ret.cloned(),
                 }),
                 Err(abort) => {
-                    if let Some(err) = last_error {
-                        cm.backtrace.begin(err);
-                    }
-
                     let (code, message, res) = match abort {
                         Abort::Exit(code, message, NO_DATA_BLOCK_ID) => (
```
Stebalien, line 615: This (and below) may actually indicate "success". This has _changed_ since FVM v2 because we now support non-local exits.

mriise (approved):

PR #1169, opened 2022-11-30
feat: remove the gas limit from the context

### LINKED ISSUE #1168: Remove "gas limit" from the message context
Users _should_ use the `gas_available` syscall to determine the actual gas available.

I'm proposing that we remove it because it interacts poorly with limiting gas on sends. Basically:

1. We currently update the gas limit to the one specified by the user, and temporarily reset "gas used" to 0.
2. We _skip_ this if the user requests a gas limit greater than gas available.

But this leads to confusing situations where the "gas limit" may be lowered, or may not be.

One fix would be to lower the gas limit but not reset "gas used" for sub-calls. I.e., set the gas limit to `gas_used + requested_gas_limit`. But I'd prefer to just not expose this to the user at all.

Stebalien:
We don't actually need this, and users should use `gas::available`.

fixes #1168

### DIFF
--- a/fvm/src/kernel/default.rs
+++ b/fvm/src/kernel/default.rs
@@ -335,7 +335,6 @@ where
                 .try_into()
                 .or_fatal()
                 .context("invalid gas premium")?,
-            gas_limit: self.call_manager.gas_tracker().gas_limit().round_down() as u64,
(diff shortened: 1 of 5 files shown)


vyzo (approved):
yeah, alright, lets kill it.

Issue #1170, opened 2022-11-30
Add license header


aakoshh:
@vyzo raised in https://github.com/filecoin-project/ref-fvm/pull/1094 that code we worked on should not contain exclusively ChainSafe license headers. 

Headers don't seem to be maintained in the project, some files that originated in Forest have them (unmodified), newer files don't. Ideally we should add Protocol Labs as copyright holder to each file we modify.

PR #1172, opened 2022-11-30
CHORE-1170: Add license headers

aakoshh:
Closes https://github.com/filecoin-project/ref-fvm/issues/1170

The PR adds a `add_license.sh` file similar to [Forest](https://github.com/ChainSafe/forest/tree/main/scripts) that can be invoked with `make license`, which is also called from CI. The script formats files where necessary and returns an error if it did so, which will stop CI, so we have to go back and commit these changes.

The script does two things:
1. It checks that each Rust code file has a license header. If not, it adds the copyright with _Protocol Labs_. 
2. It checks that every Rust file that changed compared to `master` has _Protocol Labs_ in the header, and adds it if not. At this point the common SPDX part should have already been added by 1), or present from when the file was originally copied from _ChainSafe Systems_. 

The effect should be that any file we modify will force us to add _Protocol Labs_. The ones we don't touch are kept as _ChainSafe_. 

I added _Protocol Labs_ to any file that changed since it has been taken in 5c13b82879c31c8a1e37e320b8a55fc736486e1a as a one-off step.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,9 @@ jobs:
       with:
         command: fmt
         args: -- --check
+    - name: Check license headers
+      run: make license
+
   cargo:
     needs: [rustfmt]
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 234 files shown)


vyzo (approved):
this is great, thank you!

PR #1176, opened 2022-11-30
Market singleton in integration test

rllola:
Adding a market singleton in the integration test module


fvm/src/storagemarket_actor.rs
```diff
@@ -0,0 +1,110 @@
...
+use fvm_ipld_hamt::Hamt;
+use fvm_shared::piece::PaddedPieceSize;
+use libipld_core::ipld::Ipld;
+use serde::{de, Deserialize, Deserializer, Serialize, Serializer};
+use fvm_shared::address::Address;
+
+pub type AllocationID = u64;
+
+/// Market actor state
+#[derive(Clone, Default, Serialize_tuple, Deserialize_tuple, Debug)]
```
raulk, line 20: Ouch, we absolutely don't want to have this here. And we're getting to a point where it's best for the integration test framework to stay light _without actor_ assumptions, and to have a separate utility module (in a separate repo) that composes actors with the integration test framework to "prepare" a tester.

I would be less concerned if we moved this type to _inside_ the integration test framework module, only temporarily until we figure out the final layout. WDYT, @Stebalien?

rllola: It is also duplicated code from the builtin actors.

Stebalien: We don't need this. Use the bundle and call the constructor on the market actor to construct the state.

anorth: That might not work. However I'm quite familiar with this problem because I'm actively working on tooling for profiling built-in actors. I'm not at reveal stage yet (e.g. no docs) but I'm prototyping a new thing at https://github.com/anorth/fvm-workbench

It specifically solves the problem of getting all the built-in actors into a VM for testing (and my next step is to connect it to existing builtin actors scenario tests)

Stebalien: Ah... because we assert that the immediate caller is the system actor when it'll be the init actor.

Stebalien: So, looking at the singletons, I'm pretty sure we can construct them, we just need to call the constructor from the "system" actor.

Stebalien: Yeah, it'll work. Basically:

1. Import the actors bundle.
2. Construct each singleton by calling the "Constructor" method on the singleton (method 1).

That will let us avoid having the actual _state_ here.

Stebalien: Yeah, it'll work. Basically:

1. Import the actors bundle.
2. Construct each singleton by calling the "Constructor" method on the singleton (method 1).

That will let us avoid having the actual _state_ here.

rllola:
Closing this in favor of https://github.com/filecoin-project/ref-fvm/pull/1387

Issue #1160, opened 2022-11-29
GAS: Calibration contract


aakoshh:
The conclusion of https://github.com/filecoin-project/ref-fvm/pull/1148 is that using the test vectors may a good approach to look for mispriced syscalls, but it won't work for determining appropriate prices for things that don't have a price at the moment. 

For that we will make custom contracts that we can use to gather relevant data in higher volume, to be able to estimate model parameters. This can be in the form of an additional contract following the pattern at https://github.com/filecoin-project/ref-fvm/tree/master/testing/integration/tests with a test that exercises it and exports traces for visualization and regression. 

For example we can generate an arbitrary bytevector of size N (either in test or in the contract), pass it to the contract and tell it to hash it M times (so we see the natural variance independent of size). We capture the gas metrics and export the data, _attaching N_ to it as well. Then we try with a different value of N. Finally we can run a regression to check if the size of N is a good predictor of _execution time_, and thus come up with a gas price for things that are not priced at the moment.

We can do the same kind of experiment with reading/writing IPLD data (in that case it's probably better to generate the data in the contract so the data we put can be different each time in M).

PR #1192, opened 2022-12-02
GAS-1160: Calibration contract

aakoshh:
Closes https://github.com/filecoin-project/ref-fvm/issues/1160

The PR adds a `testing/calibration` crate with a `fil-gas-calibration-actor` contract that we can use to focus in on individual syscalls and measure the time it takes to execute them with various input sizes. The crate uses the measurements to run a least squares regression to estimate coefficients that could be used to configure the gas price.

I checked in the plots so we can see the how good the model is on these isolated cases. We can repeat the calibration on a standard environment to get prices we can add to the production price list. 

One thing I want to do to validate if this approach works is to take the coefficients as they are and put them in a calibration price list, then use that price list to run the conformance test vectors, and see what the visualizations in https://github.com/filecoin-project/ref-fvm/pull/1148 look like on the same machine the prices were calibrated for. Ideally they would be more linear with higher R-squared values.

Should:
- [x] `OnHashing`
- [x] `OnBlock*`
- [x] `OnVerifySignature`
- [x] `OnRecoverSecpPublicKey`

Could:
- [ ] `OnComputeUnsealedSectorCid`
- [ ] `OnVerifySeal`
- [ ] `OnVerifyAggregateSeals`
- [ ] `OnVerifyReplicaUpdate`
- [ ] `OnVerifyPost`

Won't:
- `OnVerifyConsensusFault` (using `DummyExterns`)
- `OnGetRandomness` (using `DummyExterns`)
- `wasm_exec` (the test vectors should be okay for this, but what it would really need is opcode level breakdown)
- `wasm_memory_init`: I tweaked it a little so that the time we associate with this takes into account the everything around module instantiation. The current charge depends on the memory size, not the wasm size, though; I have no idea what the instantiation time will actually depend on. In any case the calibration actor is not a good test for this, because there's no variation in size. We should use the test vectors where at least the various built-in actors are used. It's just tricky to say how we could associate the message with the code size from the test runner. It would have to be an explicit lookup of the code based on the recipient. It looks different enough to be put into a followup PR. Alternatively, we could just load a built-in actor bundle, get their addresses, then throw random messages at them - they won't be processed, but that's fine because we'd only measure module instantiation time. Yet another alternative would be to use @Jacarte wasm corpus to get a wide variety of sizes - again with random messages, just measuring instantiation time.

Can't:
- All the new prices added in https://github.com/filecoin-project/ref-fvm/pull/1139 would need merging this first, then cherry-picking it to be on top of `master`.

# Analysis

## `overall.png`

The overall charts shows how much we spent doing each charge vs how much we charged for it in gas. The slopes and magnitudes vary wildly, which might show relative mispricing in action. The caveat is where we the prices try to incorporate delayed IO costs.

## Per charge

Each chart shows the input size used by the test driver versus the time we spent doing the operation with that input, and the gas that was charged for it. The gas is displayed as a line on the Y2 axis. Looking at this it's easy to spot two things:
* When we charge a constant gas for something where the execution time actually depends on the input size, the line is horizontal.
* When we charge some base cost, but actually the time we spend with 0 input is near zero, then the gas line intercepts the Y axis at a non-zero point, while the time points go at zero. 

It is tempting to compare the slopes of the Gas and the Time points and call it a good fit, however, that's just an illusion, because they are on different Y axes. To meaningfully compare their slopes, we'd have to agree on how much we'd _like_ to charge for a unit of time, e.g. 100 gas per nanosecond, and then _divide_ the gas with that. Then we could display both in terms of nanos, on the same Y axis. 

To see how different these seemingly fitting prices are, look at `overall.png`.

### `OnHashing`

Hashing time is very linear in the size, with the intercept being at virtually 0. However the different hashing algorithms have markedly different slopes, so it would make sense to differentiate the price by hashing type.

### `OnBlockCreate` 

There is a nice linear relationship captured on the chart. However it's exhibits strange behaviour if certain things are changed in the code in that reordering the steps (like going from high to low sizes, or interleaving puts and gets) can introduce a second, cheaper slope. 

If we look closer, up to 100,000 bytes the cheaper slope is still present.

I left it at this version because this had the highest slope and I thought over estimating the cost is a better default. 

### Other `OnBlock*`

Similar to `OnBlockCreate`. 

### `OnVerifySignature`

The chart shows that BLS and Secp256k1 are indeed different, with both having significant non-zero intercepts at 0 byte input size; actually there's huge deviation in runtime when the input size is zero. Yet the gas price is constant, although that constant depends on the signature type. We have to make it depend on input size as well.

Looking at he overall chart, we can see that the minimum time for BLS is just above the maximum time measured for Secp256k1, yet the price difference between them is 10x, which is unfair.

### `OnRecoverSecpPublicKey`

This operation works on hashes, not plain text, so the runtime doesn't depend on the input size. This is visible on the chart: we generate different input sizes but just so we can see a nice horizontal distribution (ie. the hashing is explicitly happening outside the syscall we measure). 

What's not visible on the chart is whether the actual gas price is adequate or not. This we could only tell if there was a target ratio between time and gas, with which we could divide the gas to compare nanos to nanos.

## Trial on the Conformance Test Vectors

I took the slopes from the regressions and used them in the price model of `on_block_*` method and `on_hashing` to see if the conformance test vectors would show better correlation between time and gas. I left the intercepts alone, because that was often negative, and looking at the charts did not contradict that they intercept the Y axis at zero.

Unfortunately no, it's mostly the same, some R-squared are better, some worse. 

I noticed that for example with `OnBlockCreate`, even on the summed charts that look at the aggregate time spent by a contract doing a given thing is on average 10,000 nanos, which is well in the domain where the observations exhibited diverging costs. That suggests that the contracts are writing so little data that it's not a good predictor of the runtime. 

I also suspected that measuring time using the in-memory blockstore is not a good way to model, but I'm not yet sure how to compensate for that. Maybe we can use a persistent block store, even if in production it's the buffered one; we'll write things eventually, most of the time, I guess.

# Follow up actions
* Deal with memory init in terms of Wasm size in a different PR (and rethink if it should be a different charge)
* Change the gas on the charts into nanos at 10,000milligas/nanos rate. Done in https://github.com/filecoin-project/ref-fvm/pull/1242
* Try to update the prices based on the regression and see if the overall slopes become better aligned
* Rebase onto master
* Add more charges that appeared there
* Run the same tests on a canonic environment
* Fix the prices based on the canonic regression results


fvm/src/syscalls/mod.rs
```diff
@@ -141,22 +141,34 @@ pub fn charge_for_exec<K: Kernel>(
...
     data.last_memory_bytes += min_memory_bytes;
-    data.last_charge_time = GasTimer::start();
 
-    Ok(())
+    data.kernel.charge_gas("wasm_memory_init", memory_gas)
+}
+
+/// Record the time it took to initialize a module.
+///
+/// This is
```
Stebalien, line 158: ?

Stebalien (approved):
LGTM!

fvm/src/syscalls/mod.rs
aakoshh: Thanks, no idea :sweat_smile:

aakoshh:
When the contracts do IPLD block operations they:
* read from the cache, or if it's not in the cache then from disk
* write to the cache
And only at the end of the block execution, Lotus writes everything from the cache to the disk, in parallel. 

We only have disk based block stores implemented in Lotus itself, so it's not something I can easily replace the in-memory one in _this_ repo, to do some experiments.

Apparently what the team did to account for the discrepancy between the time it takes to execute a write in the SDK and what it will eventually take in addition, was to look at the block execution time and upscale the gas based on what they measured and how long it actually take to finish a block. 

If I knew what this scaling factor was, then we could selectively multiply the `elapsed_nanos` of certain `Block*` operations (and only those, because `OnHashing` for example will not result in bytes written or read from disk). That way we can estimate the regression as-if we'd be measuring against our hypothetical disk.

@Stebalien do you know this conversion factor, perhaps?

Stebalien:
> @Stebalien do you know this conversion factor, perhaps?

Going back to my notes:

- I have no idea how this was originally decided.
- In FVM M1, we just mixed it into the instruction cost.

I also only did early gas analysis. @Kubuxu and/or @raulk may have later data.

PR #1199, opened 2022-12-05
v2: Change SDK::send's params to take new type IpldBlock

arajasek:
IpldBlock supports multiple codecs, including IPLD_RAW.

Serializing byte vectors will correctly serialize anything that serializes to "raw bytes". However, it _won't_ serialize "byte lists", so `Vec<u8>` and `[u8]` won't work. But that shouldn't be an issue in practice.

Users _should_ serialize to/from `ByteBuf`.

### DIFF
--- a/ipld/encoding/src/cbor.rs
+++ b/ipld/encoding/src/cbor.rs
@@ -9,8 +9,6 @@ use serde::{Deserialize, Serialize};
 use super::errors::Error;
 use crate::{de, from_slice, ser, to_vec};
 
-pub const DAG_CBOR: u64 = 0x71;
-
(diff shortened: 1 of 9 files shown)


ipld/encoding/src/cbor.rs
```diff
@@ -9,8 +9,6 @@ use serde::{Deserialize, Serialize};
 use super::errors::Error;
 use crate::{de, from_slice, ser, to_vec};
 
-pub const DAG_CBOR: u64 = 0x71;
```
Stebalien, line 12: unbreak please (either alias, or move it back)

Stebalien (approved):
Modulo breaking change.

Issue #1038, opened 2022-10-28
Wasm Validation Rules


Stebalien:
- Can only export an `invoke(i32) -> i32` function. This will likely change to allow a `validate` and/or `constructor` function in the near future, but we're not allowing arbitrary functions.
- Cannot export any globals? We don't currently enforce this, but modules have no reason to do so.
- Cannot import any globals.
- Can only import functions defined in the syscall API (https://github.com/filecoin-project/ref-fvm/blob/a8f15212940faa3074848605a19cd9ede286e1ef/fvm/src/syscalls/mod.rs#L97).
- Cannot use reference types.
- Cannot import or export any tables (I think?).
- Cannot import any memories.
- Must export exactly one memory ("memory").
- Cannot use atomics.

Jacarte:
Adding another rule:
- The binary should not exceed 5Mb in size.

Stebalien:
Allocates at most X MiB of memory up-front (i.e., doesn't statically declare more than some amount of minimum memory).

Stebalien:
Currently, code can run before `invoke`, but it'll run out of gas immediately and trap (leading to a fatal error).

We need to either:

1. Properly account for this kind of code.
2. Forbid it.

Ideally, we'd just forbid it. It looks like we can run `wasm-ctor-eval` (from binaryen)?

Specifically, we need to handle:

1. The start function.
2. Table initialization functions. See the `expr`s that appear in https://webassembly.github.io/spec/core/syntax/modules.html#syntax-elem.

Migrated from https://github.com/filecoin-project/ref-fvm/issues/607.

Jacarte:
Another rule to add:
- Limit the initial memory

Apparently, cranelift is packaging the initial declared memory in the module. A simple Wasm binary of 600bytes can go to several Mb in the compiled module.

Stebalien:
I wonder if it's related to https://docs.wasmtime.dev/api/wasmtime/struct.Config.html#method.memory_init_cow?

Stebalien:
Yeah, it looks like https://docs.wasmtime.dev/api/wasmtime/struct.Config.html#method.memory_guaranteed_dense_image_size. Which sets a maximum of 16MiB, which isn't _terrible_ but something we may want to lower.

Stebalien:
@aakoshh any reason to leave `memory_init_cow` enabled? IMO, we should consider _disabling_ it because we'll need to charge for it anyways.

Jacarte:
[...]

PR #1200, opened 2022-12-05
feat: disable memory_init_cow

Stebalien:
Otherwise, some Wasm modules will balloon in size when compiled.

### DIFF
--- a/fvm/src/machine/engine.rs
+++ b/fvm/src/machine/engine.rs
@@ -109,8 +109,8 @@ fn wasmtime_config(ec: &EngineConfig) -> anyhow::Result<wasmtime::Config> {
     });
 
     // wasmtime default: true
-    // Included here to make sure the memory-init-cow feature is enabled.
-    c.memory_init_cow(true);
+    // We disable this as we always charge for memory regardless and `memory_init_cow` can baloon compiled wasm modules.
+    c.memory_init_cow(false);
 
     // wasmtime default: 4GB
     c.static_memory_maximum_size(instance_memory_maximum_size);


Stebalien:
See https://github.com/filecoin-project/ref-fvm/issues/1038#issuecomment-1332614117.

aakoshh:
The `memory_init_cow` setting was to make it run faster and with potentially less memory requirement, e.g. when the module calls itself recursively it would only need physical memory once, and for the diffs, it would not need to copy every time. 

Since we limit the maximum memory on the call stack anyway, this should not be a security risk, in that as long as the validator's machine can cover the limit, then there's no way a non-optimal selection of contract calls could run out of memory, while an optimally reusable one would not. 

It's a good point whether this should be disabled "because we already make them pay of initialization". FWIW on my gas calibration branch I'm [trying to associate the time](https://github.com/filecoin-project/ref-fvm/blob/89cbf02cdfaf71b69520aa97cc873be3bfdc34ad/fvm/src/syscalls/mod.rs#L152) with init since the context was created. To calibrate the cost of initializing memory in terms of the time it takes, the CoW feature should be momentarily disabled, or it can lead to under-estimating the cost of initialisation. 

But if we disable it during calibration, it should be okay to enabled it in the runtime and make everything go faster, at least on Linux. 

I didn't know that there was a trade-off with the compile size, though. Correct me if I'm wrong, but these are not stored in the block store, right? Only the raw Wasm module binary is stored, the extra memory image is only held in memory, isn't it? 

If that were the case, wouldn't disabling this result in the initial heap content being recreated after the module is instantiated, and actually duplicated in physical memory, because it can't point at the same page?

Stebalien:
> I didn't know that there was a trade-off with the compile size, though. Correct me if I'm wrong, but these are not stored in the block store, right? Only the raw Wasm module binary is stored, the extra memory image is only held in memory, isn't it?

For now, they're stored in memory. In the future, with user specified contracts, we'll have to store them in a cache on-disk.

Basically, my goal here is to make memory usage more deterministic. It's an implementation detail of wasmtime, so we're going to have to charge for this memory _anyways_ even if we're able to take advantage of CoW semantics.

But you're right, we can always re-enable this feature later (or tune the limits of how much initial memory can be reserved).

Stebalien:
I'm merging for now so that testing, benchmarking, fuzzing, etc. take this into account.

aakoshh:
> For now, they're stored in memory. In the future, with user specified contracts, we'll have to store them in a cache on-disk.

That makes sense, of course. What I meant was that the part about the extra memory image that Cranelift generates would not be on disk, it is a purely runtime value that lives in memory to speed up instantiation. So this bloat in size is not bloating the storage cost.

Issue #918, opened 2022-10-03
EVM runtime: pass chain ID in NetworkContext


raulk:
See https://github.com/filecoin-project/builtin-actors/pull/713#discussion_r985943222.

raulk:
IMO:
- This can happen after feature completeness.
- Not consensus breaking (except for CodeCID change in EVM actor).

Stebalien:
Well, it's _definitely_ consensus breaking. It just won't break smart contracts.

But you're right, we can punt (although it would be easier to just fix this now given how simple it is).

Stebalien:
We'll put this in the "if we have time" pile.

Stebalien:
Er, well, this isn't quite done. We now feed the chain ID into the actors, but we don't actually _use_ it in the EVM. To do that, we need to:

1. Add a `chain_id` method to the actors `Runtime`.
2. Change the `CHAINID` instruction to return the real chain ID.

Stebalien:
We also need to feed the chain ID down from lotus and through the FFI.

Stebalien:
Nvm. I've just field two more issues.

https://github.com/filecoin-project/ref-fvm/issues/1231
https://github.com/filecoin-project/ref-fvm/issues/1230

PR #1203, opened 2022-12-06
Add Chain ID to `NetworkContext`

mriise:
closes https://github.com/filecoin-project/ref-fvm/issues/918


fvm/src/machine/mod.rs
```diff
@@ -39,6 +39,20 @@ pub const REWARD_ACTOR_ID: ActorID = 2;
 /// Distinguished Account actor that is the destination of all burnt funds.
 pub const BURNT_FUNDS_ACTOR_ID: ActorID = 99;
 
+#[derive(Clone, Copy, Debug)]
+pub struct ChainID(u64);
```
Stebalien, line 43: We're going to need a way to convert to/from a number from the FFI.

Stebalien (approved):

PR #1209, opened 2022-12-07
feat: mark hamt/kamt config as non_exhaustive

Stebalien:
That way, we can extend these in the future without breaking changes.


Stebalien:
Apparently the rust devs made some interesting design decisions here... So... whatever.

PR #1211, opened 2022-12-07
Revert updates to cid/multihash

Stebalien:
Unfortunately, updating cid & mutlihash breaks pretty much everything:

- The FFI imports two versions of the FVM which now use two different CID/Multihash versions.
- We have a common blockstore trait that uses CID in the interface.
- This will likely cause hell for builtin actors.

We can un-revert this if we can backport the changes. But we're not updating cid and/or multihash for the foreseeable future.


Stebalien:
If someone wants to make this happen, they'll need to either:

1. Create patch releases of cid/multihash.
2. Add an abstraction to the filecoin-ffi to handle translating CIDs and the blockstore.

But I really don't have time for that right now.

Stebalien:
@lemmih I'm really sorry about this. If you can find another way, I'd love to hear it.

arajasek (approved):

lemmih:
> @lemmih I'm really sorry about this. If you can find another way, I'd love to hear it.

Can't we just bump the CID dependency everywhere?

lemmih:
@Stebalien Given that `fvm` doesn't follow semantic versioning, how about we bump to `cid-0.9` and release a new version of `fvm-2`?

PR #1239, opened 2022-12-09
refactor: extract the engine from the machine and make it a pool

Stebalien:
Now that we're using the pooling allocator in wasmtime, we have to be careful about sharing a single engine between multiple concurrent message executions. Otherwise, we could run out of wasm "instances" and fail execution (e.g., if the user is validating two blocks in parallel).

To handle this, I've:

1. Turned the engine into an "engine pool" where we allocate `concurrency * call_depth` instances.
2. Put the EnginePool inside the Executor.
3. Put the Engine itself inside the CallManager.

That way, we "acquire" an Engine every time we handle a new message, and release the engine when we're done.

Unfortunately, putting the engine on the _machine_ didn't really make sense anymore, because the engine's lifetime is tied to a specific call and returned to the pool afterwards.

We could have acquired the engine for the lifetime of the _Machine_, but I was concerned about deadlocks.

### DIFF
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -1,6 +1,7 @@
 // Copyright 2021-2023 Protocol Labs
 // SPDX-License-Identifier: Apache-2.0, MIT
 use std::mem;
+use std::rc::Rc;
 
 use anyhow::{anyhow, Context};
 use cid::Cid;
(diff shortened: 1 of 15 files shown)


Stebalien:
I'd generally prefer not to make refactors like this... but I'm actually pretty happy with the results this time. The engine now lives inside the _call manager_, which is where it really belongs (the machine manages the state, the call manager actually executes the wasm).

vyzo (approved):
This looks reasonable.

PR #1242, opened 2022-12-09
GAS-1160: Convert gas to time on chart.

aakoshh:
Applies a transformation from gas to time on the charts, so that we can see directly how the slope of the gas model relates to the observations. 

# Analysis

Caveat: that this is just my laptop.

* `OnBlockCreate`: Overcharging.
* `OnBlockLink`: Massively undercharging.
* `OnBlockOpenPerByte`: Overcharging.
* `OnBlockRead`: Nice fit, but undercharging at the highest sizes.
* `OnRecoverSecpPublicKey`: Massively undercharging.
* `OnVerifySignature`: The intercepts are charged in nicely, but it doesn't scale with the size.
* `OnHashing`: Massively undercharged.


Stebalien (approved):

Stebalien:
> OnBlockCreate: Overcharging.

(also need to account for flush, which is kind of hard)

> OnBlockLink: Massively undercharging.

Yeah.... We basically just lumped this into OnBlockCreate before. We'll probably steal gas from there.

> OnBlockOpenPerByte: Overcharging.

This is going to be hard to measure without a real client and/or datastore.

Stebalien:
In terms of OnBlockCreate/OnBlockOpen...

We're charging an additional "memret" (memory retention) fee because we didn't have hard memory limits on the number of open blocks. This was designed to limit the cache size to 1GiB, worst case.

We should either discount the "memret" fee, or actually implement limits here.

PR #1243, opened 2022-12-09
feat: update instrumentation

Stebalien:
We're not yet taking advantage of it, but this uses the new
instrumentation logic.

### DIFF
--- a/fvm/src/engine.rs
+++ b/fvm/src/engine.rs
@@ -10,7 +10,6 @@ use anyhow::{anyhow, Context};
 use cid::Cid;
 use fvm_ipld_blockstore::Blockstore;
 use fvm_wasm_instrument::gas_metering::GAS_COUNTER_NAME;
-use fvm_wasm_instrument::parity_wasm::elements;
 use wasmtime::OptLevel::Speed;
 use wasmtime::{
     Global, GlobalType, InstanceAllocationStrategy, InstanceLimits, Linker, Memory, MemoryType,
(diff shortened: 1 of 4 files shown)


fvm/src/engine.rs
```diff
@@ -564,18 +555,3 @@ impl Engine {
         store
     }
 }
-
-// Workaround for https://github.com/filecoin-project/ref-fvm/issues/602
-//
-// This removes the out-of-order data count section, if it exists, and re-inserts it (with the
-// correct data-count).
-fn fix_wasm_sections(module: &mut elements::Module) {
```
Stebalien, line 572: We no longer need this hack.

fvm/src/engine.rs
```diff
@@ -392,17 +391,13 @@ impl Engine {
...
-
-        let m = deserialize_buffer(raw_wasm)?;
+        use fvm_wasm_instrument::{gas_metering, stack_limiter};
 
         // stack limiter adds post/pre-ambles to call instructions; We want to do that
         // before injecting gas accounting calls to avoid this overhead in every single
         // block of code.
-        let m =
-            inject_stack_limiter(m, self.0.config.max_wasm_stack).map_err(anyhow::Error::msg)?;
+        let raw_wasm = stack_limiter::inject(raw_wasm, self.0.config.max_wasm_stack)
```
magik6k, line 399: Do we validate the bytecode anywhere before this?

magik6k: Assuming that this will be done on actor install when wasm contracts are a thing

magik6k (approved):

fvm/src/engine.rs
Stebalien: We do at the top of this function as well.

Stebalien:
This has reduced gas costs of the SimpleCoin contract (see https://github.com/filecoin-project/fvm-bench/) by 4. That is, 4 measly gas.

So, now we need to know _why_.

Stebalien:
We weren't charging for memory instructions at all. Fixed.

PR #1244, opened 2022-12-09
feat: add charging for all memory copy/init operations

### LINKED ISSUE #1075: Charge for table growing

Stebalien:
I'm not actually charging anything right now, but this lets us charge for:

1. Memory copying.
2. Memory "filling" (where I'm treating all memory expansion as "filling").

fixes #1075

### DIFF
--- a/fvm/src/gas/price_list.rs
+++ b/fvm/src/gas/price_list.rs
@@ -293,8 +293,8 @@ lazy_static! {
             exec_instruction_cost: Gas::new(4),
             memory_fill_per_byte_cost: Zero::zero(),
             memory_copy_per_byte_cost: Zero::zero(),
-            memory_fill_base_cost: Gas::new(4),
-            memory_copy_base_cost: Gas::new(4),
+            memory_fill_base_cost: Gas::zero(),
+            memory_copy_base_cost: Gas::zero(),
         },
 
         event_emit_base_cost: Zero::zero(),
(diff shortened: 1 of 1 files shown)


Stebalien:
Based on #1243.

fvm/src/gas/price_list.rs
```diff
@@ -1047,22 +1076,30 @@ impl Rules for WasmGasPrices {
...
-                match expansion_cost
-                    .try_into().ok() // zero or not zero.
-                {
-                    Some(cost) => InstructionCost::Linear(0, cost),
-                    None => InstructionCost::Fixed(0),
-                }
-            }),
+            // Memory related instructions...
+            Operator::TableInit { .. } | Operator::TableCopy { .. } => linear_cost(
+                Gas::zero(),
```
magik6k, line 1081: No base?

magik6k (approved):

fvm/src/gas/price_list.rs
Stebalien: Hm. Yeah, I'll make it configurable. These are all zero right now anyways.

Stebalien: Done.

Stebalien:
This further reduces gas fees by 504. But it shouldn't change anything.

Stebalien:
Oh.... It's because we were charging a flat cost before. Let me fix that.

Stebalien:
Ok, now this costs 4 more than master... 8 more than the previous PR.

That's _probably_ because we're charging for tables.

Stebalien:
Yeah, we're now charging for exactly one more instruction than we did before.

Stebalien:
Ah... fixed.

PR #1246, opened 2022-12-09
feat: make some operations free

### LINKED ISSUE #1245: Wasm gas: Stop charging for "empty" instructions
Specifically, stop charging for:

- Const instructions.
- Cast instructions (that don't sign-extend).

Stebalien:
fixes #1245

### DIFF
--- a/fvm/src/gas/price_list.rs
+++ b/fvm/src/gas/price_list.rs
@@ -1093,6 +1093,16 @@ impl Rules for WasmGasPrices {
             | Operator::Return
             | Operator::Else
             | Operator::End => Ok(InstructionCost::Fixed(0)),
+            // Make casts and constants free.
+            Operator::I64ExtendI32U // only unsigned is free.
+            | Operator::I32WrapI64 // truncates
+            | Operator::I32ReinterpretF32 // casts
+            | Operator::I64ReinterpretF64 // casts
+            | Operator::V128Const { .. }
+            | Operator::I32Const { .. }
+            | Operator::I64Const { .. }
+            | Operator::F32Const { .. }
+            | Operator::F64Const { .. } => Ok(InstructionCost::Fixed(0)),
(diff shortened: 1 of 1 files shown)


fvm/src/gas/price_list.rs
```diff
@@ -1076,6 +1076,16 @@ impl Rules for WasmGasPrices {
             | Operator::Return
             | Operator::Else
             | Operator::End => Ok(InstructionCost::Fixed(0)),
+            // Make casts and constants free.
+            Operator::I64ExtendI32U // only unsigned is free.
+            | Operator::I32WrapI64 // truncates
+            | Operator::I32ReinterpretF32 // casts
+            | Operator::I64ReinterpretF64 // casts
+            | Operator::V128Const { .. }
```
magik6k, line 1084: V128 is simd, I'm not sure if we can do simd

Stebalien: We don't.... I guess I can just nop-out all SIMD instructions.

Stebalien:
Saves about 5% of gas.

fvm/src/gas/price_list.rs
```diff
@@ -1093,6 +1093,16 @@ impl Rules for WasmGasPrices {
...
+            | Operator::I32Const { .. }
+            | Operator::I64Const { .. }
+            | Operator::F32Const { .. }
+            | Operator::F64Const { .. } => Ok(InstructionCost::Fixed(0)),
```
Stebalien, line 1105: @Jacarte so.... have you seen any cases where one might be able to make a very slow wasm executable with many consts? My assumption is that they'll slow down compilation, but basically get folded into instructions at runtime. But I'm not 100% sure.

Stebalien:
_This_ one changes fees, so we'll need to sit on this for a bit.

PR #1248, opened 2022-12-11
Implement MMVP Account Abstraction

arajasek:
Description to come, FYI for @Stebalien

### DIFF
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -27,7 +27,7 @@ use crate::state_tree::ActorState;
 use crate::syscalls::error::Abort;
 use crate::syscalls::{charge_for_exec, update_gas_available};
 use crate::trace::{ExecutionEvent, ExecutionTrace};
-use crate::{account_actor, syscall_error, system_actor};
+use crate::{syscall_error, system_actor};
 
 /// The default [`CallManager`] implementation.
 #[repr(transparent)]
@@ -424,34 +424,6 @@ where
         Ok(id)
     }
 
-    fn transform_into_eoa<K>(&mut self, addr: ActorID, actor: ActorState) -> Result<()>
-    where
-        K: Kernel<CallManager = Self>,
-    {
-        self.charge_gas(self.price_list().on_create_actor(false))?;
-
-        // Transform the actor code CID into the EOA code CID.
-        let eoa_code_cid = self.builtin_actors().get_eoa_code();
-        self.state_tree_mut().set_actor(
-            addr,
-            ActorState {
-                code: eoa_code_cid,
-                ..actor
-            },
-        )?;
-
-        // Now invoke the constructor.
-        self.send_resolved::<K>(
-            system_actor::SYSTEM_ACTOR_ID,
-            addr,
-            fvm_shared::METHOD_CONSTRUCTOR,
-            None,
-            &TokenAmount::zero(),
-        )?;
-
-        Ok(())
-    }
-
     fn create_embryo_actor<K>(&mut self, addr: &Address) -> Result<ActorID>
     where
         K: Kernel<CallManager = Self>,
@@ -477,14 +449,6 @@ where
     where
         K: Kernel<CallManager = Self>,
     {
-        let from_actor = self.state_tree().get_actor(from)?.ok_or_else(Err(
-            syscall_error!(NotFound; "sender does not exist: {}", to).into(),
-        ))?;
-
-        if self.builtin_actors().is_embryo_actor(from_actor.code) {
-            self.transform_into_eoa(from, from_actor)?;
-        }
-
         // Get the receiver; this will resolve the address.
         let to = match self.state_tree().lookup_id(&to)? {
             Some(addr) => addr,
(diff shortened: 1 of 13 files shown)


fvm/src/machine/manifest.rs
```diff
@@ -200,4 +212,9 @@ impl Manifest {
     pub fn get_embryo_code(&self) -> &Cid {
         &self.embryo_code
     }
+
+    /// Returns the code CID for the EOA actor.
```
Stebalien, line 216: Can we explicitly call this the _ethereum_ EOA? Or just, Ethereum Account?

fvm/src/call_manager/default.rs
```diff
@@ -415,6 +415,34 @@ where
         Ok(id)
     }
 
+    fn transform_into_eoa<K>(&mut self, addr: ActorID, actor: ActorState) -> Result<()>
```
Stebalien, line 418: 1. Call this `maybe_deploy_account`? In the future, it'll deploy other types.
2. Check if address is f410.

fvm/src/call_manager/default.rs
```diff
@@ -440,6 +468,14 @@ where
     where
         K: Kernel<CallManager = Self>,
     {
+        let from_actor = self.state_tree().get_actor(from)?.ok_or_else(Err(
+            syscall_error!(NotFound; "sender does not exist: {}", to).into(),
+        ))?;
+
+        if self.builtin_actors().is_embryo_actor(from_actor.code) {
+            self.transform_into_eoa(from, from_actor)?;
+        }
```
Stebalien, line 477: We shouldn't do this inside the call manager as it only applies to external sends. Can we put this in the executor before we call into the callmanager?

fvm/src/call_manager/default.rs
Stebalien: (should also be logic in the executor and/or machine (likely the gas charging in the executor, and the functionality in the machine?)

fvm/src/executor/default.rs
```diff
@@ -385,22 +385,26 @@ where
...
-
-        // Unless we have the f4-as-account hack enabled, and the sender is an embryo created by the EAM.
-        #[cfg(feature = "f4-as-account")]
-        let sender_is_account = sender_is_account
-            || sender
+        // Sender is valid if it is:
+        // - an account actor
+        // - an EOA actor
+        // - an embryo actor with nonce 0 that has an f4 address in the EAM's namespace
+
```
Stebalien, line 392: Can we somehow deploy right here? Or is this too early.

fvm/src/machine/manifest.rs
arajasek: Sure thing

arajasek:
@Stebalien I've added a commit moving this into the executor. It's not very pretty, but I think it'll be fine with some beautification (it should be functionally correct).

fvm/src/executor/default.rs
```diff
@@ -123,17 +124,52 @@ where
...
+                        cm.state_tree_mut().set_actor(
+                            sender_id,
+                            ActorState {
+                                code: eoa_code_cid,
+                                ..sender_state
+                            },
+                        )?;
+
+                        // Now actually invoke the constructor
+                        let ret = cm.send::<K>(
```
Stebalien, line 149: Hm. Now I see why you wanted this inside the call manager.

For now, I'd consider just not running the constructor (it doesn't actually do anything...).

In the future, I expect we'll want to move all of the logic (inside this `with_transaction` into some `send_external` function on the `CallManager`.

But up to you.

Stebalien: (Really, I think we're going to need a bigger refactor in the future.

fvm/src/executor/default.rs
```diff
@@ -123,17 +124,52 @@ where
...
+                            None,
+                            &TokenAmount::zero(),
+                            None,
+                        )?;
+
+                        // If we _failed_ to deploy the EOA actor, we abort early.
+                        // If we succeeded, proceed to the actual invocation.
+                        if !ret.exit_code.is_success() {
+                            return Ok(ret);
+                        }
```
Stebalien, line 162: Should exit with `SYS_SENDER_INVALID` (or a new exit code). We can't return the exit code from this actor, because, if this fails, this actor doesn't actually exist at this address and this call never happened...

fvm/src/executor/default.rs
```diff
@@ -123,17 +124,52 @@ where
             });
 
             let result = cm.with_transaction(false, |cm| {
+                let pl = cm.context().price_list;
+
+                if let Some(sender_state) = sender_st {
```
Stebalien, line 129: note: we can probably afford to lookup the sender state again. I'd consider just returning some form of "sender needs to be created" flag, or something like that?

fvm/src/executor/default.rs
arajasek: Yeah, you're right.

fvm/src/executor/default.rs
Stebalien: I don't have a strong preference. My concern is that we might insert some other mutation between there and here, but if you don't think that's an issue, I have no strong objections.

Stebalien (approved):

arajasek:
Integ tests will need the actors PR to land before :green_heart:

Issue #1247, opened 2022-12-10
Memory Gas Costs


Stebalien:
We need to know:

- Cost to randomly read one word.
- Cost to randomly write one word.
- Write cost per byte (bulk).
- Read cost per byte (bulk).

From this, we can derive most of the "variable" instruction costs in the FVM.

Stebalien:
Measurement strategy....

1. Perform setup.
2. Log nothing 10 times.
3. Perform the test (ideally _many_ times).
4. Log once.

We can use the first 10 logs to determine accurate syscall overhead, and the time between the last two logs (minus that overhead) will be the time spent running the test itself.

# Read Latency

To time reads, we need to:

**Setup**:

Allocate 256MiB of memory, initialize to random addresses in the range `[0, 256MiB)`.

**Test**:

1. Read the first address.
2. Repeatedly dereference addresses (some fixed number of times). Prefer to use a long sequence of instructions, not a loop.

# Memcpy

Once we have the read timing, we can measure memcpys. We'll need to re-run this test for different copy sizes (N), likely powers of two (fuzzed a bit to break alignment).

**Setup**:

Allocate 256MiB+N of memory (or a bit more, initialize to random addresses in the range `[0, 256MiB)`.

The `+N` avoids an out of bounds error if we copy past the end.

**Test**:

Like above, we'll be reading addresses from memory. However, instead of _just_ chaining reads, we'll need to:

1. Starting at address 0.
2. Read three addresses from memory (sequential loads).
3. Memcpy (using the bulk memory instruction) some fixed number of bytes N from the first address to the second.
4. Go back to step 2 starting at the third address.

When determining gas values, _subtract_ gas for one random memory access for each iteration.

# Fill

Finally, we need to measure the cost of "filling". Again, we'll need to repeat for various sizes N.

**Setup**:

Allocate 256MiB+N of memory, initialize the first half to random addresses in the range `[0, 128MiB)`.

**Test**:

1. Starting at address 0.
2. Read an address (A).
3. Multiply by 2 (B).
4. Fill N bytes at address B with some X (chosen when writing the benchmark).
5. Go back to 1 but read the address from memory index A instead of 0.

Stebalien:
For now, I'm going to _assume_:

1. 0.5gas per memory byte copied and/or filled.
2. 100gas per random read (10ns which is about the random access speed of most DDR4 RAM).

Stebalien:
[...]

PR #1249, opened 2022-12-11
feat: explicit gas charges for different instruction types

### LINKED ISSUE #1245: Wasm gas: Stop charging for "empty" instructions

Stebalien:
Motivation: We need to in some way reprice memory instructions to avoid pathological smart contracts that read random memory locations (can cost up to 100gas per read).

Additionally, this patch:

1. Makes consts and basic casts free (fixes #1245).
2. Makes register operations free (locals/globals).
3. Charges 100gas per memory access (10ns). This is based on normal DDR4 memory latency. Includes:
    - Load ops.
    - Bulk memory copy.
    - Indirect jumps/calls.
4. Charges 0.5gas per byte copied (fills and bulk copy). This is based on our previous memcpy benchmarks. This is slightly more than the 0.4gas/byte implied by 3200 speed ram.

As it stands, this patch about doubles the gas of the "simplecoin" EVM contract (so we're back up to where we were before the O3 optimizations).

TODO:

- [ ] More benchmarks #1247.
- [ ] Determine if we can reduce the per-instruction cost. This was an average that included bulk memory instructions, and can probably be dropped to 2 gas per instruction (IIRC).
- [ ] Especially, see if we can reduce the memory initialization fees. It's probably cheaper than memcpy.
- [ ] Consider having a separate fee schedule for everything except the EVM. We can't afford a huge gas hit on other smart contracts, and we're mostly worried about malicious smart contracts.

### DIFF
--- a/fvm/src/gas/price_list.rs
+++ b/fvm/src/gas/price_list.rs
@@ -148,12 +148,18 @@ lazy_static! {
         syscall_cost: Zero::zero(),
         extern_cost: Zero::zero(),
 
-        wasm_rules: WasmGasPrices{
-            exec_instruction_cost: Zero::zero(),
-            memory_fill_per_byte_cost: Zero::zero(),
+        wasm_rules: WasmGasPrices {
+            instruction_default:  Zero::zero(),
+            math_default:  Zero::zero(),
+            math_sqrt:  Zero::zero(),
+            jump_unconditional:  Zero::zero(),
+            jump_conditional:  Zero::zero(),
+            jump_indirect:  Zero::zero(),
+            call:  Zero::zero(),
+            memory_fill_base_cost:  Zero::zero(),
+            memory_fill_per_byte_cost:  Zero::zero(),
+            memory_access_cost:  Zero::zero(),
             memory_copy_per_byte_cost: Zero::zero(),
-            memory_fill_base_cost: Zero::zero(),
-            memory_copy_base_cost: Zero::zero(),
         },
 
         event_emit_base_cost: Zero::zero(),
@@ -290,11 +296,17 @@ lazy_static! {
         extern_cost: Gas::new(21000),
 
         wasm_rules: WasmGasPrices{
-            exec_instruction_cost: Gas::new(4),
+            instruction_default: Gas::new(4),
+            math_default: Gas::new(4),
+            math_sqrt: Gas::new(4),
+            jump_unconditional: Gas::new(4),
+            jump_conditional: Gas::new(4),
+            jump_indirect: Gas::new(4),
+            call: Zero::zero(), // In skyr, we didn't charge for more than the jump.
+            memory_fill_base_cost: Zero::zero(),
             memory_fill_per_byte_cost: Zero::zero(),
+            memory_access_cost: Zero::zero(),
             memory_copy_per_byte_cost: Zero::zero(),
-            memory_fill_base_cost: Gas::zero(),
-            memory_copy_base_cost: Gas::zero(),
         },
 
         event_emit_base_cost: Zero::zero(),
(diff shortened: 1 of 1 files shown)


Stebalien:
>  Especially, see if we can reduce the memory initialization fees. It's probably cheaper than memcpy.

This is pretty critical.

To be precise:

- Right now, SimpeCoin costs 78% more gas.
- If we dropped the memory fill cost to 10gas/byte, we'd go down to 31% more gas.

Of course, this is a very _simple_ contract. We need to see the impact on larger contracts that do more math.

Stebalien:
So, I'm not sure how much we can do here.

3200 speed ram copies data at ~25 bytes per nanosecond, or 2.5 bytes per gas. So we could probably decrease the memcpy gas to 400milligas, but that's about it and that removes our margin of error.

Stebalien:
The memory access charge is a _real_ killer. It doesn't really matter in the "simplecoin" contract, but for the Fibonacci benchmark, it triples the gas charges.

Unfortunately, it doesn't account for the fact that _most_ of these memory accesses will be in cache, because we have no great way to know that.

I can see a few options:

1. Improve the instrumentation so we can detect a load from a constant addresses (hard, may not work, may not be secure).
2. Don't apply these new charges to "system" actors, just the EVM. This will be... annoying, but could make this viable.
3. Implement this in userspace by charging for gas (somehow) inside any memory read operation. Making a syscall is right-out, but we might be able to make something work with gas instrumentation (e.g., call a well-known function and have the instrumentation logic replace that with an explicit gas charge).
4. Do extensive testing and determine that this isn't an issue. But I'm pretty sure we're going to find that it is, in fact, an issue.

Non-options:

1. Require that everything fits in the cache. L1/L2 aren't large enough, and L3 isn't fast enough (honestly, the current gas charge is about right for _L3_ and probably less than what's required for RAM).

vyzo:
I fear we might be doing something profoundly wrong in flat charging all memory accesses as if they are foing through the bus.

The fibo benchmark doesnt even use mstore/mload in its loop, it is all stack accesses.

So we need to find a way to account for those.

Jacarte:
I will just add another small thing here (if it makes sense)

Cranelift is making a lot of optimizations. Therefore, we might be charging for something that in the end might not be executed as it is expected. This case will only overcharge gas.

Jacarte:
Regarding making `const` operations free.
Cranelift can use a lot of `spill/reload` which in the end are memory operations. 

(I think this case is extreme, but bad actors could take advantages ?)

Stebalien:
> Interesting. Am I right in thinking that from a security perspective, memory accesses in this system design can be categorized in two broad categories?

Yes. Right now, this PR charges the same for all "memory" loads. Option 3 from https://github.com/filecoin-project/ref-fvm/pull/1249#issuecomment-1345754723 is about charging only for explicit memory accesses.

If we did anything here, we wouldn't imply a discount to implicit memory access, we'd just apply an explicit charge to explicit access (but we'd need to find a way to do this that doesn't incur syscall overhead.

> We know the system is more efficient under the hood thanks to L1/L2 caches and cache prefetching in ways that we cannot predict transactionally (at instruction site), so we are forced to obviate the effect of such optimizations in our model.

Pretty much, yes.

---

@Jacarte 

> Cranelift is making a lot of optimizations. Therefore, we might be charging for something that in the end might not be executed as it is expected. This case will only overcharge gas.

Absolutely. Unfortunately, we can't rely on cranelift. I.e., we have to assume that someone can use "a reasonable wasm runtime" and leave it at that.

>  Cranelift can use a lot of spill/reload which in the end are memory operations.
> (I think this case is extreme, but bad actors could take advantage of it ?)

For now, I think it's fine. At a minimum, these values will be in the cache.

In the future, we'll need to be careful.

---

@vyzo 

> Also, charging per byte is meaningless, nothing in modern cpus happens at byte granularity.

We generally don't. We charge a fixed fee for loads.

We charge per-byte for bulk memory operations for convenience because charging step-wise fee is overkill (and would probably end up costing even more).

Stebalien:
I'm starting to think that the right solution for now is to:

1. Do charge the bulk memory fees in wasm instrumentation.
2. Don't charge memory latency fees in wasm, but find a way to charge for them in the EVM (likely adding what would effectively be a `charge_gas` _intrinsic_ (a known function call that the instrumentor would lower for us).

fvm/src/gas/price_list.rs
```diff
@@ -1074,56 +1110,277 @@ impl Rules for WasmGasPrices {
...
+            I32ReinterpretF32, I64ReinterpretF64,   // casts
+            F32ReinterpretI32, F64ReinterpretI64,   // casts other way
+            I32Const, I64Const, F32Const, F64Const, // inline constants
+            => free(),
+
+            // Locals (TODO M2.2). Free because these are register references.
+            LocalGet, LocalSet, LocalTee => free(),
+
+            // Globals (TODO M2.2). Free because these are register references.
+            GlobalGet, GlobalSet         => free(),
```
raulk, line 1182: The way this is written, wouldn't this apply the exemption to Skyr and Shark too?

raulk: We probably need a `noop_equivalent` gas param that we can set to 4 in previous network versions.

raulk: @Stebalien -- can you follow up here?

Stebalien: Sorry, I never hit the submit button... See my reply above.

NOte: I'll merge this once I get a chance to merge the syscall gas changes.

Issue #1231, opened 2022-12-08
EVM: Get the ChainID from the FVM


Stebalien:
Currently, we hard-code the chain ID in the build. Instead, we need to retrieve it from the FVM itself by:

1. Adding a chain_id method to the builtin-actors `Runtime`.
2. Changing the CHAINID instruction to return the real chain ID.

Stebalien:
Depends on #1230. Please implement that _first_. Otherwise, we'll end up returning 0 for the chain ID.

Issue #1230, opened 2022-12-08
Lotus/FFI: Plumb the ChainID through from lotus


Stebalien:
1. We need to specify the Chain ID in the lotus "build" config.
2. We need to plumb this through the FFI into the FVM.

PR #1251, opened 2022-12-12
FVM: Move ChainID to shared

arajasek:
#1230 
#1231 

This can't be in FVM if we want built-in actors to actually reference the type (which it doesn't _have_ to, but IMO it should).

### DIFF
--- a/sdk/src/network.rs
+++ b/sdk/src/network.rs
@@ -3,6 +3,7 @@
 use std::convert::TryInto;
 
 use cid::Cid;
+use fvm_shared::chainid::ChainID;
 use fvm_shared::clock::ChainEpoch;
 use fvm_shared::econ::TokenAmount;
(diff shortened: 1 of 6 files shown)


sdk/src/network.rs
```diff
@@ -21,8 +22,11 @@ lazy_static::lazy_static! {
     };
 }
 
-pub fn chain_id() -> u64 {
-    NETWORK_CONTEXT.chain_id
+pub fn chain_id() -> ChainID {
+    NETWORK_CONTEXT
+        .chain_id
+        .try_into()
+        .expect("invalid chain id")
```
Stebalien, line 29: Just `into`. All chain IDs are valid.

Stebalien (approved):

sdk/src/network.rs
arajasek: thanks!

Issue #1147, opened 2022-11-23
GAS: Visualize syscalls gas traces


aakoshh:
Visualize data collected in https://github.com/filecoin-project/ref-fvm/pull/1054

PR #1259, opened 2022-12-12
GAS-1051: Gas time stats and visualization (to master)

aakoshh:
Closes https://github.com/filecoin-project/ref-fvm/issues/1051 , https://github.com/filecoin-project/ref-fvm/issues/1160 and https://github.com/filecoin-project/ref-fvm/issues/1147

This is a rebased and cherry-picked version of these PRs:
* https://github.com/filecoin-project/ref-fvm/pull/1054
* https://github.com/filecoin-project/ref-fvm/pull/1148
* https://github.com/filecoin-project/ref-fvm/pull/1192
* https://github.com/filecoin-project/ref-fvm/pull/1242

See the individual PRs for the description of their content. 

The TL;DR version is that this PR adds:
* Gas time tracing exports to conformance tests with visualization and regressions to check how well the models work on the test vectors. Unfortunately these won't work on master because there aren't test vectors for the latest builtin-actor bundle. The previous experience was also that the data sizes used by these are in a range where the model fit isn't great. 
* A targeted gas calibration contract that we can use to run specific instructions with various inputs we control, to establish the model parameters on a wider domain. 

This PR doesn't modify the prices. That will happen in follow up PRs, focusing on one thing at a time, and using these visualizations and models to see if the new results are reasonable. Ideally we would do these on some canonic environment. 

What would be nice is to have an idea of:
* What should be a representative machine config to run the calibration on?
* Is there a conversion factor we can use to extrapolate IO access time based on the in-memory measurements, based on some existing statistics from Lotus?


Stebalien (approved):
Reviewed as separate PRs, but this rollup looks good to me as well.

Stebalien:
Need to merge https://github.com/filecoin-project/ref-fvm/pull/1270 to get this working.

PR #1261, opened 2022-12-12
feat: explicit gas charges for different instruction types

Stebalien:

Unlike #1249, this patch leaves the actual _charges_ the same, it just puts the framework in place so we can get reduce the diff and decide on values later.


Stebalien:
(I'd need to rework some things to get the same gas charges).

Stebalien:
Ok, I'm:

1. Opening a new PR that doesn't change the gas at all.
2. Then I'll open a final PR to charge for copying.

PR #1270, opened 2022-12-13
fix: avoid default features and disable in CI

Stebalien:
1. Avoid cases where we need to specify default features (except opencl in the fvm crate).
2. Disable default features in CI.

This lets us _reliably_ disable OpenCL. Otherwise, we need OpenCL to build, which isn't usually present in CI.

This also removes the "debug" feature from the SDK, because we always enable that anyways.

### DIFF
--- a/sdk/Cargo.toml
+++ b/sdk/Cargo.toml
@@ -15,12 +15,11 @@ cid = { version = "0.8.5", default-features = false }
 fvm_shared = { version = "3.0.0-alpha.14", path = "../shared" }
 ## num-traits; disabling default features makes it play nice with no_std.
 num-traits = { version = "0.2.14", default-features = false }
-lazy_static = { version = "1.4.0", optional = true }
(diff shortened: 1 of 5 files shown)


Stebalien:
This will let me compile and merge https://github.com/filecoin-project/ref-fvm/pull/1259.

sdk/Cargo.toml
```diff
@@ -15,12 +15,11 @@ cid = { version = "0.8.5", default-features = false }
 fvm_shared = { version = "3.0.0-alpha.14", path = "../shared" }
 ## num-traits; disabling default features makes it play nice with no_std.
 num-traits = { version = "0.2.14", default-features = false }
-lazy_static = { version = "1.4.0", optional = true }
+lazy_static = { version = "1.4.0" }
```
Stebalien, line 18: was not actually optional.

arajasek (approved):

Issue #1257, opened 2022-12-12
Fail to compile when `m2-native` feature is activated


rllola:
Currently when I am trying to build ref-fvm in order to use for integration test I have this error.
```
   --> /home/lola/.cargo/git/checkouts/ref-fvm-aa0150694327c131/b2c07f5/fvm/src/kernel/default.rs:867:14
    |
867 |             .engine()
    |              ^^^^^^ method not found in `&<C as CallManager>::Machine`

For more information about this error, try `rustc --explain E0599
```

If I am remove the feature `m2-native`, I am not able to use the integration module to test the contracts (we are using it for the https://github.com/Zondax/filecoin-solidity-api).

See error log:
```
    failure_info: Some(
        MessageBacktrace(
            Backtrace {
                frames: [
                    Frame {
                        source: 1,
                        method: 3,
                        code: ExitCode {
                            value: 10,
                        },
                        message: "fatal error",
                    },
                    Frame {
                        source: 10,
                        method: 3,
                        code: ExitCode {
                            value: 10,
                        },
                        message: "fatal error",
                    },
                ],
                cause: Some(
                    Fatal {
                        error_msg: "[from=f1d3nehuc4u3l5mn7hazppnogf3oe6l6ymaicbkhi, to=f010, seq=0, m=3, h=0]: unknown import: `actor::install_actor` has not been defined",
                        backtrace: "disabled backtrace",
                    },
                ),
            },
        ),
    ),
```

It looks like recent PRs have changed the way it is working but the CI might not be testing the build for this specific feature. I have investigated back to this changes : https://github.com/filecoin-project/ref-fvm/commit/fde08e02cf127aaf54cf5d2325f671c08948cb59#diff-056ae66290ecf827b5a2ed546e39dd28fb18963218c33bc116c6dbfe83c08b9a

PR #1271, opened 2022-12-13
fix the m2-native feature missing engine getter

rllola:
fix [#1257](https://github.com/filecoin-project/ref-fvm/issues/1257)


Stebalien (approved):
We'll have to consider other fixes later (I generally don't want to expose the engine here). But this should be good enough for now.

PR #1280, opened 2022-12-14
refactor: update the gas schedule

Stebalien:
# Code Changes

1. Remove the old gas schedules. We don't support these releases and we don't need to try to maintain compatibility.
2. Use ScalingCost everywhere to make things simpler.
3. Add clear TODOs for gas prices that need to be figured out.
4. Try to clearly split up operations in a way that should be easier to measure.
5. Rename the "storage gas" field on gas charges to "other gas", and move all charges that aren't charging for _immediate_ computation. This makes benchmarking easier. Includes:
    - Deferred operations (e.g., flush).
    - Memory retention
    - Storage
    
# Syscall Gas Changes

The following changes have been made since M1. This includes reverting some of the changes that are currently on master.

## Signature Validation

- Secp256k1 signature verification has gained a new 10gas/byte cost to reflect the cost of hashing (blake2b).
- BLS signature verification has gained a new 26gas/byte cost to reflect the cost of hashing.

## Hashing

Hashing no longer has a flat cost (was 31355) but has the following costs (per algorithm):

- sha256: 7gas/byte
- blake2b: 10gas/byte
- keccak256: 33gas/byte
- ripemd160: 35gas/byte

For sha256, the "break-even" is ~4.5KiB, for blake2b the break-even is ~3.1KiB. Overall, these changes should result in a gas cost reduction as most things we hash on-chain are tiny.

TODO: sha256 likely assumes hardware support and will need more testing.

## Memory

- Memory copy costs have been reduced from 0.5gas/byte to 0.4gas/byte.
- The "memory retention" cost of 10gas/byte has been split into a 2gas/byte memory _allocation_ cost, and an 8gas/byte memory retention cost. This change does not affect actual gas costs, it just splits the cost into a "compute" cost (allocation) and a non-compute cost (memory retention).

## Storage

Block storage costs have increased by 13.8gas/byte (from 1300gas/byte to 1313.8gas/byte, or 1%):

- PLUS 2 * 2 = 4 gas for an expected 2 allocations (one on write, one on flush).
- PLUS 10gas/byte for hashing.
- MINUS 0.1 * 2 gas/byte for the reduction in memcpy costs.

## Randomness

Randomness now charges for hashing. Specifically, it charges:

1. 1400 for the "extern" call.
2. 10gas/byte of entropy plus 480 gas for hashing the randomness itself.

This is likely still a small under-charge (we're missing some hashing and copying costs), but it brings us closer.

### DIFF
--- a/fvm/src/gas/price_list.rs
+++ b/fvm/src/gas/price_list.rs
@@ -64,374 +64,70 @@ macro_rules! total_enum_map {
     };
 }
 
-/// Help create scaling costs in the format of `m*x+b` (conventional from the linear regression variables).
-macro_rules! scaling_mg {
-    ($m:literal * x + $b:literal) => {
-        ScalingCost::new(Gas::from_milligas($b), Gas::from_milligas($m))
-    };
-}
-
 lazy_static! {
-    static ref OH_SNAP_PRICES: PriceList = PriceList {
-        storage_gas_multiplier: 1300,
-
-        on_chain_message_compute_base: Gas::new(38863),
-        on_chain_message_storage_base: Gas::new(36),
-        on_chain_message_storage_per_byte: Gas::new(1),
-
-        on_chain_return_value_per_byte: Gas::new(1),
-
-        send_base: Gas::new(29233),
-        send_transfer_funds: Gas::new(27500),
-        send_transfer_only_premium: Gas::new(159672),
-        send_invoke_method: Gas::new(-5377),
-
-        create_actor_compute: Gas::new(1108454),
-        create_actor_storage: Gas::new(36 + 40),
-        delete_actor: Gas::new(-(36 + 40)),
-
-        sig_cost: total_enum_map!{
-            SignatureType {
-                Secp256k1 => ScalingCost::new(Gas::new(1637292), Gas::zero()),
-                BLS       => ScalingCost::new(Gas::new(16598605), Gas::zero())
-            }
-        },
-
-        secp256k1_recover_cost: Gas::new(1637292), // TODO measure & revisit this value
-
-        hashing_cost: {
-            let flat = ScalingCost::new(Gas::new(31355), Gas::zero());
-            total_enum_map! {
-                SupportedHashes {
-                    Sha2_256   => flat,
-                    Blake2b256 => flat,
-                    Blake2b512 => flat,
-                    Keccak256  => flat,
-                    Ripemd160  => flat,
-                }
-            }
-        },
-
-        compute_unsealed_sector_cid_base: Gas::new(98647),
-        verify_seal_base: Gas::new(2000), // TODO revisit potential removal of this
-
-        verify_aggregate_seal_base: Zero::zero(),
-        verify_aggregate_seal_per: [
-            (
-                RegisteredSealProof::StackedDRG32GiBV1P1,
-                Gas::new(449900)
-            ),
-            (
-                RegisteredSealProof::StackedDRG64GiBV1P1,
-                Gas::new(359272)
-            )
-        ].iter().copied().collect(),
-        verify_aggregate_seal_steps: [
-            (
-                RegisteredSealProof::StackedDRG32GiBV1P1,
-                StepCost (
-                    vec![
-                        Step{start: 4, cost: Gas::new(103994170)},
-                        Step{start: 7, cost: Gas::new(112356810)},
-                        Step{start: 13, cost: Gas::new(122912610)},
-                        Step{start: 26, cost: Gas::new(137559930)},
-                        Step{start: 52, cost: Gas::new(162039100)},
-                        Step{start: 103, cost: Gas::new(210960780)},
-                        Step{start: 205, cost: Gas::new(318351180)},
-                        Step{start: 410, cost: Gas::new(528274980)},
-                    ]
-                )
-            ),
-            (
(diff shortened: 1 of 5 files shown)


fvm/src/gas/price_list.rs
```diff
@@ -27,33 +28,44 @@ use crate::gas::Gas;
...
-        create_actor_storage: Gas::new(36 + 40),
-        delete_actor: Gas::new(-(36 + 40)),
+        create_actor_storage: Gas::new((36 + 40) * 1300),
+        delete_actor: Gas::new(-(36 + 40)*1300),
 
-        bls_sig_cost: Gas::new(16598605),
+        // TODO(#1279)
+        bls_sig_cost: ScalingCost::fixed(Gas::new(16598605)),
+        // TODO(#1279)
         secp256k1_sig_cost: Gas::new(1637292),
```
Stebalien, line 58: These are importantt.

fvm/src/gas/price_list.rs
```diff
@@ -27,33 +28,44 @@ use crate::gas::Gas;
...
-        secp256k1_recover_cost: Gas::new(1637292), // TODO measure & revisit this value
 
-        hashing_base: Gas::new(31355),
+        // TODO(#1279)
+        hashing: [
+            SupportedHashes::Sha2_256,
+            SupportedHashes::Blake2b256,
+        ].into_iter().map(|h| (h as u64, ScalingCost::fixed(Gas::new(31355)))).collect(),
```
Stebalien, line 64: This is important.

fvm/src/gas/price_list.rs
```diff
@@ -128,309 +140,58 @@ lazy_static! {
...
-        block_create_base: Zero::zero(),
-        block_create_memret_per_byte_cost: Zero::zero(),
-
-        block_read_base: Zero::zero(),
-        block_stat_base: Zero::zero(),
-
-        syscall_cost: Zero::zero(),
-        extern_cost: Zero::zero(),
+        // TODO(#1277): Implement this first before benchmarking.
+        // TODO(#1264): do look at lookback here, but don't bother with the amount of entropy (see the line above).
```
Stebalien, line 144: This is important.

fvm/src/gas/price_list.rs
```diff
@@ -128,309 +140,58 @@ lazy_static! {
...
-            )
-        ].iter()
-        .cloned()
-        .collect(),
+        // TODO(#1279)
+        builtin_actor_manifest_lookup: Zero::zero(),
+        // TODO(#1279)
+        network_context: Zero::zero(),
+        // TODO(#1279)
+        message_context: Zero::zero(),
```
Stebalien, line 192: These are probably less important.

fvm/src/gas/price_list.rs
```diff
@@ -64,217 +64,72 @@ macro_rules! total_enum_map {
...
+        delete_actor: Gas::new(-(36 + 40)*1300),
 
         sig_cost: total_enum_map!{
             SignatureType {
-                Secp256k1 => ScalingCost::new(Gas::new(1637292), Gas::zero()),
-                BLS       => ScalingCost::new(Gas::new(16598605), Gas::zero())
-            }
-        },
-
-        secp256k1_recover_cost: Gas::new(1637292), // TODO measure & revisit this value
```
arajasek, line 100: Should we dropping this TODO here?

arajasek (commented):
So this is a lot of repeated magic numbers, which is _fine_ (the entire list is magic numbers), but I'm wondering if we can just have hard constants for (eg. storage_gas_multiplier, the individual hash costs, etc.)

Now that we only have one pricelist anyway, it feels like it should be pretty okay to just define those as consts here. I think this will be more maintainable if we wanna change these values, and more extensible when we introduce new operations involving hashing etc.

arajasek (approved):
Okay, I've confirmed the numbers are as you say they are in the description (but no thought into the numbers _themselves_).

fvm/src/gas/price_list.rs
Stebalien: Yes. It's the same flat fee (the linear fee is due to hashing).

fvm/src/gas/price_list.rs
```diff
@@ -64,217 +64,72 @@ macro_rules! total_enum_map {
...
-    ($m:literal * x + $b:literal) => {
-        ScalingCost::new(Gas::from_milligas($b), Gas::from_milligas($m))
-    };
-}
-
 lazy_static! {
-    static ref OH_SNAP_PRICES: PriceList = PriceList {
-        storage_gas_multiplier: 1300,
+    static ref HYGGE_PRICES: PriceList = PriceList {
+        // START (Copied from SKYR_PRICES)
```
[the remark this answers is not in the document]
Stebalien: Removed entirely (that used to guard an unchanged section, but that's no longer the case.

Stebalien:
> So this is a lot of repeated magic numbers, which is fine (the entire list is magic numbers), but I'm wondering if we can just have hard constants for (eg. storage_gas_multiplier, the individual hash costs, etc.)

My plan is to try to refactor so we never have that. E.g.:

- Have a separate (generalized) "storage" cost.
- Split hashing and signature verification (likely depends on something to fix our signature verification syscall first).

PR #1290, opened 2022-12-15
feat: only store delegated addresses in the state-tree

### LINKED ISSUE #1289: F4: Only store delegated addresses in the state-tree
We're currently storing all "predictable" addresses, but we've decided to simplify this and only store f4 (delegated) addresses for now.

Stebalien:
This:

- Simplifies the FIP (we can add it to FIP0048 and be done).
- Simplifies the concepts (predictable was nasty).
- Allows us to eventually just assign f4 addresses to everything.

There are three parts to this change:

1. Restore the logic for resolving key addresses when verifying signatures.
2. Mass rename of predictable -> delegated.
3. No longer store f1/f3 in this field.

fixes #1289

### DIFF
--- a/fvm/src/gas/price_list.rs
+++ b/fvm/src/gas/price_list.rs
@@ -950,7 +950,7 @@ impl PriceList {
         GasCharge::new("OnResolveAddress", self.state_read_base, Zero::zero())
     }
 
-    /// Returns the gas required for looking up an actor address.
+    /// Returns the gas required for looking up an actor's delegated address.
     #[inline]
     pub fn on_lookup_delegated_address(&self) -> GasCharge {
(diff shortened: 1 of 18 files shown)


Stebalien:
Based on #1288.

Stebalien:
See https://github.com/filecoin-project/ref-fvm/issues/1297

fvm/src/gas/price_list.rs
```diff
@@ -952,7 +952,7 @@ impl PriceList {
 
     /// Returns the gas required for looking up an actor address.
```
arajasek, line 953:
```suggestion
    /// Returns the gas required for looking up an actor's delegated address.
```

arajasek (approved):

Issue #1279, opened 2022-12-14
GAS: Re-price syscalls based on _only_ FVM costs


Stebalien:
Basically, we need to use the benchmarks in https://github.com/filecoin-project/ref-fvm/pull/1259 to determine the numbers we _would_ assign to the syscalls without any extern costs. Then we need to add in the extern costs after we finish #1264.

Stebalien:
We've repriced:

- Hashing.
- Memory copying.
- Memory allocation.
- Signature verification.

Ongoing:

- #1332
- #1109 
- #1346
- #1347

PR #1291, opened 2022-12-15
GAS-1279: Reprice syscalls for which we have models

aakoshh:
Addresses https://github.com/filecoin-project/ref-fvm/issues/1279

The PR changes the price formula and prices of the syscalls for which we have calibration results in https://github.com/filecoin-project/ref-fvm/pull/1259. The prices are applied in the `HYGGE` version. 

__Caveat__: This is _still_ just my laptop. But the PR shows the relative price differences on the overall charts are [coming together](https://github.com/filecoin-project/ref-fvm/pull/1291/files#diff-e7f726e3f513790f38b00ad5d6b02385f856265b549f0481216d264a37ed9f41), so we just have to do this on an agreed VM at some point to hone them into something we can do repeatedly.

For the record, the calibration is done like so:

```shell
cd testing/calibration
make all
```

Then we can run the following to see what variables we could copy into the prices (but do always look at the charts to check if an intercept should be used or not):

```console
$ make proposals                                                                                                                                                                                                 
{"name":"OnBlockCreate","label":"","base_cost":-58258446,"variable_cost":6526}
{"name":"OnBlockLink","label":"","base_cost":-59014226,"variable_cost":18212}
{"name":"OnBlockOpenPerByte","label":"","base_cost":-1388723,"variable_cost":6545}
{"name":"OnBlockRead","label":"","base_cost":-8130814,"variable_cost":820}
{"name":"OnHashing","label":"Sha2_256","base_cost":-26396178,"variable_cost":63663}
{"name":"OnHashing","label":"Blake2b256","base_cost":-22436,"variable_cost":11450}
{"name":"OnHashing","label":"Blake2b512","base_cost":1485664,"variable_cost":11395}
{"name":"OnHashing","label":"Keccak256","base_cost":-10929332,"variable_cost":48275}
{"name":"OnHashing","label":"Ripemd160","base_cost":12187337,"variable_cost":44114}
{"name":"OnRecoverSecpPublicKey","label":"","base_cost":2680991218,"variable_cost":46}
{"name":"OnVerifySignature","label":"BLS","base_cost":19683263976,"variable_cost":35226}
{"name":"OnVerifySignature","label":"Secp256k1","base_cost":2897640639,"variable_cost":15830}
```

These values are in _milligas_. 

I copied them into the prices in the following format:

```rust
        sig_cost: total_enum_map!{
            SignatureType {
                Secp256k1 =>  scaling_mg!(11400 * x +  2639001000),
                BLS       =>  scaling_mg!(35500 * x + 18719325000),
            }
        },
```

The `scaling_mg!` macro means we'll use a formula like `<variable_cost> * x + <base_cost>`, with both values in milligas (mg). 

The `total_enum_map!` macro is used to make sure I haven't missed any member and won't get a runtime panic because of a missing value.


fvm/src/gas/price_list.rs
```diff
@@ -415,16 +509,19 @@ lazy_static! {
         get_randomness_base: Zero::zero(),
         get_randomness_per_byte: Zero::zero(),
 
-        block_memcpy_per_byte_cost: Gas::from_milligas(500),
-
-        block_open_base: Gas::new(114617),
-        block_open_memret_per_byte_cost: Gas::new(10),
+        block_memcpy_per_byte_cost: Gas::from_milligas(800),
+        block_create_memret_per_byte_cost: Gas::from_milligas(6400),
```
Stebalien, line 513: memret is for memory "retention". I'm guessing that this per-byte cost needs to go somewhere else?

Stebalien: For timing analysis, we should set this to 0.

aakoshh: Thanks for commenting on it. I played around with the prices and eyeballed the formulas for a while to develop my own understanding of what retention might mean, based on the evidence. 

Here's what we started with IIRC:
* `on_block_read`: `block_read_base + block_memcpy_per_byte_cost * data_size`
* `on_block_create`: `block_create_base + (block_memcpy_per_byte_cost + block_create_memret_per_byte_cost) * data_size`

The value of `block_memcpy_per_byte_cost` was 500 and `block_create_memret_per_byte_cost` was 10. 

The first case is very simple with just one price in the formula, which is great because it determines `block_memcpy_per_byte_cost` unambiguously. That came out to be ~800. But then `on_block_create` was heavily under priced on the charts, so there had to be some other cost. Based on earlier experiments `Block::new` takes a long time when it comes from the slice of the memory passed to it, so I figured we can put the remaining of the cost into `block_create_memret_per_byte_cost`. Fore sure it is doing something different than `block_memcpy_per_byte_cost`. 

Then I looked at `on_block_open` which also creates a `Block`, but it does so from a `Vec<u8>` not a slice. What happens there is that the vector is adjusted to have 0 capacity, but there's no copying. 

Its formula was: `(block_open_memret_per_byte_cost + block_memcpy_per_byte_cost) * data_size`. I didn't actually see anything that looked comparable to `block_read` to justify `block_memcpy_per_byte_cost`, so I decided it might be best to remove this term from both `on_block_open` and `on_block_create`, and I thought we might want to live with the interpretation of "retention" as "create a `Block`", in which case the fact that in `block_create` we copy and in `block_open` we don't might be the reason why there are separate `open_memret` and `close_memret` variants. Otherwise I don't know why you would differentiate between the two. 

> For timing analysis, we should set this to 0.

Well, if you set it to zero, that means you treat it as a cost similar to the storage costs: people would have to pay just for data being in memory. And this memory is the host side block registry, not the Wasm memory, which they already paid for. 

That could make sense, although again, I don't know why `open` and `create` would have different prices in that case. It would potentially be more convenient to put it into the `storage_gas`. 

However, that would mean the formulas would be lacking some term to express the time cost. The current formula for `on_block_create` is `block_create_base + block_create_memret_per_byte_cost`. If we set the latter to 0, it becomes flat; I can't just put back `block_memcpy_per_byte_cost` because that is demonstrated to be cheaper by `on_block_read`. So we'd have to come up with something new. 

I actually tried this in https://github.com/filecoin-project/ref-fvm/pull/1291/commits/3a2317d79cef0d5fabaa1537bc423002988d4f59 but then I thought wait, there is a memret for open and close, maybe _that's_ why they are different. 

Maybe you can use these changes and the calibration methods in your PR to come up with new formulas, based on your deeper intuition about what's going on and what were the original intentions?

Stebalien: Ah, got it. Ok, I'll try to fit that cost somewhere else.

Basically, the "memory retention" cost was purely a way to prevent us from running out of memory (which is why I combined them into a single charge in the other PR). Ideally, we'd just track memory of open blocks, but that's more complicated (and has been punted out till M2.2).

aakoshh: Yeah, I thought it's something like that, it was just confusing to see it in `compute_gas`, which I thought relates to time. It would be awkward to have to remember to go back to prices to zero some costs out before we can run a time analysis calibration. Better put them into `storage_gas` then - it's only for the purpose of tracing anyway. And then it could have the same cost as the wasm memory expansion.

Stebalien: Yeah, you're right.

Stebalien: Ok, it looks like this is an _allocation_ cost. That's why we're seeing the same 6.5gas/byte cost show up in multiple places.

Stebalien (approved):
This isn't quite correct, but it's better than what we had so I'm going to merge it then fix it in a followup PR.

PR #1299, opened 2022-12-15
Remove the Cbor trait and its uses

anorth:
This "convenience" trait in the shared libraries introduced additional coupling between actors and FVM libraries, e.g. where an actor object that didn't provide the empty `impl Cbor` block would be unusable in some places.

It's not that convenient, since removing it gives a nice net reduction in code.

The built-in actors no longer use this as of https://github.com/filecoin-project/builtin-actors/pull/960

### DIFF
--- a/ipld/encoding/src/cbor.rs
+++ b/ipld/encoding/src/cbor.rs
@@ -12,6 +12,7 @@ use super::errors::Error;
 use crate::{de, from_slice, ser, strict_bytes, to_vec};
 
 /// Cbor utility functions for serializable objects
+#[deprecated(note = "use to_vec or from_slice directly")]
 pub trait Cbor: ser::Serialize + de::DeserializeOwned {
     /// Marshalls cbor encodable object into cbor bytes
     fn marshal_cbor(&self) -> Result<Vec<u8>, Error> {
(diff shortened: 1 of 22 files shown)


ipld/encoding/src/cbor.rs
```diff
@@ -9,26 +9,10 @@ use std::rc::Rc;
 use serde::{Deserialize, Serialize};
 
 use super::errors::Error;
-use crate::{de, from_slice, ser, strict_bytes, to_vec};
+use crate::{de, from_slice, strict_bytes, to_vec};
 
 pub const DAG_CBOR: u64 = 0x71;
 
-/// Cbor utility functions for serializable objects
-pub trait Cbor: ser::Serialize + de::DeserializeOwned {
```
Stebalien, line 17: Let's deprecate this in the encoding crate, but otherwise leave it (and all the impls in the encoding crate) alone. Otherwise, we need to release:

1. A minor (breaking) version of the endcoding crate.
2. A minor (breaking) version of all the crates that depend on it.

Basically, it makes a bunch of interfaces incompatible.

Alternatively, if you're willing to cut the releases and update everything, go for it. But it's not going to be a quick fix.

Stebalien: Note: definitely remove the implementations from _other_ crates, I'd just leave the encoding crate alone for now (except deprecating it).

Stebalien (approved):

Stebalien:
LGTM, modulo CI.

PR #1305, opened 2022-12-16
CallManager: Restrict embryo creation to the EAM's namespace

arajasek:
This isn't strictly necessary, since there _shouldn't_ be anything wrong with deploying embryos to other ID namespaces. However, only embryos in the EAM's namespace can be deployed at today, so we might as well restrict this for now and not have to think about it.


Stebalien:
https://github.com/filecoin-project/FIPs/pull/574

Stebalien:
LGTM but the integration tests need to be fixed.

arajasek:
You mean i can't just open one-line PRs and assume everything will be fine?

Stebalien (approved):

PR #1330, opened 2022-12-17
fix: IpldBlock::serialize_cbor should return an Option

arajasek:
This matches the behaviour introduced on the v2 FVM in #1199


arajasek:
(approved in #1199)

PR #1331, opened 2022-12-17
Release ipld_encoding 0.3.2

arajasek:
The only change is a tweak of the behaviour of `IpldBlock::serialize_cbor` to return `Option<IpldBlock>` instead of `IpldBlock`


arajasek:
merging to unblock

jennijuju:
One damn minute

PR #1333, opened 2022-12-19
wip: basic send benchmarks

Stebalien:


fvm/src/gas/price_list.rs
```diff
@@ -80,11 +78,8 @@ lazy_static! {
             scale: Gas::new(1300),
         },
 
-        // TODO(#1332)
-        send_base: Gas::new(29233),
-        send_transfer_funds: Gas::new(27500),
-        send_transfer_only_premium: Gas::new(159672),
-        send_invoke_method: Gas::new(-5377),
+        send_transfer_funds: Gas::new(6000),
```
Stebalien, line 81: TODO: this will need to account for actually persisting state, which I'm still working on.

PR #1334, opened 2022-12-19
Split the "read" cache from the "write" cache in the state-tree

### LINKED ISSUE #120: Keep track of "dirty" actors in the state tree
Currently, when we _read_ an actor, we stick it in the state-tree cache. But that means we try to write it back to the blockstore at the end of the block, even if it hasn't changed.

We should add a dirty bit to to the cache so we don't do this.

Stebalien:
This will let us charge the right amount for updating state. Otherwise, we end up making changes to the state-tree even if we only _read_ some actor's state.

fixes #120

### DIFF
--- a/fvm/src/state_tree.rs
+++ b/fvm/src/state_tree.rs
@@ -3,6 +3,7 @@
 // SPDX-License-Identifier: Apache-2.0, MIT
 
 use std::cell::RefCell;
+use std::collections::hash_map::Entry;
 use std::collections::HashMap;
 
 use anyhow::{anyhow, Context as _};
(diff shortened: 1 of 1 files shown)


fvm/src/state_tree.rs
```diff
@@ -31,8 +32,10 @@ pub struct StateTree<S> {
     version: StateTreeVersion,
     info: Option<Cid>,
 
-    /// State cache
+    /// State snapshots.
     snaps: StateSnapshots,
+    /// Read-cache (always check the snapshots first)
+    read_cache: RefCell<HashMap<ActorID, Option<ActorState>>>,
```
raulk, line 38: Improving the docs. Our future selves will appreciate it.

```suggestion
    /// Read-cache for the backing blockstore. This DOES NOT contain
    /// mutations in progress (contained in the snapshots). Always check
    /// snapshots first, then the read cache, then fall back to the backing
    /// blockstore.
    read_cache: RefCell<HashMap<ActorID, Option<ActorState>>>,
```

raulk (approved):

PR #1338, opened 2022-12-19
Explicit gas charges for different instruction types (take 2)

Stebalien:

This doesn't (confirmed with fvm-bench) change the actual _costs_ of any instructions (yet).

### DIFF
--- a/fvm/src/gas/price_list.rs
+++ b/fvm/src/gas/price_list.rs
@@ -250,15 +250,21 @@ lazy_static! {
         install_wasm_per_byte_cost: Zero::zero(),
 
         wasm_rules: WasmGasPrices{
-            exec_instruction_cost: Gas::new(4),
-            // TODO GAS_PARAM
-            memory_fill_per_byte_cost: Zero::zero(),
-            // TODO GAS_PARAM
-            memory_copy_per_byte_cost: Zero::zero(),
-            // TODO GAS_PARAM
+            // Use the default instruction cost of 4 everywhere.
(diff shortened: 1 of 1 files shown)


fvm/src/gas/price_list.rs
```diff
@@ -913,56 +929,281 @@ impl Rules for WasmGasPrices {
...
+            // Charge one instruction for getting a table/memory size.
+            MemorySize, TableSize => fixed(self.instruction_default),
+
+            /******************/
+            /*  Unsupported   */
+            /******************/
+
+            // Exception handling.
+
+            Try, Catch, Throw, Rethrow, CatchAll, Delegate,
```
Stebalien, line 1116: These instructions have never been supported, we're just explicitly handling them now.

fvm/src/gas/price_list.rs
```diff
@@ -250,15 +250,21 @@ lazy_static! {
...
+            // Don't add any additional costs for calls/memory access for now.
+            call: Zero::zero(),
             memory_fill_base_cost: Gas::zero(),
-            // TODO GAS_PARAM
-            memory_copy_base_cost: Gas::zero(),
+            memory_access_cost: Gas::zero(),
+
+            // Don't yet charge anything for copying.
+            memory_copy_per_byte_cost: Gas::zero(),
+            memory_fill_per_byte_cost: Gas::zero(),
```
Stebalien, line 267: I'm going to set these to 400mg in a subsequent patch.

arajasek (approved):

PR #1348, opened 2022-12-20
fix: remove support for non-key addresses from verify_signature

### LINKED ISSUE #1228: Remove address lookup functionality from verify_signature syscall, just do simple verification
The `verify_signature` syscall has functionality to accept an address that is not a public key address (namely an ID address) and lookup that address to find the addressed built-in account actor and then extract its public key address from state, and then verify the signature according to that address. This functionality has existed for a while, originally motivated as an ergonomic improvement to dealing with the multiple addresses. This was changed in #999 to lookup the address stored in the root Actor object in the state tree ([FIP discussion](https://github.com/filecoin-project/FIPs/discussions/477)), and extended in #1002 "Hack: allow f4-embryos to be used as accounts" to hardcode and special-case delegated addresses under the namespace `10`, intended for the EVM Account Manager actor, thus privileging it.

https://github.com/filecoin-project/ref-fvm/blob/c6a153b9f1dbe6b9bfa880a472239bcb508f2285/fvm/src/kernel/default.rs#L444-L464

**This resolution behaviour is incompatible with account abstraction** and should be removed. Under account abstraction, there is no necessary correspondence between an actor's address and its signing authority. This is obviously the case for smart-contract wallets or any other thing-that-might-authenticate-messages deployed at f2 or f4 addresses. Even actors deployed at an f1/f3 address should eventually be able to upgrade into a smart-contract wallet and change their validation logic. Preventing smart wallets is no-no, and implicitly validating messages with against an address/key other than what the user expects could be a huge vulnerability class.

This is easy to fix. The right way for actors to check authentication by other actors is to use the [FIP-0044](https://github.com/filecoin-project/FIPs/blob/master/FIPS/fip-0044.md) standard authentication function. Signature verification is still a necessary primitive, but it should be _just signature verification_, call straight through to the crypto.

There are only two current callers of `verify_signature`: the built-in account actor in its own implementation of `AuthenticateMessage`, and the verified registry which should be changed to use `AuthenticateMessage` https://github.com/filecoin-project/builtin-actors/issues/922

Thus, if we remove this functionality right away, the only impact is that the submission of a FIL+ datacap revocation request must be sure to use the f1/3 address rather than ID address to refer to the verifier. But if we update the verified registry to use FIP-0044 at the same time, there's no other callers. This may require a FIP.

At a minimum I think the EVM hack should be removed immediately (otherwise you'll need to FIP this behaviour change).

@Stebalien @arajasek

Stebalien:
This change doesn't require a FIP because the builtin actors won't call this syscall with anything but a key address.

fixes #1228

### DIFF
--- a/fvm/src/kernel/default.rs
+++ b/fvm/src/kernel/default.rs
@@ -460,65 +460,13 @@ where
                 .on_verify_signature(sig_type, plaintext.len()),
         )?;
 
-        // Resolve to key address before verifying signature.
+        // We only support key addresses (f1/f3). This change does not require a FIP, because no
+        // actors invoke this method with non-key addresses.
         let signing_addr = match signer.payload() {
-            // Already a key address.
             Payload::BLS(_) | Payload::Secp256k1(_) => *signer,
-            // Resolve and re-check.
-            Payload::ID(id) => {
-                // TODO(#1228): remove this logic
-                let act = self
-                    .call_manager
-                    .machine()
-                    .state_tree()
-                    .get_actor(*id)?
-                    .context("state tree doesn't contain actor")
-                    .or_error(ErrorNumber::NotFound)?;
-
-                let is_account = self
-                    .call_manager
-                    .machine()
-                    .builtin_actors()
-                    .is_account_actor(&act.code);
-
-                if !is_account {
-                    // TODO: this is wrong. Maybe some InvalidActor type?
-                    // The argument is syntactically correct, but semantically wrong.
-                    return Err(
-                        syscall_error!(IllegalArgument; "target actor is not an account").into(),
-                    );
-                }
-
-                let _ = self
-                    .call_manager
-                    .charge_gas(self.call_manager.price_list().on_block_open_base())?;
-
-                let state_block = self
-                    .call_manager
-                    .state_tree()
-                    .store()
-                    .get(&act.state)
-                    .context("failed to look up state")
-                    .or_fatal()?
-                    .context("account actor state not found")
-                    .or_fatal()?;
-
-                let _ = self.call_manager.charge_gas(
-                    self.call_manager
-                        .price_list()
-                        .on_block_open_per_byte(state_block.len()),
-                )?;
-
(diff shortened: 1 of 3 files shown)


raulk (commented):
Should update the sdk docs, which we use as the source of truth to document syscalls.

Stebalien:
Good point. I've changed it to return `IllegalArgument` on error.

raulk (commented):
LGTM because it makes operations more atomic, and their costs easier to account for. However, this is ultimately problematic without a `lookup_key_address` syscall (which we used to have) that allows an actor to go from an ID into a key address (if one exists).

Are we 100% sure that built-in actors never pass an ID address through user controlled input? If so, I would be fine recording an issue in the backlog for M2.2 and moving on.

If we have the slightest doubt because user controlled input is involved, I would be very hesitant to merge this patch without that other syscall.

Stebalien:
> However, this is ultimately problematic without a lookup_key_address syscall (which we used to have) that allows an actor to go from an ID into a key address (if one exists).

I don't think we ever had a `lookup_key_address` syscall.

- We have a method on the account actor.
- We used to have a method on the kernel, but it was never exposed as a syscall.

> Are we 100% sure that built-in actors never pass an ID address through user controlled input? If so, I would be fine recording an issue in the backlog for M2.2 and moving on.

This is only called by the account actor with the address stored in the state. Technically, this _can_ be called with an ID address (because some very special accounts don't have proper key addresses), but then this syscall would have failed any which way.

Stebalien:
The code in question is:

https://github.com/filecoin-project/builtin-actors/blob/393d7daeb89fc54286daeb0cf5078fb01f4756d8/actors/account/src/lib.rs#L76-L86

Which pulls the address directly out of the state. This is _usually_ an f1/f3 address. The only exception are special accounts like f99 (actually, that might literally be the only exception).

Stebalien:
(treating that as an LGTM)

anorth:
Thanks!

PR #1351, opened 2022-12-21
feat: refactor memory limits and apply to tables

Stebalien:
Motivation:

1. Explicitly limit table memory. We shouldn't be growing tables, but it's nice to have this check in place.
2. Pave the way for explicitly limiting block memory. We still need to _benchmark_ maximum blockstore memory before we can actually enable a limit, but I'd like to get _a_ limit into M2.1 if possible (instead of relying on gas).

Changes:

First, this refactors memory limits to remove the per-instance limits from the limiter itself (this is enforced by wasmtime) and instead track a global/shared "memory" limit. It then applies this to tables.

Second, this patch removes wasmtime interfaces from the limiter, instead implementing a wrapper to adapt to wasmtime's expected interface. I generally try to avoid exposing wasmtime types.

Finally, this fixes a small bug in the test limiter's minimum memory tracking (to correctly track the minimum non-zero amount allocated by any instance). This patch also tracks the minimum/maximum _table_ sizes.

### DIFF
--- a/fvm/src/engine.rs
+++ b/fvm/src/engine.rs
@@ -560,9 +560,20 @@ impl Engine {
         fn as_wasmtime_limiter<K: Kernel>(
             data: &mut InvocationData<K>,
         ) -> &mut dyn wasmtime::ResourceLimiter {
-            unsafe {
-                &mut *(data.kernel.limiter_mut() as *mut _ as *mut WasmtimeLimiter<K::Limiter>)
-            }
+            // SAFETY: This is safe because WasmtimeLimiter is `repr(transparent)`.
+            // Unfortunately, we can't simply wrap the limiter as we need to return a reference.
+            let limiter: &mut WasmtimeLimiter<K::Limiter> = unsafe {
+                let limiter_ref = data.kernel.limiter_mut();
+                // (debug)-assert that these types have the same layout (guaranteed by
+                // `repr(transparent)`).
+                debug_assert_eq!(
+                    std::alloc::Layout::for_value(&*limiter_ref),
+                    std::alloc::Layout::new::<WasmtimeLimiter<K::Limiter>>()
+                );
+                // Then cast.
+                &mut *(limiter_ref as *mut K::Limiter as *mut WasmtimeLimiter<K::Limiter>)
+            };
+            limiter as &mut dyn wasmtime::ResourceLimiter
         }
 
         store.limiter(as_wasmtime_limiter);
--- a/fvm/src/engine.rs
+++ b/fvm/src/engine.rs
@@ -17,7 +17,7 @@ use wasmtime::{
 };
 
 use crate::gas::{GasTimer, WasmGasPrices};
-use crate::machine::limiter::ExecMemory;
+use crate::machine::limiter::MemoryLimiter;
 use crate::machine::{Machine, NetworkConfig};
 use crate::syscalls::{bind_syscalls, charge_for_init, record_init_time, InvocationData};
 use crate::Kernel;
@@ -539,7 +539,7 @@ impl Engine {
 
     /// Construct a new wasmtime "store" from the given kernel.
     pub fn new_store<K: Kernel>(&self, mut kernel: K) -> wasmtime::Store<InvocationData<K>> {
-        let memory_bytes = kernel.limiter_mut().curr_exec_memory_bytes();
+        let memory_bytes = kernel.limiter_mut().memory_used();
 
         let id = InvocationData {
             kernel,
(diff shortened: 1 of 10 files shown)


fvm/src/machine/limiter.rs
```diff
@@ -1,137 +1,126 @@
...
+        // we charge 8 bytes per table element
+        self.grow_memory(to.saturating_sub(from).saturating_mul(8) as usize)
+    }
 }
 
 /// Limit resources throughout the whole message execution,
 /// across all Wasm instances.
-pub struct ExecResourceLimiter {
-    /// Maximum bytes that a single Wasm instance can use.
-    max_inst_memory_bytes: usize,
```
Stebalien, line 27: The instance limit is enforced by wasmtime anyways. We _could_ enforce it here, but it's tricky when we take tables (and IPLD blocks) into account.

- For instance memory, wasmtime tells us how much memory is allocated when it requests more memory.
- For tables, wasmtime tells us how many elements are already in the table. But this doesn't help us apply an instance _memory_ limit.

So right now, this "instance" limit _only_ applies to the instance's _memory_.

fvm/src/engine.rs
```diff
@@ -557,8 +557,97 @@ impl Engine {
             .expect("failed to create available_gas global");
         store.data_mut().avail_gas_global = gg;
 
-        store.limiter(|id| id.kernel.limiter_mut());
+        fn as_wasmtime_limiter<K: Kernel>(
+            data: &mut InvocationData<K>,
+        ) -> &mut dyn wasmtime::ResourceLimiter {
+            unsafe {
+                &mut *(data.kernel.limiter_mut() as *mut _ as *mut WasmtimeLimiter<K::Limiter>)
+            }
```
aakoshh, line 565: Ugh, I don't know even know what I'm looking at with all this `as *mut _ as *mut` voodoo. 
I reckon if we don't _have_ to use `unsafe` we shouldn't. Isn't there a way to avoid this?

fvm/src/engine.rs
```diff
@@ -557,8 +557,97 @@ impl Engine {
...
+        }
+
+        store.limiter(as_wasmtime_limiter);
 
         store
     }
 }
+
+#[repr(transparent)]
+struct WasmtimeLimiter<L>(L);
```
aakoshh, line 575: Couldn't this be moved into `limiter.rs`? It's quite substantial with its tests to be in `engine.rs` IMO.

It's the same crate, so I don't think an adapter to `wasmtime::ResourceLimiter` cannot be pushed to where we generally deal with limits.

fvm/src/machine/limiter.rs
```diff
@@ -1,137 +1,127 @@
...
 use crate::machine::NetworkConfig;
 
 /// Execution level memory tracking and adjustment.
-pub trait ExecMemory
-where
-    Self: Sized,
-{
+pub trait MemoryLimiter: Sized {
     /// Get a snapshot of the total memory required by the modules on the call stack so far.
-    fn curr_exec_memory_bytes(&self) -> usize;
```
aakoshh, line 13: I included `_bytes` so we know it's not pages for example. Ideally if doesn't say it's bytes, it should return some domain type like `struct Bytes(usize)`. If that's not desirable, at least the docstring could be updated to to say what the measure is.

fvm/src/machine/limiter.rs
```diff
@@ -1,137 +1,127 @@
...
+    fn memory_used(&self) -> usize;
+
+    /// Returns `true` if growing by `bytes` is allowed. Implement this memory to track and limit
+    /// memory usage.
+    fn grow_memory(&mut self, bytes: usize) -> bool;
```
aakoshh, line 13:
```suggestion
    fn grow_memory(&mut self, delta: usize) -> bool;
```

fvm/src/machine/mod.rs
```diff
@@ -125,10 +124,11 @@ pub struct NetworkConfig {
     /// DEFAULT: 512MiB
     pub max_inst_memory_bytes: u64,
 
-    /// Maximum size of memory used during the entire (recursive) message execution.
+    /// Maximum size of memory used during the entire (recursive) message execution. Includes blocks
+    /// and
```
aakoshh, line 128: and table elements?

aakoshh (commented):
The logic looks good to me, and I agree that the instance limits are enforced by the pool :+1: 

I don't really like the `unsafe` magic. I think I understand the motivation not to have to add a wasmtime interface into the `Kernel` trait (I tried to avoid having to add it), but at least that solution was a bit more straight forward, and this part of the interface is sort of internal anyway; nobody should be calling it apart from wasmtime, it's only there for technical reasons. Maybe there's a way to at least not use `unsafe`?

fvm/src/engine.rs
Stebalien: I'll add a comment. The outer struct is defined to be "transparent" so this cast is guaranteed to be safe.

Unfortunately, the store expects a _borrowed_ limiter, so I can't just wrap the internal limiter unless I store the wrapper somewhere. I'm cheating here by just casting because I know the "wrapper" is transparent.

fvm/src/engine.rs
Stebalien: The difference is that _this_ is wasmtime specific, and the rest of the limiter isn't. We've tried to isolate wasmtime specific types as much as possible to:

- Make testing easier.
- Make upgrading wasmtime easier.
- Allow us to eventually support multiple wasm runtimes.

fvm/src/engine.rs
aakoshh: My worry is that these kind of clever tricks will be normalised across the codebase for getting around finding better ways to design the code. It worked before without using unsafe. We didn't like having to have a limiter on kernel in the first place, but it was a necessary evil to make it available where it had to be. Now it feels like adding insult to injury by having code that at least one person on the team (me) wouldn't know how to write or maintain.  

Normally I'd expect unsafe to be used in places like data structures and the Wasm SDK, where they are wrapped so the users of those components don't have to worry about them; they are written by an expert and air gapped. This feels like a more ad-hoc use that is more of a code smell.

Issue #1354, opened 2022-12-21
Ethereum Account abstraction: Refinements


arajasek:
- FVM: Always transform the code CID on first send from an embryo that meets inclusion costs
- FVM: Stop calling the Ethereum Account actor's constructor
- FVM: enforce that sends from embryos have nonce 0
- Client: enfore that sends from embryos have nonce 0

PR #1355, opened 2022-12-21
Executor: Always transform embryo to eth_account if executing message

arajasek:
#1354

### DIFF
--- a/fvm/src/executor/default.rs
+++ b/fvm/src/executor/default.rs
@@ -402,8 +402,7 @@ where
                 .map(|a| matches!(a.payload(), Payload::Delegated(da) if da.namespace() == EAM_ACTOR_ID))
                 .unwrap_or(false) {
             sender_is_valid = true;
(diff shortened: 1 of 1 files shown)


fvm/src/executor/default.rs
```diff
@@ -489,6 +443,8 @@ where
         self.state_tree_mut().mutate_actor(sender_id, |act| {
             act.deduct_funds(&gas_cost)?;
             act.sequence += 1;
+            // In case we've transformed an Embryo into an EthAccount
+            act.code = sender_state.code;
             Ok(())
         })?;
```
Stebalien, line 449: This is a case where it's probably safe to just set instead of mutating. Or, even better, turn this entire block into one big mutation?

Stebalien (approved):
Looks like the correct approach, with one suggestion.

Stebalien (approved):

Issue #1356, opened 2022-12-21
EVM: Optimize KAMT extension encoding


Stebalien:
We're currently serializing our KAMT links as `[cid, {"consumed": u8, "path": [u8, u8, u8] }]`. I.e.:

1. The Extension is being serialized as a CBOR _map_.
2. The path is getting serialized as a list of bytes instead of as a CBOR bytes.
3. We have a list within a list and could save a byte if we just had one list.

PR #1357, opened 2022-12-21
Small KAMT improvements

Stebalien:
This patch contains two small KAMT improvements that make it easier to understand and reason about.

(prep work for #1356)

### DIFF
--- a/ipld/kamt/src/ext.rs
+++ b/ipld/kamt/src/ext.rs
@@ -20,11 +20,6 @@ pub(crate) struct Extension {
     /// and the node the `Link` is pointing to. It might be less than the length of the
     /// slice of the path in the extension, for example if the bit width is 3 and we
     /// consumed only 6 bits out of 8, which is the length of a byte.
-    ///
-    /// The `bit_width` of a KAMT can at most be 8 (`HashBits::next` doesn't work with more),
-    /// resulting in a maximum of 256 slots. The minimum `bit_width` is 1, which would be
-    /// a binary tree. Because the root will take at least 1 bit from the hashed key,
-    /// we know the extension can consume at most 255 bits, which should fit in a `u8`.
     length: u32,
     /// A non-empty part of the `HashedKey` that is covered by the extension.
     /// It could be represented as a vector of indices in the levels of `Node`s
--- a/ipld/kamt/src/ext.rs
+++ b/ipld/kamt/src/ext.rs
@@ -25,7 +25,7 @@ pub(crate) struct Extension {
     /// resulting in a maximum of 256 slots. The minimum `bit_width` is 1, which would be
     /// a binary tree. Because the root will take at least 1 bit from the hashed key,
     /// we know the extension can consume at most 255 bits, which should fit in a `u8`.
-    length: u8,
+    length: u32,
     /// A non-empty part of the `HashedKey` that is covered by the extension.
     /// It could be represented as a vector of indices in the levels of `Node`s
     /// which were skipped, but that could take up more space. And because the
@@ -38,16 +38,16 @@ pub(crate) struct Extension {
 }
 
 impl Extension {
-    pub fn new(length: u8, path: Vec<u8>) -> Self {
+    pub fn new(length: u32, path: Vec<u8>) -> Self {
         Self { length, path }
     }
 
-    pub fn len(&self) -> u8 {
+    pub fn len(&self) -> u32 {
         self.length
     }
 
     pub fn path_bits(&self) -> HashBits {
-        HashBits::new_from_slice(&self.path, self.length as u32)
+        HashBits::new_from_slice(&self.path, self.length)
     }
 
     pub fn is_empty(&self) -> bool {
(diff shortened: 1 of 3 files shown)


ipld/kamt/src/ext.rs
```diff
@@ -16,16 +16,16 @@ use crate::{Error, HashedKey};
...
+    /// and the node the `Link` is pointing to. It might be less than the length of the
+    /// slice of the path in the extension, for example if the bit width is 3 and we
+    /// consumed only 6 bits out of 8, which is the length of a byte.
     ///
     /// The `bit_width` of a KAMT can at most be 8 (`HashBits::next` doesn't work with more),
     /// resulting in a maximum of 256 slots. The minimum `bit_width` is 1, which would be
     /// a binary tree. Because the root will take at least 1 bit from the hashed key,
     /// we know the extension can consume at most 255 bits, which should fit in a `u8`.
-    consumed: u8,
+    length: u32,
```
Stebalien, line 28: Due to alignment, this almost certainly won't use more memory (although I guess we're removing a potential niche for discriminants?).

ipld/kamt/src/ext.rs
```diff
@@ -145,13 +145,13 @@ impl Extension {
...
+    /// far it has been consumed so far, taking the next `length` bits.
+    pub fn from_bits(bits: &mut HashBits, mut length: u32) -> Result<Extension, Error> {
         let mut builder = ExtensionBuilder::new();
-        while consumed > 0 {
-            let i = min(consumed, 8);
+        while length > 0 {
+            let i = min(length, 8);
             let n = bits.next(i as u32)? as u8;
-            consumed -= i;
+            length -= i;
```
Stebalien, line 154: This will probably be _fewer_ instructions in wasm. It may be slower (wall clock) on x86, but that's not too important.

ipld/kamt/src/ext.rs
aakoshh: Using `u32` instead of `u8`? I don't fully understand the alignment; are you saying a single byte is stored as 4 bytes? 

The main reason I used `u8` is because the `Hamt` and the `Kamt` both only support `bit_width` up to 8, which can result in at maximum 256 slots, which is exactly what `u8` can represent. Using a larger data type for indexes and the number of bits consumed is actively misleading IMO.

aakoshh: I didn't want to distrub the `HashBits` to take and `u8` in `next`, but it would be more befitting the domain.

ipld/kamt/src/ext.rs
```diff
@@ -38,36 +38,36 @@ pub(crate) struct Extension {
...
+        while matched < self.length {
             let consumed = hashed_key.consumed;
             let n1 = hashed_key.next(bit_width)?;
             let n2 = path.next(bit_width)?;
             if n1 != n2 {
                 hashed_key.consumed = consumed;
                 break;
             }
-            matched += bit_width as u8;
+            matched += bit_width;
```
aakoshh, line 70: Here too, I never understood why `bit_width` is `u32` when we reject anything above 8.

ipld/kamt/src/ext.rs
aakoshh: Sorry, what will be fewer instructions? `u32`? I can't see any change in the code :thinking:

aakoshh (commented):
This looks fine to me, but can you explain why it's better to use `u32` where an index can only be on the domain of `[0, 255]`? Do you foresee longer hashes than 256 bits, and longer bit widths than 8?

ipld/kamt/src/ext.rs
Stebalien: Yes, u32.

ipld/kamt/src/ext.rs
aakoshh: If you really want to use `u32` here then I think the docstring should explain why, because currently it argues that the data fits into a `u8`, so anyone coming at a later time would be scratching their heads why it's not a `u8` then, or maybe change it back because they don't have intimate knowledge of how many instructions it takes in Wasm to add these numbers up.

Stebalien:
I generally try to avoid using the smallest type that fits the value as it:

1. Generally doesn't help performance.
2. Often makes APIs even _harder_ to use (lots of casts).
3. Is more likely to introduce overflow issues than it is to prevent errors.

I _will_ use, e.g., u8 or u16 when:

1. Size is really important (performance).
2. Using, e.g., u8 lets us skip a bunch of checks because all values are valid. E.g., for lookup tables.

In this particular case, while we don't really need to support hashes larger than 256 bits, there's nothing in the documentation to indicate that that might be the case.

ipld/kamt/src/ext.rs
Stebalien: Removed.

But also, think of it this way: you had to write a long comment on why it was ok to use a u8 here. But using a u32 means we don't even need to talk about it.

ipld/kamt/src/ext.rs
Stebalien: Not really? That just leads to more casting (and potential errors) and won't improve performance.

ipld/kamt/src/ext.rs
Stebalien: If we rejected everything above 256 exactly, I'd use a u8. Really, when I see someone taking a u8 as a value, I generally _assume_ that the entire range is likely valid because the programmer explicitly chose to use a u8 instead of the more normal u32.

Otherwise, the rule of thumb is to just use u32 for small numbers, u64 for large numbers, and usize for offsets.

PR #1364, opened 2022-12-23
KAMT: fix serialization/deserialization

### LINKED ISSUE #1356: EVM: Optimize KAMT extension encoding

Stebalien:
We were using the "struct" (json-like) representation instead of the tuple representation, and we were encoding the extension as a list of bytes (numbers) instead of as a byte-string.

This version:

1. Simplifies the serialization/deserialization logic. We had all the IPLD stuff to because we didn't have a tag. Now that we have a tag, we can use CBOR's built-in support for it.
2. Encodes links as a single tuple of CID, length, and pointer.

Unfortunately, it I couldn't find a simple way to _entirely_ rely on the deriving logic, but I've made it do most of the work.

I'm inlining the extension into the link to save a byte in the encoding.

fixes #1356

### DIFF
--- a/ipld/kamt/src/pointer.rs
+++ b/ipld/kamt/src/pointer.rs
@@ -71,23 +71,23 @@ where
     where
         S: Serializer,
     {
-        use serde::ser::SerializeMap;
-
-        // Using a `Map` for everything to keep things simple.
-        // Constructing the map manually so we don't have to clone the extension and give it to a struct.
-        let mut map = serializer.serialize_map(Some(1))?;
+        #[derive(Serialize)]
+        enum PointerSer<'a, K, V> {
+            #[serde(rename = "v")]
+            Values(&'a [KeyValuePair<K, V>]),
+            #[serde(rename = "l")]
+            Link(&'a Cid, u32, BytesSer<'a>),
+        }
         match self {
-            Pointer::Values(vals) => {
-                map.serialize_entry("v", vals)?;
-            }
+            Pointer::Values(vals) => PointerSer::Values(vals),
             Pointer::Link { cid, ext, .. } => {
-                map.serialize_entry("l", &(cid, ext.len(), BytesSer(ext.path_bytes())))?;
+                PointerSer::Link(cid, ext.len(), BytesSer(ext.path_bytes()))
             }
             Pointer::Dirty { .. } => {
                 return Err(ser::Error::custom("Cannot serialize cached values"))
             }
         }
-        map.end()
+        .serialize(serializer)
     }
 }
 
--- a/ipld/kamt/src/pointer.rs
+++ b/ipld/kamt/src/pointer.rs
@@ -247,3 +247,67 @@ fn unsplit_ext(
 
     Ok(ext)
 }
+
+#[cfg(test)]
+mod test {
+    use std::collections::BTreeMap;
+    use std::fmt::Debug;
+
+    use fvm_ipld_encoding::de::DeserializeOwned;
+    use fvm_ipld_encoding::{from_slice, to_vec, BytesDe};
+    use serde::Serialize;
+
+    use crate::ext::Extension;
+    use crate::pointer::Pointer;
+    use crate::KeyValuePair;
+
+    fn check_encoding<T: DeserializeOwned, V: Serialize>(expected: &T, input: &V)
+    where
+        T: Debug + Eq,
+    {
+        let decoded: T = from_slice(&to_vec(&input).unwrap()).unwrap();
+        assert_eq!(expected, &decoded);
+    }
+
+    #[test]
+    fn test_values_representation() {
+        use crate::id::Identity;
+
+        let v: Pointer<&str, &str, Identity, 32> =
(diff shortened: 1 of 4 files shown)


raulk (commented):
We should probably be having a test here to verify serde behaves correctly.

raulk (commented):
We should probably be having a test here to verify serde behaves correctly.

Stebalien:
Yep, good point. Done.

ipld/kamt/src/pointer.rs
```diff
@@ -79,66 +71,51 @@ where
...
+            Link(Cid, u32, BytesDe),
+        }
+        match PointerDe::<K, V>::deserialize(deserializer)? {
+            PointerDe::Values(vals) => Ok(Self::Values(vals)),
+            PointerDe::Link(k, l, BytesDe(p)) => Ok(Self::Link {
+                cid: k,
+                ext: Extension::new(l, p),
+                cache: Default::default(),
+            }),
+        }
```
aakoshh, line 118: To be honest I liked the way it was better because it was symmetrical: we manually inserted things into a map and then we manually parsed it out using the tags and the types, approximating what I'm guessing the derive macro generates. Now we manually fill the map using the serializer API but use derivation to pull it back out. It begs the question: why not use these DTO structs for serialization as well? I didn't want to use such technical structs to avoid having to allocate them even as temporary structures, although arguably tuples must be allocated too. 

Isn't there a way to use the serde API to look for the tags? Maybe it's just my sense of beauty, but it seems a bit arbitrary this way.

aakoshh (commented):
Thanks, it's good to see how the lower level serde API should be used to directly utilise CBOR. I thought IPLD was the way to go, don't quite know when to use which one. 

It looks correct, but as I commented below I'm not sure it's worth using derivation for half the job.

ipld/kamt/src/pointer.rs
Stebalien: > why not use these DTO structs for serialization as well?

We can, it's just more code. Unfortunately, I couldn't use the _same_ struct due to borrowing/cloning/copying.

> Isn't there a way to use the serde API to look for the tags? Maybe it's just my sense of beauty, but it seems a bit arbitrary this way.

What kind of tags? CBOR? NOTE: the current logic _isn't_ CBOR specific, I'm just avoiding constructing temporary maps.

ipld/kamt/src/pointer.rs
```diff
@@ -293,3 +247,67 @@ fn unsplit_ext(
...
+    use crate::ext::Extension;
+    use crate::pointer::Pointer;
+    use crate::KeyValuePair;
+
+    fn check_encoding<T: DeserializeOwned, V: Serialize>(expected: &T, input: &V)
+    where
+        T: Debug + Eq,
+    {
+        let decoded: T = from_slice(&to_vec(&input).unwrap()).unwrap();
+        assert_eq!(expected, &decoded);
```
aakoshh, line 269: Shouldn't it check the other way around as well?

Stebalien: I've changed this to check that both types converge to the same encoding. That should be safer.

ipld/kamt/src/pointer.rs
Stebalien: I've changed this code to be the same in both cases, avoiding the manual logic.

Stebalien: >  I didn't want to use such technical structs to avoid having to allocate them even as temporary structures, although arguably tuples must be allocated too.

This will involve a _tiny_ bit of memory copying, but not much. Honestly, it should just get optimized away and inlined.

Unfortunately, the _serialize_ side is easy but the _deserialize_ side is more painful to implement manually.

But note: using _maps_ was definitely slower because that actually has to allocate.

ipld/kamt/src/pointer.rs
```diff
@@ -293,3 +247,67 @@ fn unsplit_ext(
...
+    use crate::KeyValuePair;
+
+    fn check_encoding<T, V>(expected: &T, input: &V)
+    where
+        T: Serialize,
+        V: Serialize,
+    {
+        let encoded_expected = to_vec(&expected).unwrap();
+        let encoded_input = to_vec(&input).unwrap();
+        assert_eq!(encoded_expected, encoded_input);
```
aakoshh, line 269: Why not test the decoding part as well?

ipld/kamt/src/pointer.rs
aakoshh: By tags I meant `"v"` and `"l"`, and whatever the opposite of `serialize_map` is.

Now that you have changed it to have two versions of the DTOs it's easier to see why it can't be done with just one :+1:

PR #1366, opened 2022-12-23
SDK: refactor: send should only return exit code and return data

arajasek:
More work towards #987 

`Send` now returns a `Response` comprising of an exitcode and an optional IpldBlock return.

### DIFF
--- a/sdk/src/send.rs
+++ b/sdk/src/send.rs
@@ -1,26 +1,29 @@
 use std::convert::TryInto;
 
 use fvm_ipld_encoding::ipld_block::IpldBlock;
-use fvm_ipld_encoding::RawBytes;
 use fvm_shared::address::Address;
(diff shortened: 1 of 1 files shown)


Stebalien (approved):
`Response` or `Return` are likely reasonable names for a struct, but a tuple is also pretty reasonable.

Stebalien:
So, looking at the other patches, it would be _really_ nice to introduce a named struct here to avoid a bunch of `ret.0` and `ret.1` lookups.

arajasek:
> So, looking at the other patches, it would be _really_ nice to introduce a named struct here to avoid a bunch of `ret.0` and `ret.1` lookups.

Yeah, done, we now have a `Response`

anorth (commented):
LGTM

PR #1368, opened 2022-12-28
test: events: add a test case for out-of-gas.

raulk:
Also moves the setup logic to a helper method.


raulk:
Self-merging since I need a permalink and this is uncontroversial.

Issue #1186, opened 2022-12-01
Kernel::create_actor should be restricted to the Init actor as caller


anorth:
https://github.com/filecoin-project/ref-fvm/blob/8b4f157c09714cd7fd1b89f1a930b3cbec92518a/fvm/src/kernel/default.rs#L771

PR #1373, opened 2023-01-02
Restrict create_actor to the InitActor as caller

maciejwitowski:
Closes https://github.com/filecoin-project/ref-fvm/issues/1186

### DIFF
--- a/fvm/src/kernel/default.rs
+++ b/fvm/src/kernel/default.rs
@@ -32,6 +32,7 @@ use super::*;
 use crate::call_manager::{CallManager, InvocationResult, NO_DATA_BLOCK_ID};
 use crate::externs::{Chain, Consensus, Rand};
 use crate::gas::GasTimer;
+use crate::init_actor::INIT_ACTOR_ID;
 use crate::machine::{MachineContext, NetworkConfig};
 use crate::state_tree::ActorState;
 use crate::syscall_error;
@@ -819,6 +820,15 @@ where
         actor_id: ActorID,
         delegated_address: Option<Address>,
     ) -> Result<()> {
+        if self.caller != INIT_ACTOR_ID {
+            return Err(syscall_error!(
+                Forbidden,
+                "create_actor is restricted to InitActor. Called by {}",
+                self.caller
+            )
+            .into());
+        }
+
         self.call_manager
             .create_actor(code_id, actor_id, delegated_address)
     }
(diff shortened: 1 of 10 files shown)


Stebalien (approved):
❤️

Stebalien:
It looks like this breaks the integration tests. We could try deploying that test actor at the init actor's address, but it may be better to just use the "testing" feature flag. I.e.:

1. Pick some well-known "test" actor ID (less than 100 so it won't be automatically assigned. 98 is a good ID because that was previously assigned to the "chaos" actor (used for testing).
2. Deploy the failing integration testing actor as actor 98.
3. When the "testing" feature is enabled, allow actor 98 to call this syscall.

Alternatively, we could just allow _anyone_ to call this syscall when the testing feature is enabled, although that makes be a _bit_ nervous.

maciejwitowski:
Yep, still working on the failed tests. Thanks @Stebalien for the suggestions!

fvm/src/kernel/default.rs
```diff
@@ -815,6 +816,15 @@ where
         actor_id: ActorID,
         delegated_address: Option<Address>,
     ) -> Result<()> {
+        if self.caller != INIT_ACTOR_ID {
```
raulk, line 819: I think you're looking for `self.actor_id` here, since `caller` would be the ActorID of the actor that immediately called the actor that's currently executing (the _caller_ in this context).

raulk (dismissed):

maciejwitowski:
@Stebalien Tests fixed, please confirm that's the approach you had in mind. Thanks!

maciejwitowski:
One thing which isn't clear to me is that `ExitCode::SYS_ILLEGAL_INSTRUCTION` is returned when I would expect `ExitCode::USR_FORBIDDEN`. Maybe there's something obvious that I'm missing but I wasn't able to debug it...

fvm/src/kernel/default.rs
```diff
@@ -819,6 +823,21 @@ where
         actor_id: ActorID,
         delegated_address: Option<Address>,
     ) -> Result<()> {
+        let allowed_actor = if cfg!(feature = "testing") {
+            TEST_ACTOR_ALLOWED_TO_CALL_CREATE_ACTOR
+        } else {
+            INIT_ACTOR_ID
+        };
```
Stebalien, line 830: When testing, we should allow both.

Stebalien:
> One thing which isn't clear to me is that ExitCode::SYS_ILLEGAL_INSTRUCTION is returned when I would expect ExitCode::USR_FORBIDDEN. Maybe there's something obvious that I'm missing but I wasn't able to debug it...

I assume you were calling `unwrap`? That will exit with `SYS_ILLEGAL_INSTRUCTION`. Unless you've registered the panic hook by calling `sdk::vm::set_panic_handler()` or `sdk::initialize()` in which case it'll exit with `SYS_ASSERTION_FAILED`.

Stebalien (commented):
Mostly LGTM, but there are also some compile issues (warnings).

Stebalien (approved):

PR #1375, opened 2023-01-03
feat: rewrite the state snapshot system to have O(1) lookups

### LINKED ISSUE #1343: FVM: Make state-tree operations O(1) with respect to the call depth
Currently, everything gets more expensive the deeper in the call tree we get but we don't charge for that.

Stebalien:
Instead of keeping true state "layers", we now keep an undo history.

1. When opening a new transaction, we record the current undo history height for both the resolve and state cache.
2. On abort, we walk the undo history backwards, undoing any changes.

This is `O(changes)` on revert, but that's fine. The cost of a potential revert can be charged when the change is made in the first place.

This is strictly better than the current situation where we spend `O(changes)` on every commit, which turns into `O(changes * call_depth)`.

Specifically:

1. Commits are O(1).
2. Actor Insertions/Deletions are O(1).
3. Actor address/state are O(1).
4. Reverts are O(N) in the amount of history being reverted.

I've also removed the `StateSnapshots` abstraction as it was just getting in the way.

fixes #1343

### DIFF
--- a/fvm/src/state_tree.rs
+++ b/fvm/src/state_tree.rs
@@ -663,6 +663,7 @@ mod tests {
     use fvm_shared::{ActorID, IDENTITY_HASH, IPLD_RAW};
     use lazy_static::lazy_static;
 
+    use super::HistoryMap;
     use crate::init_actor;
     use crate::init_actor::INIT_ACTOR_ID;
     use crate::state_tree::{ActorState, StateTree};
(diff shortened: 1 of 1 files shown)


fvm/src/state_tree.rs
```diff
@@ -32,193 +34,111 @@ pub struct StateTree<S> {
...
     /// 2. Creating/discarding a layer simply adds/subtracts from this number
     read_only_layers: u32,
 }
 
-/// State snap shot layer
-#[derive(Debug, Default)]
-struct StateSnapLayer {
-    actors: RefCell<HashMap<ActorID, Option<ActorState>>>,
-    resolve_cache: RefCell<HashMap<Address, ActorID>>,
+/// A map with an "undo" history. All inserts into this map
```
Kubuxu, line 50: Half cut off comment

Kubuxu (approved):
SGWM

raulk (approved):
Very nice.

PR #1378, opened 2023-01-03
Audit fixes

ZenGround0:
From auditing syscalls found a typo and a mismatch between sdk and kernel semantics on a non existent actor when calling balance_of


Stebalien:
The underlying behavior is wrong, this should return NotFound. I believe this was intended to match EVM semantics of "everything magically exists", but that's not how we do things.

Stebalien:
Thanks for catching this!

Stebalien:
(currently working on a fix)

Stebalien:
Fix in https://github.com/filecoin-project/ref-fvm/pull/1383 (with a test this time...).

Issue #1233, opened 2022-12-08
Update cid and multi hash


jennijuju:
#1211 reverted the pkg update, we should introduce them back.

however, given the original committer, forest team has confirmed this update is not too urgent - we will do this in a later release, i.e: post nv18 release

lemmih:
@jennijuju @Stebalien @shawnrader This is becoming an increasingly urgent issue for Forest as we cannot work on NV18 before this can been resolved. Are there any blockers for completing this task by next week? If there _are_ any blockers, let's schedule a meeting to discuss a way forward.

vmx:
It's in the works: https://github.com/filecoin-project/filecoin-ffi/pull/355

lemmih:
> It's in the works: [filecoin-project/filecoin-ffi#355](https://github.com/filecoin-project/filecoin-ffi/pull/355)

That won't work for Forest. We need FVM2 to be updated.

lemmih:
> however, given the original committer, forest team has confirmed this update is not too urgent - we will do this in a later release, i.e: post nv18 release

Oops, I think there's been a miscommunication here. When we talked in early December, I meant it wasn't so urgent that it had to be done before the Christmas break. It definitely has to happen before the nv18 release.

lemmih:
To add some context for those who weren't part of the discussion(s) in December:

- The CID and multihash dependencies have to be updated in FVM2.
- It is a show-stopper for Forest if this doesn't happen.
- Updating FVM2 is trivial (Forest already submitted a patch for doing this) but, due to an unfortunately versioning scheme, it will break a lot of downstream projects (including filecoin-ffi).
- It was agreed that FVM2 would get a new semver-violating minor release with updated CID/multihash dependencies.

shawnrader:
Filecoin-ffi as it is now should not have any issues with updating cid and multihash crates as long as FVM3 and FVM2 are using the same version of those crates.

lemmih:
> Filecoin-ffi as it is now should not have any issues with updating cid and multihash crates as long as FVM3 and FVM2 are using the same version of those crates.

Excellent! This means we're ready to bump the CID/multihash dependencies in FVM2?

jennijuju:
> > Filecoin-ffi as it is now should not have any issues with updating cid and multihash crates as long as FVM3 and FVM2 are using the same version of those crates.
> 
> Excellent! This means we're ready to bump the CID/multihash dependencies in FVM2?

[...]

Issue #1217, opened 2022-12-07
Revert 1211


Stebalien:
We need to revert 1211 and update cid/multihash. Unfortunately, to do this, we need to (in order).

1. [ ] Build abstractions in https://github.com/filecoin-project/filecoin-ffi to abstract over multiple `cid` and `fvm_ipld_blockstore` versions. We currently have absractions over most of the other components, but not these.
    - Specifically, in this case, I'd write a general-purpose `Blockstore2(Blockstore3)` wrapper that wraps a v3 blockstore in a struct that implements the _v2_ interfaces.
3. [ ] Release new minor versions of all IPLD crates in this repo.
4. [ ] Release new "alpha" versions of the `fvm`, `fvm_shared`, and `fvm_sdk` crates.
5. [ ] Update `cid`, `multihash`, `fvm_shared`, and `fvm_sdk` in builtin-actors _next_.
6. [ ] Update all the "v3" crates in the FFI itself.

Stebalien:
@shawnrader: can you implement the first point here? We don't need to upgrade cid/multihash yet to do that. Basically, just introduce an abstraction assuming that:

1. There will be new `fvm_ipld_blockstore` and `fvm_ipld_encoding` releases.
2. That both versions will use different CID types.

Alternatively, you can (in an FVM branch):

1. Revert #1211 (restoring the new CID version).
2. Use `cargo set-version` to _locally_ release new versions (steps 2 and 3 above).
3. Use patches in the FFI to use your local version of the FVM.
4. Then develop the fixes.

This may help because you'll actually be able to test against the FVM as it will exist. If this doesn't make sense, I can also get this started and give you an FVM branch to target.

shawnrader:
I don't have permissions to assign this issue to myself, but I will work on this.

shawnrader:
development branch for filecoin-ffi fvm abstractions: https://github.com/filecoin-project/filecoin-ffi/tree/dependency-cid

Stebalien:
This has been un-reverted in the latest releases.

PR #1380, opened 2023-01-03
chore: upgrade `cid`, `multihash`, `libipld-core` crates

hanabi1224:
Changes:
- upgrade and unify `libipld-core` version
- upgrade `cid`, `multihash` to match their versions in `libipld-core`


Stebalien:
Unfortunately, we can't upgrade cid/multihash for now because they're breaking changes.

See https://github.com/filecoin-project/ref-fvm/issues/1233 and https://github.com/filecoin-project/ref-fvm/issues/1217.

hanabi1224:
@Stebalien  Got it, thx!  Then how about repurposing this PR by only moving the versions of these crates to workspace dependencies to make it easier for the future upgrades?

PR #1382, opened 2023-01-03
chore: migrate to `rust-toolchain.toml`

hanabi1224:
Changes:
- migrate to `rust-toolchain.toml` which is capable of specifying required`components` and `targets` so that contributors won't need to install them manually
- cleanup CI yaml accordingly


Stebalien (approved):
Awesome! Thanks!

PR #1383, opened 2023-01-03
fix: return NotFound from balance_of

Stebalien:
The second half of #949. Found by @ZenGround0 in #1378

fixes #1378


Stebalien:
Note: the EVM already handles this case and returns "0". However, in the FVM, we care about "0" versus "doesn't exist".

ZenGround0 (approved):

PR #1385, opened 2023-01-04
feat: finalize write costs

### LINKED ISSUE #1264: GAS: Determine extern-related costs
We're adjusting some of the system call gas but don't have a good way to estimate the actual cost of writing to and reading from the lotus blockstore. We'd like numbers on lotus related things:

- Cost of flushing per-block and per-byte (assuming batch writes).
- Cost of reading per-block and per-byte (assuming sequential reads).

Basically, we need to:

1. Derive the costs above.
2. Subtract the extern gas that we're already charging (21000).
3. Add those costs to the costs determined in https://github.com/filecoin-project/ref-fvm/pull/1242

All together, that should give us _accurate_ costs.

Stebalien:
From #1264, we've learned that the compute gas for flushing state is
approximately 171963 per block.

I've:

1. Rounded this up to 172k (our measurements aren't that accurate).
2. Added a new "100 byte" (`100 * 1300 = 130k` gas) cost to the storage fee to cover per-block metadata and overhead.

This comes out to ~52k _less_ gas/block than the current per-block fee of 353640. However, when taking all costs into account (hashing, syscall overhead, allocations, memcpy, etc.) we're about even with pre-FVM gas fees for blocks less than 1KiB.

I'm _not_ charging per-byte for compute on flush because we're already charging 1300gas/byte in storage fees. This vastly exceed the per-byte compute cost, a couple of memory copies and allocations that would work out to be < 5gas/byte.

fixes #1264

### DIFF
--- a/fvm/src/gas/price_list.rs
+++ b/fvm/src/gas/price_list.rs
@@ -235,16 +235,15 @@ lazy_static! {
             scale: Gas::zero(),
         },
 
-        block_storage: ScalingCost {
-            flat: Gas::new(353640),
+        block_persist_storage: ScalingCost {
+            flat: Gas::new(130000), // ~ Assume about 100 bytes of metadata per block.
             scale: Gas::new(1300),
         },
 
+        block_persist_compute: Gas::new(172000),
+
         syscall_cost: Gas::new(14000),
         extern_cost: Gas::new(21000),
-
-        // TODO(#1264) (Consider subtracting from the block storage flat fee).
-        block_flush: ScalingCost::zero(),
 
         // TODO(#1279)
         state_read_base: Zero::zero(),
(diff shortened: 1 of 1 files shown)


Stebalien:
> This comes out to ~52k less gas/block than the current per-block fee of 353640. However, when taking all costs into account (hashing, syscall overhead, allocations, memcpy, etc.) we're about even with pre-FVM gas fees for blocks less than 1KiB.

Specifically:

- 2 syscalls (create + link): `14000 * 2`
- 2 (memcpy + alloc): `2 * (2 + .4) * 1024`
- blake2b hashing: `10 * 1024`

Sums to ~43155 gas/block (plus some wasm compute gas).

Kubuxu (approved):
I like the adjustments you've made.

PR #1387, opened 2023-01-04
Added singletons actors in the integration module

rllola:
Integration testing is missing singletons. This PR add them and deploy them automatically when instantiating Tester.


fvm/src/machine/manifest.rs
```diff
@@ -124,13 +140,38 @@ impl Manifest {
...
+            .context("manifest missing verifiedregistry actor")?;
+
+        let datacap_code = *by_name
+            .get(DATA_CAP_ACTOR_NAME)
+            .context("manifest missing datacap actor")?;
+
+        let reward_code = *by_name
+            .get(REWARD_ACTOR_NAME)
+            .context("manifest missing reward actor")?;
+
```
Stebalien, line 162: We should just store the "by name" mapping in the manifest object, and look these up in the tests. This manifest object is really designed for use by the FVM _machine_ itself and shouldn't explicitly have any understanding of other actors.

testing/integration/Cargo.toml
```diff
@@ -30,6 +31,8 @@ serde_tuple = "0.5"
 serde_repr = "0.1"
 thiserror = "1.0.30"
 
+fil_actors_runtime = { git = "https://github.com/filecoin-project/builtin-actors", branch = "next", features = ["m2-native"] }
```
Stebalien, line 34: This will make it impossible to publish this crate.

testing/integration/src/builtin.rs
```diff
@@ -11,20 +12,30 @@ use fvm_shared::ActorID;
 use multihash::Code;
 
 use crate::error::Error::{FailedToLoadManifest, FailedToSetActor, FailedToSetState};
+use crate::{
+    datacap_actor, reward_actor, storagemarket_actor, storagepower_actor, verifiedregistry_actor,
+};
+
+type SingletonCodeCID = (Cid, Cid, Cid, Cid, Cid, Cid, Cid, Cid, Cid, Cid);
```
Stebalien, line 19: ?????????????

testing/integration/src/reward_actor.rs
```diff
@@ -0,0 +1,100 @@
...
+
+    /// 330M for mainnet
+    pub(super) static ref SIMPLE_TOTAL: TokenAmount = TokenAmount::from_whole(330_000_000);
+    /// 770M for mainnet
+    pub(super) static ref BASELINE_TOTAL: TokenAmount = TokenAmount::from_whole(770_000_000);
+}
+
+/// Reward actor state
+#[derive(Serialize_tuple, Deserialize_tuple, Default, Debug, Clone)]
+pub struct State {
```
Stebalien, line 36: Let's use the constructors. I.e.:

1. Create an FVM machine.
2. From the "system" actor, call the constructor on the reward, etc. actor code.

Stebalien (changes requested):

fvm/src/machine/manifest.rs
Stebalien: Eh, we can refactor this part later. The code here is just following the existing conventions.

testing/integration/src/builtin.rs
Stebalien: Ah, why don't we just drop this helper and use the manifest directly? That should simplify things.

PR #1396, opened 2023-01-05
[v2] Add a Log execution event, created with the log syscall

anorth:
I am attempting to improve instrumentation to support analysis of gas charges across spans of code. See #1397.

This PR adds a `Log` execution event to the trace format. A log event is recorded when the `log` syscall is invoked. The log event carries a single `String` message.

A string is the most basic payload to carry here, so I think the right place to start. Some structure may be encoded into strings (for subsequent offline analysis). We could consider adding some kind of key/value pair metadata, but I suggest we defer until observed usage justifies it.

Note that the GasCharge event's `name` is a `Cow<'static, str>`. I presume this is because the same literal name is expected to be used frequently and this type will lead to meaningful optimisation. I'm not sure that that's appropriate for the `Log` event, if it would rule out dynamically constructed messages.

This PR is against the `release/v2` branch, because I am analysing the built-in actors that are not yet using fvm v3. I intend to forward-port it when stable.

### DIFF
--- a/fvm/src/call_manager/mod.rs
+++ b/fvm/src/call_manager/mod.rs
@@ -17,7 +17,7 @@ mod default;
 
 pub use default::DefaultCallManager;
 
-use crate::trace::ExecutionTrace;
+use crate::trace::{ExecutionEvent, ExecutionTrace};
 
 /// BlockID representing nil parameters or return data.
 pub const NO_DATA_BLOCK_ID: u32 = 0;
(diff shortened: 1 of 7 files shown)


fvm/src/call_manager/mod.rs
```diff
@@ -121,6 +121,12 @@ pub trait CallManager: 'static {
         self.gas_tracker_mut().apply_charge(charge)?;
         Ok(())
     }
+
+    /// Whether tracing is requested.
+    fn tracing(&self) -> bool;
```
Stebalien, line 126: Any reason not to just call `call_manager.machine().context().tracing`? It's long, but we only call this once.

Otherwise, I'd provide a default implementation that delegates to `self.machine().context().tracing`.

Stebalien (approved):
Nits but LGTM.

fvm/src/call_manager/mod.rs
anorth: Added the default implementation, thanks. There are additionally 3 internal calls to this, and the many hops to get there adds coupling through those objects at more places.

Stebalien (approved):

PR #1402, opened 2023-01-07
Add custom Debug impl for `IpldBlock` formatting to use hex

mriise:


ipld/encoding/src/ipld_block.rs
```diff
@@ -1,16 +1,30 @@
...
-#[derive(Debug, PartialEq, Eq, Clone, Default)]
+#[derive(PartialEq, Eq, Clone, Default)]
 pub struct IpldBlock {
     pub codec: u64,
     pub data: Vec<u8>,
 }
 
+impl Debug for IpldBlock {
+    fn fmt(&self, f: &mut Formatter<'_>) -> std::fmt::Result {
+        write!(f, "IpldBlock {{ ")?;
```
Stebalien, line 18: Use https://doc.rust-lang.org/std/fmt/struct.Formatter.html#method.debug_struct.

mriise: originally avoided because you cant hex encode in a loop like that without making a buffer which i didnt want. Didn't think hard enough about it though, new solution that doesnt need a container and still uses debug struct :)

Stebalien (approved):

PR #1410, opened 2023-01-10
feat: adjust memory retention fee

Stebalien:
TL;DR: This reduces read/write costs by 0.4.

The point of this fee is to charge at _least_ 10gas/byte. So that's what we now do. Instead of charging an additional fee, we:

1. Compute the actual per-byte "compute" fees.
2. Compute the minimum memory retention fee.
3. Apply a memory retention surcharge up to the minimum memory retention fee.

This seems complex, but the effect is simple: We're now charging 0.4 gas less per byte for reads and writes. I could have just reduced the memory retention cost by 0.4, but the new logic more accurately reflects our intentions with this fee.

### DIFF
--- a/fvm/src/gas/price_list.rs
+++ b/fvm/src/gas/price_list.rs
@@ -424,10 +424,11 @@ pub struct PriceList {
     /// Gas cost per byte allocated (computation cost).
     pub(crate) block_allocate: ScalingCost,
 
-    /// Gas cost for every block retained in memory (read and/or written) to ensure we can't retain
-    /// more than 1GiB of memory while executing a block.
+    /// Minimum gas cost for every block retained in memory (read and/or written) to ensure we can't
+    /// retain more than 1GiB of memory while executing a block.
     ///
-    /// This is applied along with `block_allocate` but kept separate so we can benchmark properly.
+    /// This is just a _minimum_. The final per-byte charge of retaining a block is:
+    /// `min(block_memory_retention.scale, compute_costs)`.
     pub(crate) block_memory_retention: ScalingCost,
 
     /// Gas cost for opening a block.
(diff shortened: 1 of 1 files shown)


fvm/src/gas/price_list.rs
```diff
@@ -219,7 +219,7 @@ lazy_static! {
 
         block_memory_retention: ScalingCost {
             flat: Gas::zero(),
-            scale: Gas::new(8),
+            scale: Gas::new(10),
```
Stebalien, line 222: This looks like an increase, but it's not. This fee now represents the _minimum_ per/byte gas fee, not an additional charge.

raulk: Should we copy the reasoning in the PR body to rustdoc comments on the struct field definition, so that it's in-band with the code?

fvm/src/gas/price_list.rs
```diff
@@ -219,7 +219,7 @@ lazy_static! {
 
         block_memory_retention: ScalingCost {
```
raulk, line 220: Suggest a rename here to convey that this is not an additive cost, but rather a minimum baseline.

raulk (approved):
This looks good.

The 0.4 gas reduction per byte is not immediately obvious to me. Is that an approximation for a particular distribution of block sizes? For smaller block sizes, the fixed costs should dominate... or am I getting it wrong?

Do you happen to have a chart that you can post here? Would be useful to better understand the crossover point.

Another thing: the updated event gas schedule in FIP-0049 reuses the block retention cost, so we'll need to figure out what we do here, since I don't expect to apply any baseline/saturation calculation there, yet we've increased the scale cost to 10. We might need to bifurcate this fee.

Stebalien:
> The 0.4 gas reduction per byte is not immediately obvious to me. Is that an approximation for a particular distribution of block sizes? For smaller block sizes, the fixed costs should dominate... or am I getting it wrong?

Previously, we'd charge a memcpy cost PLUS the memory retention fee. Now, we're just charging the memory retention fee.

PR #1412, opened 2023-01-10
deps: Update wasmtime to v4.0.0

magik6k:
Doesn't build yet because error handling hates me

```
   Compiling fvm v3.0.0-alpha.17 (/home/magik6k/github.com/filecoin-project/ref-fvm/fvm)
error[E0277]: `?` couldn't convert the error to `Abort`
   --> fvm/src/call_manager/default.rs:598:23
    |
598 |                 Ok(res?)
    |                       ^ the trait `From<anyhow::Error>` is not implemented for `Abort`
    |
    = note: the question mark operation (`?`) implicitly performs a conversion on the error value using the `From` trait
    = help: the trait `From<wasmtime::Trap>` is implemented for `Abort`
    = note: required for `std::result::Result<u32, Abort>` to implement `FromResidual<std::result::Result<Infallible, anyhow::Error>>`

error[E0277]: the trait bound `wasmtime::Trap: From<Box<dyn StdError + Sync + std::marker::Send>>` is not satisfied
  --> fvm/src/syscalls/error.rs:56:20
   |
56 |         Trap::from(Box::new(Envelope::wrap(a)) as Box<dyn std::error::Error + Send + Sync + 'static>)
   |         ---------- ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ the trait `From<Box<dyn StdError + Sync + std::marker::Send>>` is not implemented for `wasmtime::Trap`
   |         |
   |         required by a bound introduced by this call
   |
   = help: the trait `From<Abort>` is implemented for `wasmtime::Trap`

error[E0599]: no method named `trap_code` found for enum `wasmtime::Trap` in the current scope
  --> fvm/src/syscalls/error.rs:66:31
   |
66 |         if let Some(code) = t.trap_code() {
   |                               ^^^^^^^^^ method not found in `wasmtime::Trap`

error[E0277]: the trait bound `Abort: StdError` is not satisfied
   --> fvm/src/syscalls/bind.rs:126:53
    |
126 |                         charge_for_exec(&mut caller)?;
    |                                                     ^ the trait `StdError` is not implemented for `Abort`
...
196 | impl_bind_syscalls!();
    | --------------------- in this macro invocation
    |
    = help: the following other types implement trait `FromResidual<R>`:
              <std::result::Result<T, F> as FromResidual<Yeet<E>>>
              <std::result::Result<T, F> as FromResidual<std::result::Result<Infallible, E>>>
    = note: required for `anyhow::Error` to implement `From<Abort>`
    = note: required for `std::result::Result<u32, anyhow::Error>` to implement `FromResidual<std::result::Result<Infallible, Abort>>`
    = note: this error originates in the macro `impl_bind_syscalls` (in Nightly builds, run with -Z macro-backtrace for more info)


```

### DIFF
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -596,11 +596,7 @@ where
                 // detected it and returned OutOfGas above. Any other invocation failure is returned
                 // here as an Abort
 
-                Ok(res.map_err(|e: anyhow::Error|match e.anyhow_kind() {
-                    // translate  anyhow error we get into an abort
-                    wasmtime::Trap => e,
-                    Abort => e,
-                })?)
+                Ok(res?)
             })();
 
             let invocation_data = store.into_data();
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -535,35 +535,14 @@ where
 
             // From this point on, there are no more syscall errors, only aborts.
             let result: std::result::Result<BlockId, Abort> = (|| {
-                use wasmtime_runtime::InstantiationError;
                 // Instantiate the module.
                 let instance = engine
                     .get_instance(&mut store, &state.code)
                     .and_then(|i| i.context("actor code not found"))
-                    .map_err(|e| match e.downcast::<InstantiationError>() {
-                        Ok(e) => match e {
-                            // This will be handled in validation.
-                            InstantiationError::Link(e) => Abort::Fatal(anyhow!(e)),
-                            // TODO: We may want a separate OOM exit code? However, normal ooms will usually exit with SYS_ILLEGAL_INSTRUCTION.
-                            InstantiationError::Resource(e) => Abort::Exit(
-                                ExitCode::SYS_ILLEGAL_INSTRUCTION,
-                                e.to_string(),
-                                NO_DATA_BLOCK_ID,
-                            ),
-                            // TODO: we probably shouldn't hit this unless we're running code? We
-                            // should check if we can "validate away" this case.
-                            InstantiationError::Trap(e) => Abort::Exit(
-                                ExitCode::SYS_ILLEGAL_INSTRUCTION,
-                                format!("actor initialization failed: {:?}", e),
-                                0,
-                            ),
-                            // TODO: Consider using the instance limit instead of an explicit stack depth?
-                            InstantiationError::Limit(limit) => Abort::Fatal(anyhow!(
-                                "did not expect to hit wasmtime instance limit: {}",
-                                limit
-                            )),
-                        },
-                        Err(e) => Abort::Fatal(e),
+                    .map_err(|e| {
+                        // todo try to distinguish resource limit / trap / fatal
+                        log::error!("failed to instantiate actor: {:?}", e);
+                        Abort::Fatal(e)
                     })?;
 
                 // Resolve and store a reference to the exported memory.
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -595,7 +595,12 @@ where
                 // If the invocation failed due to running out of exec_units, we have already
                 // detected it and returned OutOfGas above. Any other invocation failure is returned
                 // here as an Abort
-                Ok(res?)
+
+                Ok(res.map_err(|e: anyhow::Error|match e.anyhow_kind() {
+                    // translate  anyhow error we get into an abort
+                    wasmtime::Trap => e,
+                    Abort => e,
+                })?)
             })();
 
             let invocation_data = store.into_data();
--- a/fvm/src/engine.rs
+++ b/fvm/src/engine.rs
@@ -171,13 +171,15 @@ fn wasmtime_config(ec: &EngineConfig) -> anyhow::Result<wasmtime::Config> {
     c.generate_address_map(false);
     c.cranelift_debug_verifier(false);
     c.native_unwind_info(false);
-    #[allow(deprecated)] // TODO https://github.com/bytecodealliance/wasmtime/issues/5037
-    c.wasm_backtrace(false);
     c.wasm_reference_types(false);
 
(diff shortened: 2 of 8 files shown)


magik6k:
Update - The last commit reduces the number of errors which appear in the build output, but I can't tell if this is because things got more or less broken

Most of this mess is caused by changes in this wasmtime commit - https://github.com/bytecodealliance/wasmtime/commit/2afaac5181f4b73e86fac39d095c84a9b8e59129

Essentially wasmtime Traps are now a simple enum, but it's now possible to return any error types we want from syscalls and, IIUC, we just get them back from `instance.get_typed_func([..])[..].call()`.

So in theory:
* We want to make Abort be a normal std (or anyhow?) error
* We don't need the hacks for wrapping Abort into/from Traps, we just need to translate 'normal' Traps into Aborts
* Then we need to do a small bit of translating the (anyhow?) Error we get from wasm call into Abort we expect in the part of CallManager which makes that call

fvm/src/call_manager/default.rs
```diff
@@ -595,7 +595,12 @@ where
                 // If the invocation failed due to running out of exec_units, we have already
                 // detected it and returned OutOfGas above. Any other invocation failure is returned
                 // here as an Abort
-                Ok(res?)
+
+                Ok(res.map_err(|e: anyhow::Error|match e.anyhow_kind() {
+                    // translate  anyhow error we get into an abort
+                    wasmtime::Trap => e,
+                    Abort => e,
+                })?)
```
magik6k, line 603: This is trying to translate from the anyhow error we get from wasm call into an abort, tho I'm not quite able to find a good way to do that

fvm/src/syscalls/error.rs
```diff
@@ -50,13 +51,33 @@ impl Abort {
...
+        let desc = match self {
+            Exit(_, _, _) => "exit",
+            OutOfGas => "out of gas",
+            Fatal(_) => "fatal error",
+        };
+        write!(f, "wasm trap: {desc}")
+    }
+}
+
+impl std::error::Error for Abort {
```
magik6k, line 68: Is there a better way to do that? Or at least a way to get rid of the manual Display impl above?

fvm/src/call_manager/default.rs
Stebalien: Ah, you need `downcast`.

Stebalien:
Ok, I _think_ I fixed it. We'll see if the tests pass.

Stebalien:
Ok, so, one error down but... I have no idea how to detect "out of memory" on initialization. We used to be able to do this, but it looks like the error types just disappeared. Options are:

1. Have some form of "ran out of memory" flag that we record inside the limiter and/or kernel?
2. Somehow determine if we're likely to run out of memory.
3. Treat all instantiation failures the same (e.g., as ILLEGAL_INSTRUCTION).

Honestly, the last one is probably the simplest. It's not great, but...


We should probably write a test to see what error wasmtime gives us because there may be some error type that I'm missing.

fvm/src/call_manager/default.rs
```diff
@@ -535,35 +535,14 @@ where
...
-                            ),
-                            // TODO: Consider using the instance limit instead of an explicit stack depth?
-                            InstantiationError::Limit(limit) => Abort::Fatal(anyhow!(
-                                "did not expect to hit wasmtime instance limit: {}",
-                                limit
-                            )),
-                        },
-                        Err(e) => Abort::Fatal(e),
+                    .map_err(|e| {
+                        // todo try to distinguish resource limit / trap / fatal
```
magik6k, line 543: Todo

fvm/src/engine.rs
```diff
@@ -170,14 +171,16 @@ fn wasmtime_config(ec: &EngineConfig) -> anyhow::Result<wasmtime::Config> {
...
-    c.wasm_backtrace(false);
     c.wasm_reference_types(false);
 
     // Reiterate some defaults
     c.guard_before_linear_memory(true);
     c.parallel_compilation(true);
 
+    #[allow(deprecated)] // TODO https://github.com/bytecodealliance/wasmtime/issues/5037
+    // has to be enabled, otherwise returning errors from syscalls will not work
+    c.wasm_backtrace(true);
```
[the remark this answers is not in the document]
magik6k: user errors when returned have needs_backtrace set to true:

* https://github.com/bytecodealliance/wasmtime/blob/main/crates/runtime/src/traphandlers.rs#L160-L164
* https://github.com/bytecodealliance/wasmtime/blob/main/crates/runtime/src/traphandlers.rs#L181-L185

With seemingly no way to tell wasmtime that we really don't care about that backtrace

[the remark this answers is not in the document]
magik6k: Unfortunately enabling backtraces makes native_stack_overflow take 13 minutes on my machine.

The native_stack_overflow test is invoking an actor which:

* Makes 396 recursive calls inside wasm to fill the entire wasm stack limit
* Then calls itself recursively until we hit recursive call limit

This test was written to ensure that in the most extreme abuse wasm actors can't exhaust native stacks, as that fault can't be caught / handled safely in any way. So why is this so slow? Didn't wasmtime fix backtrace re-capture in 3.0.0 (https://github.com/bytecodealliance/wasmtime/issues/5037)? They did.. but when we return from a failed actor send syscall, we're returning a new user error, for which wasmtime will re-capture backtrace again..Ideally we'd have a way to tell wasmtime that we don't want to capture a backtrace when returning from syscalls, but I don't see a way today

magik6k:
Wasmtime issue about backtrace error bug: https://github.com/bytecodealliance/wasmtime/issues/5577

Stebalien:
Replaced by https://github.com/filecoin-project/ref-fvm/pull/1672

PR #1445, opened 2023-01-12
Refactor: Move Response from SDK to shared

arajasek:
As in the title


Stebalien (approved):
I'd just put this in the root, honestly. But any which way is fine.

arajasek:
> I'd just put this in the root, honestly. But any which way is fine.

Sure, works for me. Making the change.

PR #1487, opened 2023-01-17
fix: correctly return out of gas errors when instantiating

### LINKED ISSUE #1486: EVM: Unexpected behavior of recursive DelegateCall
found by Mikers: https://filecoinproject.slack.com/archives/C029MT4PQB1/p1673891621728659

https://github.com/filecoin-project/lotus/blob/05d7d4b2a6179c1280048a2250b486c200d6fbcc/itests/fevm_test.go#L218 for some context. Primary causeTBD

Stebalien:
We were implicitly converting them to "anyhow" errors, then treating them as fatal.

fixes #1486

### DIFF
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -535,36 +535,11 @@ where
 
             // From this point on, there are no more syscall errors, only aborts.
             let result: std::result::Result<BlockId, Abort> = (|| {
-                use wasmtime_runtime::InstantiationError;
                 // Instantiate the module.
                 let instance = engine
-                    .get_instance(&mut store, &state.code)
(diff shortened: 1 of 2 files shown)


Stebalien:
We should test with https://github.com/filecoin-project/ref-fvm/issues/1489.

Stebalien:
@mriise can you manually test this fix for now?

mriise:
We no longer panic with this fix.

mriise (approved):
LGTM, Error handling around here is a bit sketchy IMO since there are so many `?` and `map_err`. Future improvement may be to add some custom error types to convert into/out of so we are explicit about what went wrong and how that failure case is converted/cast

PR #1493, opened 2023-01-17
Integrate fvm-bench and the basics of a testkit

vyzo:
This moves the fvm-bench tool in tree, and refactors it to create a basic testkit for writing fevm integration tests.

Note that the move preserves history of the original tool.


tools/testkit/src/fevm.rs
```diff
@@ -0,0 +1,109 @@
...
+#[allow(dead_code)]
+#[repr(u64)]
+pub enum EVMMethod {
+    Constructor = METHOD_CONSTRUCTOR,
+    Resurrect = 2,
+    GetBytecode = 3,
+    GetBytecodeHash = 4,
+    GetStorageAt = 5,
+    InvokeContractDelegate = 6,
+    // it is very unfortunate but this hasher creates a circular dependency and is useless in
```
arajasek, line 87: Please drop comments like this before merging.

aakoshh:
Maybe the contracts could be left out, until there are actual tests that use them.

tools/testkit/src/fevm.rs
vyzo: I have seen this,and I disagree with removing the reserved method.
Anyway, we have to use the magic number here.

aakoshh:
Do these crates need to have their own rust tool chain?

vyzo:
we still need to house them somewhere, and we use those already for basic benchmarks, so I stashed them in the `contracts/benchmarks` directory for now.
We can certainly move them around later.

vyzo:
> Do these crates need to have their own rust tool chain?

Probably not, i will remove.

tools/contracts/benchmarks/jmpfest-gen.ss
```diff
@@ -0,0 +1,58 @@
+#!/usr/bin/env gxi
```
raulk, line 1: I don't know how to maintain this, and I don't have gxi installed to run this. Can you please port to Rust -- the language of this repo?

tools/fvm-bench/README.md
```diff
@@ -0,0 +1,40 @@
+# fvm-bench
+Tools for testing and benchmarking FVM
+
+## Status
+
+Currently this is a program called `fvm-bench`, which allows you to execute and gas-benchmark
+fevm contracts.
+
+This is a barebones MVP, it requires you to have checked out ref-fvm in a sibling directory
+in order to build.
```
raulk, line 10: Is this still the case?

vyzo: no, this readme is obsolete.

tools/contracts/benchmarks/jmpfest-gen.ss
vyzo: no to porting, i am not going to waste time on this. 
Its job is done, it generated the contracts; it is here as a reference and in case we need it in the future.

tools/fvm-bench/README.md
vyzo: updated the readme.

tools/testkit/src/lib.rs
```diff
@@ -0,0 +1,79 @@
...
+        }
+    };
+    Tester::new(
+        NetworkVersion::V18,
+        StateTreeVersion::V5,
+        bundle_cid,
+        blockstore,
+    )
+    .unwrap_or_else(|what| {
+        exit_with_error(format!("error creating execution framework: {}", what));
```
aakoshh, line 46: It's surprising to see `sys::exit` in a library function.

tools/testkit/src/fevm.rs
```diff
@@ -0,0 +1,111 @@
...
+    Constructor = METHOD_CONSTRUCTOR,
+    Resurrect = 2,
+    GetBytecode = 3,
+    GetBytecodeHash = 4,
+    GetStorageAt = 5,
+    InvokeContractDelegate = 6,
+    // it is very unfortunate but the hasher creates a circular dependency, so we use the raw
+    // number.
+    //InvokeContract = frc42_dispatch::method_hash!("InvokeEVM"),
+    InvokeContract = 3844450837,
```
aakoshh, line 92: Just curious: why is this method using a hash, when the others are just numbers?

aakoshh:
> we still need to house them somewhere, and we use those already for basic benchmarks

I get it, I just don't see any automation for how these contracts were built, how they can be rebuilt, and how they are used in benchmarks. From what I understand, you run them manually with the `fevm-bench` CLI, with hexadecimal parameters like [this](https://github.com/filecoin-project/ref-fvm/pull/1493/files#diff-28ca9679dbd43166bf9f99efb7bdc18e9c98b0ba0ad34ab2d29f759c1aa28c85R38). Without reproducibility the value of dumping these files into the main repo is I think questionable. It would be better to introduce the tool, then open individual PRs for the contracts with examples on how to use them. 

If I missed the instructions on using them I'm sorry, it's a large PR.

tools/fvm-bench/LICENSE
```diff
@@ -0,0 +1,21 @@
+MIT License
+
+Copyright (c) 2022 Filecoin
```
Stebalien, line 3: Protocol Labs

tools/contracts/benchmarks/jmpfest-gen.ss
Stebalien: Let's add a comment explaining that this was used to generate the jumpfest contract.

tools/testkit/src/fevm.rs
```diff
@@ -0,0 +1,111 @@
...
+        params: RawBytes::serialize(BytesSer(contract)).unwrap(),
+        sequence: owner.seqno,
+        ..Message::default()
+    };
+    let create_mlen = create_msg.params.len();
+
+    let create_res = tester
+        .executor
+        .as_mut()
+        .unwrap()
```
aakoshh, line 30:
```suggestion
        .expect("executor not initialized")
```

tools/testkit/src/fevm.rs
```diff
@@ -0,0 +1,111 @@
...
+        method_num: EVMMethod::InvokeContract as u64,
+        params: RawBytes::serialize(BytesSer(input_data)).unwrap(),
+        ..Message::default()
+    };
+    let invoke_mlen = invoke_msg.params.len();
+
+    let invoke_res = tester
+        .executor
+        .as_mut()
+        .unwrap()
```
aakoshh, line 59: .expect("executor not initialized")

tools/contracts/benchmarks/jmpfest-gen.ss
vyzo: ok.

tools/testkit/src/lib.rs
vyzo: yeah, I can remove it and just panic I guess, depending on how strongly you feel on this guys.
Not really frettting however.

tools/testkit/src/fevm.rs
```diff
@@ -0,0 +1,111 @@
...
+    #[serde(with = "strict_bytes")]
+    pub initcode: Vec<u8>,
+    #[serde(with = "strict_bytes")]
+    pub salt: [u8; 32],
+}
+
+#[derive(Serialize_tuple, Deserialize_tuple)]
+pub struct CreateReturn {
+    pub actor_id: ActorID,
+    pub robust_address: Address,
```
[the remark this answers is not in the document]
vyzo: oops, fixing.

tools/testkit/src/fevm.rs
vyzo: i think we still need the unwrap, no?

tools/fvm-bench/LICENSE
vyzo: fixed.

tools/testkit/src/fevm.rs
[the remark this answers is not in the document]
vyzo: fixed.

tools/testkit/src/lib.rs
vyzo: turned to panics.

tools/contracts/benchmarks/jmpfest-gen.ss
vyzo: done.

tools/testkit/src/fevm.rs
vyzo: We should probably fix the offending crate to be freestanding, it is very annoying that we can't actually use it inside ref-fvm.

vyzo:
merged master and resolved conflicts

tools/testkit/src/fevm.rs
Stebalien: Expect is the same thing, it just gives a nicer error message.

tools/testkit/src/lib.rs
```diff
@@ -0,0 +1,74 @@
+// Copyright 2021-2023 Protocol Labs
+// SPDX-License-Identifier: Apache-2.0, MIT
+
+#![allow(dead_code)]
```
Stebalien: Please don't do this. If you think you need something like this, you need a comment in the code explaining _why_.

tools/testkit/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+    pub events: bool,
+}
+
+pub type BasicTester = Tester<MemoryBlockstore, DummyExterns>;
+pub struct Account {
+    pub account: TAccount,
+    pub seqno: u64,
+}
+
+pub fn new_tester(bundle_path: String) -> BasicTester {
```
Stebalien, line 28: Make this a method on the testing framework?

tools/testkit/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+        StateTreeVersion::V5,
+        bundle_cid,
+        blockstore,
+    )
+    .unwrap_or_else(|what| {
+        panic!("error creating execution framework: {}", what);
+    })
+}
+
+pub fn create_account(tester: &mut BasicTester) -> Account {
```
Stebalien, line 47: Also make this a method?

tools/testkit/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+    let accounts: [TAccount; 1] = tester.create_accounts().unwrap();
+    Account {
+        account: accounts[0],
+        seqno: 0,
+    }
+}
+
+pub fn create_accounts<const N: usize>(tester: &mut BasicTester) -> [Account; N] {
```
Stebalien, line 55: Really, why do we have a any of these.

tools/testkit/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+    let accounts: [TAccount; N] = tester.create_accounts().unwrap();
+    accounts.map(|a| Account {
+        account: a,
+        seqno: 0,
+    })
+}
+
+pub fn prepare_execution(tester: &mut BasicTester, options: &ExecutionOptions) {
```
Stebalien, line 63: What does this do? This doesn't seem like something that should be public.

(I'm not asking you to _tell_ me what this does, I'm pointing out that we have a super-generic public method called `prepare_execution` with no documentation that likely belongs on the test framework itself.)

tools/testkit/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+        }
+    };
+    Tester::new(
+        NetworkVersion::V18,
+        StateTreeVersion::V5,
+        bundle_cid,
+        blockstore,
+    )
+    .unwrap_or_else(|what| {
+        panic!("error creating execution framework: {}", what);
```
Stebalien, line 43: Why are we not just returning an error?

Stebalien (changes requested):
Let's remember that this is no longer a random testing tool, but actual code we intend to maintain. This is not code I want to maintain, and I'm not going to add it to the pile of "things Steven will fix later".

tools/testkit/src/fevm.rs
```diff
@@ -0,0 +1,99 @@
...
+use fvm_shared::address::Address;
+use fvm_shared::message::Message;
+use fvm_shared::{ActorID, METHOD_CONSTRUCTOR};
+
+use crate::{Account, BasicTester};
+
+pub fn create_contract(tester: &mut BasicTester, owner: &mut Account, contract: &[u8]) -> ApplyRet {
+    let create_msg = Message {
+        from: owner.account.1,
+        to: Address::new_id(10), // EAM
```
Stebalien, line 16: Write a constant, users will need it.

tools/testkit/src/fevm.rs
```diff
@@ -0,0 +1,99 @@
...
+        gas_limit: 10_000_000_000,
```
Stebalien, line 17: Constant or parameter.

Stebalien (changes requested):

tools/testkit/src/fevm.rs
```diff
@@ -0,0 +1,99 @@
...
+    let create_mlen = create_msg.params.len();
+
+    let create_res = tester
+        .executor
+        .as_mut()
+        .unwrap()
+        .execute_message(create_msg, ApplyKind::Explicit, create_mlen)
+        .unwrap();
+
+    owner.seqno += 1;
```
aakoshh, line 32: Does this always need to be incremented? I use this method to query the balance in SimpleCoin, which is a `view`, so it can be run locally without sending a transaction. Shouldn't the nonce stay the same?

tools/testkit/src/fevm.rs
```diff
@@ -0,0 +1,99 @@
...
+    let invoke_mlen = invoke_msg.params.len();
+
+    let invoke_res = tester
+        .executor
+        .as_mut()
+        .unwrap()
+        .execute_message(invoke_msg, ApplyKind::Explicit, invoke_mlen)
+        .unwrap();
+
+    src.seqno += 1;
```
aakoshh, line 61: Does this always need to be incremented? I use this method to query the balance in SimpleCoin, which is a `view`, so it can be run locally without sending a transaction. Shouldn't the nonce stay the same?

tools/testkit/src/lib.rs
aakoshh: FWIW I moved it [inside my test driver](https://github.com/filecoin-project/ref-fvm/pull/1507/files#diff-0b38f777658656ae3e7df7b053d6daa49846b1af6f559bfd60c726b0d58cd904R138). Turns out I didn't understand when exactly we should call this.

vyzo:
Sure, let's make it better and something that will not be a maintenance hazard.

A couple basic questions:
- should we move the testkit inside the integration crate?
- should we make a higher level kit object that encloses the tester, accounts, and their nonces?

tools/testkit/src/lib.rs
```diff
@@ -0,0 +1,71 @@
...
+    })
+}
+
+pub fn create_account(tester: &mut BasicTester) -> Account {
+    let accounts: [TAccount; 1] = tester.create_accounts().unwrap();
```
aakoshh, line 48: I have a similar approach to creating multiple accounts and just noticed that while the accounts have different actor ID, they had the same Secp256k1 key.

aakoshh: Fixed in https://github.com/filecoin-project/ref-fvm/pull/1507/commits/d0203c0d1f70342651c228fd5b658d58f8eab536

tools/testkit/src/fevm.rs
```diff
@@ -0,0 +1,99 @@
...
+use fvm::executor::{ApplyKind, ApplyRet, Executor};
+use fvm_ipld_encoding::tuple::*;
+use fvm_ipld_encoding::{strict_bytes, BytesSer, RawBytes};
+use fvm_shared::address::Address;
+use fvm_shared::message::Message;
+use fvm_shared::{ActorID, METHOD_CONSTRUCTOR};
+
+use crate::{Account, BasicTester};
+
+pub fn create_contract(tester: &mut BasicTester, owner: &mut Account, contract: &[u8]) -> ApplyRet {
```
aakoshh, line 13: Should there be some optional constructor arguments passed here?

aakoshh: Never mind, I see the constructor only expects `initcode`. I thought maybe you can pass parameters to it.

tools/fvm-bench/src/main.rs
```diff
@@ -0,0 +1,88 @@
...
+    /// Emit user generated logs
+    #[arg(short, long, default_value = "false")]
+    events: bool,
+
+    /// Builtin actors bundle to use.
+    #[arg(short, long)]
+    bundle: String,
+
+    /// Contract file.
+    contract: String,
```
aakoshh, line 35: Okay so what if the contract has a constructor with non-zero arguments? I just read that they should be appended to the `initcode`. If you expect developers to append those to the actual file, it would be worth mentioning it here. Or should the tool take further parameters?

tools/testkit/src/fevm.rs
vyzo: done

Stebalien (approved):
The integration framework still needs work, but this is a step in the right direction.

tools/fvm-bench/src/main.rs
aakoshh: We can't just say constructor args are not supported whatsoever, when a Solidity contract can have constructor arguments.

vyzo: really?
We had this at the beginning and got removed aling the way.
Now it is just iniycode you pass.

cc @stebalien.

Stebalien: I'm confused. What are you referring to here?

wadeAlexC: Solidity constructors can have parameters, but parameters just get tacked onto the end of the initcode, so there probably doesn't need to be any additional code to support them.

PR #1495, opened 2023-01-17
add feature to support hyperspace version lineages.

raulk:
We expect to upgrade Hyperspace in ways that do not require changes to ref-fvm. Instead of modelling every network version independently, we absorb future network versions into a general enum value "NetworkVersion::VNNOnwards".

Future upgrades involving ref-fvm changes will motivate creating new physical network versions.


shared/src/sector/registered_proof.rs
```diff
@@ -33,23 +33,13 @@ pub enum RegisteredSealProof {
...
     /// Returns registered seal proof for given sector size
-    pub fn from_sector_size(size: SectorSize, network_version: NetworkVersion) -> Self {
-        if network_version < NetworkVersion::V7 {
-            match size {
-                SectorSize::_2KiB => Self::StackedDRG2KiBV1,
-                SectorSize::_8MiB => Self::StackedDRG8MiBV1,
-                SectorSize::_512MiB => Self::StackedDRG512MiBV1,
-                SectorSize::_32GiB => Self::StackedDRG32GiBV1,
-                SectorSize::_64GiB => Self::StackedDRG64GiBV1,
-            }
```
raulk, line 44: ref-fvm cannot service nv7, and nv7 actors were Go specs-actors, so this can be removed.

shared/src/version/mod.rs
```diff
@@ -85,4 +96,13 @@ impl TryFrom<u32> for NetworkVersion {
             _ => Err(value),
         }
     }
+
+    #[cfg(feature = "hyperspace")]
+    fn try_from(value: u32) -> Result<Self, Self::Error> {
+        use NetworkVersion::*;
+        match value {
+            18..=u32::MAX => Ok(V18Onwards),
```
raulk, line 104: Every future network version is mapped to v18Onwards.

raulk:
There's also a case here for retyping `NetworkVersion` out of an enum, similar to what we did with `ExitCode` a while ago. Everywhere we use this type is to refer to a number, so having constants instead would work just as well.

Stebalien:
https://github.com/filecoin-project/ref-fvm/pull/1496

raulk:
Rejiggered/simplified on top of #1496.

Stebalien (approved):

PR #1506, opened 2023-01-18
FVM: Machine: Put the Empty Array CID on creation, if missing

### LINKED ISSUE #1359: Write the "empty array" block when constructing a machine
We assume it exists, so we should probably write it. Technically, it's guaranteed to exist (for now) as the state-tree root links to an empty CBOR "metadata" object, but it costs us (almost) nothing to just write it once whenever we create a new machine. We can even pre-hash it, so we're just copying some bytes around.

arajasek:
Fixes #1359 

I wanted to put this in a separate method (something like `put_empties_in_blockstore()`, because we'll probably want to expand this to include things like empty maps, but wasn't sure how to specify the type for the blockstore param (and didn't want to fight Rust). Happy to make that change if we think it's better (and @Stebalien shows me how to do it) :P

### DIFF
--- a/fvm/src/machine/default.rs
+++ b/fvm/src/machine/default.rs
@@ -5,8 +5,8 @@ use std::ops::RangeInclusive;
 use anyhow::{anyhow, Context as _};
 use cid::Cid;
 use fvm_ipld_amt::Amt;
-use fvm_ipld_blockstore::{Blockstore, Buffered};
-use fvm_ipld_encoding::CborStore;
+use fvm_ipld_blockstore::{Block, Blockstore, Buffered};
+use fvm_ipld_encoding::{to_vec, CborStore, DAG_CBOR};
 use fvm_shared::address::Address;
 use fvm_shared::econ::TokenAmount;
 use fvm_shared::error::ErrorNumber;
@@ -14,6 +14,7 @@ use fvm_shared::event::StampedEvent;
 use fvm_shared::version::NetworkVersion;
 use fvm_shared::ActorID;
 use log::debug;
+use multihash::Code::Blake2b256;
 
 use super::{Machine, MachineContext};
 use crate::blockstore::BufferedBlockstore;
@@ -24,10 +25,17 @@ use crate::kernel::{ClassifyResult, Result};
 use crate::machine::limiter::DefaultMemoryLimiter;
 use crate::machine::Manifest;
 use crate::state_tree::{ActorState, StateTree};
-use crate::syscall_error;
 use crate::system_actor::State as SystemActorState;
+use crate::{syscall_error, EMPTY_ARR_CID};
 
 pub const EVENTS_AMT_BITWIDTH: u32 = 5;
+
+lazy_static::lazy_static! {
+    /// Pre-serialized block containing the empty array
+    pub static ref EMPTY_ARRAY_BLOCK: Block<Vec<u8>> = {
+        Block::new(DAG_CBOR, to_vec::<[(); 0]>(&[]).unwrap())
+    };
+}
 
 pub struct DefaultMachine<B, E> {
     /// The initial execution context for this epoch.
@@ -91,6 +99,8 @@ where
                 &context.initial_state_root
             ));
         }
+
+        put_empty_blocks(&blockstore)?;
 
         // Create a new state tree from the supplied root.
         let state_tree = {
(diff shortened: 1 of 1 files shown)


fvm/src/machine/default.rs
```diff
@@ -92,6 +93,20 @@ where
...
+        {
+            let empty_cid = blockstore.put_cbor::<[(); 0]>(&[], Blake2b256)?;
+            if empty_cid != *EMPTY_ARR_CID {
+                return Err(anyhow!(
+                    "empty CID sanity check failed, {} != {}",
+                    empty_cid,
+                    *EMPTY_ARR_CID,
+                ));
+            }
+        }
```
Stebalien, line 108: nit: I'd just put it without checking. It'll likely be _faster_.

Stebalien:
> I wanted to put this in a separate method (something like put_empties_in_blockstore(), because we'll probably want to expand this to include things like empty maps, but wasn't sure how to specify the type for the blockstore param (and didn't want to fight Rust)

I'd pre-serialize, honestly. I.e., build up a (lazy static) list of pre-serialized blocks, then just put each one.

fvm/src/machine/default.rs
Stebalien: It'll be faster because:

1. Hashing is pretty fast.
2. Reading from the blockstore is flow.
3. A couple of extra puts likely won't cause any issues (and, to be honest, this block will likely be written by several actors as well.

Stebalien: It'll also be more correct. With the way the code is currently written, the splitstore could come in and delete the block.

fvm/src/machine/default.rs
```diff
@@ -24,11 +25,18 @@ use crate::kernel::{ClassifyResult, Result};
...
 use crate::state_tree::{ActorState, StateTree};
-use crate::syscall_error;
 use crate::system_actor::State as SystemActorState;
+use crate::{syscall_error, EMPTY_ARR_CID};
 
 pub const EVENTS_AMT_BITWIDTH: u32 = 5;
 
+lazy_static::lazy_static! {
+    /// The CBOR-serialized bytes of the empty array
+    pub static ref EMPTY_ARRAY_BLOCK_DATA: Vec<u8> = {
```
arajasek, line 35: I wanted to have this be a `Block`, but that requires specifying the trait type `D: AsRef<[u8]> + ?Sized`, which led to Rust being Rust.

fvm/src/machine/default.rs
```diff
@@ -92,6 +100,17 @@ where
             ));
         }
 
+        let empty_arr_cid =
+            blockstore.put(Blake2b256, &Block::new(DAG_CBOR, &*EMPTY_ARRAY_BLOCK_DATA))?;
+
+        if empty_arr_cid != *EMPTY_ARR_CID {
```
arajasek, line 106: Worth doing, or just drop?

arajasek (commented):
Two bikeshed questions, but we can land this whenever we want to do more important work :P

fvm/src/machine/default.rs
arajasek: Ack, thanks!

fvm/src/machine/default.rs
Stebalien: You should be able to specify `D` as `Vec<u8>`?

fvm/src/machine/default.rs
Stebalien: `debug_assert`.

fvm/src/machine/default.rs
arajasek: Done!

fvm/src/machine/default.rs
```diff
@@ -92,6 +100,13 @@ where
             ));
         }
 
+        let empty_arr_cid = blockstore.put(Blake2b256, &EMPTY_ARRAY_BLOCK)?;
+
+        debug_assert!(
+            empty_arr_cid == *EMPTY_ARR_CID,
+            "empty CID sanity check failed",
+        );
```
Stebalien, line 108: Helper "put empty blocks" function? With a comment explaining why we do this because this is really non-obvious?

(but, like, lowest priority ever)

arajasek: done

Stebalien (approved):

PR #1511, opened 2023-01-19
fix: fix build when testing isn't enabled

Stebalien:
This appears to be a casualty of rust's feature unification. We should have caught this in CI, but feature unification ensured that this feature got _enabled_, so we didn't.

### DIFF
--- a/fvm/src/kernel/default.rs
+++ b/fvm/src/kernel/default.rs
@@ -823,9 +823,11 @@ where
         actor_id: ActorID,
         delegated_address: Option<Address>,
     ) -> Result<()> {
-        let is_allowed_to_create_actor = self.actor_id == INIT_ACTOR_ID
-            || (cfg!(feature = "testing")
-                && self.actor_id == TEST_ACTOR_ALLOWED_TO_CALL_CREATE_ACTOR);
+        let is_allowed_to_create_actor = self.actor_id == INIT_ACTOR_ID;
+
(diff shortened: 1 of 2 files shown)


fvm/src/kernel/default.rs
```diff
@@ -823,9 +823,11 @@ where
...
         delegated_address: Option<Address>,
     ) -> Result<()> {
-        let is_allowed_to_create_actor = self.actor_id == INIT_ACTOR_ID
-            || (cfg!(feature = "testing")
-                && self.actor_id == TEST_ACTOR_ALLOWED_TO_CALL_CREATE_ACTOR);
+        let is_allowed_to_create_actor = self.actor_id == INIT_ACTOR_ID;
+
+        #[cfg(feature = "testing")]
+        let is_allowed_to_create_actor =
+            is_allowed_to_create_actor || self.actor_id == TEST_ACTOR_ALLOWED_TO_CALL_CREATE_ACTOR;
```
Stebalien, line 830: This is an idiom I use when using feature flags. Re-using the variable is painful because:

1. When the feature is enabled, the binding needs to be mutable.
2. When disabled, the binding needs to be immutable (you'll get a warning).

Simply re-binding the variable like this avoids that situation.

maciejwitowski: Good to know, thanks!

PR #1512, opened 2023-01-19
fix: expose the effective gas premium

### LINKED ISSUE #1471: The `GASPRICE` opcode does not return the effective gas price
The [EIP-1559 states](https://github.com/ethereum/EIPs/blob/master/EIPS/eip-1559.md#specification) that the `GASPRICE` opcode **must** return the `effective_gas_price` as defined in the specification. The effective gas price is the sum of the base fee per gas and the priority fee per gas (miner tip), capped at a maximum of `max_fee_per_gas`.

However, the FEVM implementation in `builtin-actors/actors/evm/src/interpreter/instructions/context.rs#L84` for the `GASPRICE` opcode does not return the capped value. Instead, it returns the sum of the base fee and the priority fee per gas.

##### Recommendation

We recommend returning the capped value of `effective_price`.

Stebalien:
We previously exposed the specified gas premium, but that's not all that useful. The user pays the _effective_ gas premium, which is capped at `feecap - basefee`. I.e., the user never pays more than the fee cap.

Unfortunately, this needs to be behind a feature flag on hyperspace until the c.3 upgrade.

fixes #1471

### DIFF
--- a/fvm/src/executor/default.rs
+++ b/fvm/src/executor/default.rs
@@ -80,6 +80,27 @@ where
             events: Vec<StampedEvent>, // TODO consider removing if nothing in the client ends up using it.
         }
 
+        // Filecoin caps the premium plus the base-fee at the fee-cap.
+        // We expose the _effective_ premium to the user.
+        let effective_premium = msg
+            .gas_premium
+            .clone()
+            .min(&msg.gas_fee_cap - &self.context().base_fee)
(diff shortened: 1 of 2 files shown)


maciejwitowski:
Makes sense. Added issue to clean the flag https://github.com/filecoin-project/ref-fvm/issues/1515

PR #1516, opened 2023-01-19
Release ipld_encoding 0.3.3, FVM 3.0.0-alpha.21

arajasek:
As in the title.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1521,7 +1521,7 @@ dependencies = [
  "frc42_dispatch",
  "fvm_actor_utils",
  "fvm_ipld_blockstore 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
- "fvm_ipld_encoding 0.3.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "fvm_ipld_encoding 0.3.2",
  "fvm_shared 3.0.0-alpha.17 (registry+https://github.com/rust-lang/crates.io-index)",
  "num-derive",
  "num-traits",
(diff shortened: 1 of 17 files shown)


Cargo.lock
```diff
@@ -1541,7 +1541,7 @@ dependencies = [
  "futures",
  "fvm_ipld_blockstore 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "fvm_ipld_car 0.6.0 (registry+https://github.com/rust-lang/crates.io-index)",
- "fvm_ipld_encoding 0.3.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "fvm_ipld_encoding 0.3.2",
```
jennijuju, line 1544: what does (registry+https://github.com/rust-lang/crates.io-index)", mean and do

Cargo.lock
```diff
@@ -1521,7 +1521,7 @@ dependencies = [
  "frc42_dispatch",
  "fvm_actor_utils",
  "fvm_ipld_blockstore 0.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
- "fvm_ipld_encoding 0.3.2 (registry+https://github.com/rust-lang/crates.io-index)",
+ "fvm_ipld_encoding 0.3.2",
```
jennijuju, line 1524: is this suppose to be 0.3.3?

arajasek: Yeah, that one makes sense because the version of builtin-actors we're using hasn't updated to 0.3.3 yet.

Cargo.lock
arajasek: I have no idea, I don't question the lock file.

PR #1524, opened 2023-01-20
feat: `CarReader::read_into()`

### LINKED ISSUE #1522: fvm_ipld_car: load_car with existing CarReader failing
Due to the `CarReader` taking the header bytes on creation; when passing a car reader to `load_car`, the new reader created in load car fails from the header bytes missing.

It would be useful in certain cases to be able to perform checks and or manipulate a car file before loading it into the internal store.

For example, requesting a car file from an arbitrary gateway, you'd want to make sure that the car file you get is for the data you intended for, otherwise you could potentially load a bunch of arbitrary content into the store, only know after the fact, and now need to garbage collect the store

ozwaldorf:
# Why

resolves #1522 

# What

- Move private fn `load_car_inner` to `CarReader::read_into()`
- `load_car` and `load_car_unchecked` construct their own car readers and call `read_into`
- Add file test using CarReader directly 

# Summary

Provide a way for users to perform actions/checks on car data before loading it into the store

### DIFF
--- a/ipld/car/src/lib.rs
+++ b/ipld/car/src/lib.rs
@@ -126,6 +126,24 @@ where
             Ok(None)
         }
     }
+
+    /// Loads the CAR file into the given blockstore
+    pub async fn read_into<B: Blockstore>(mut self, s: &B) -> Result<Vec<Cid>, Error> {
+        // Batch write key value pairs from car file
+        // TODO: Stream the data once some of the stream APIs stabilize.
+        let mut buf = Vec::with_capacity(100);
+        while let Some(block) = self.next_block().await? {
+            buf.push((block.cid, block.data));
+            if buf.len() > 1000 {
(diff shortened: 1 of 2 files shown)


Stebalien:
I see. Yeah, that makes sense.

But how about making this a method on `CarReader`? I.e., `read_into`?

That way, the user can either:

1. Load the car in one pass with `load_car`.
2. Open a car, do some checking, then call `reader.read_into(&bs)`.

ozwaldorf:
Interesting, that actually could be a nice change. Would load_car_inner even be needed at that point? load_car/unchecked can just create the reader and immediately call read_into and could completely remove load_car_inner

ozwaldorf:
Updated the description according to `read_into` approach

Stebalien (approved):
Thanks!

PR #1561, opened 2023-01-25
HAMT iteration

alexytsu:
A more general approach for iteration

Spiked in the HAMT library, showing skipping of nodes that are definitely "smaller" than a particular starting node

This technique should also work for the AMT. 

This API specifies a `max` number of values traversed, rather than nodes visited/expanded. Nodes visited/expanded can probably be passed back through the closure to the caller in order for the predicate to terminate iteration at a certain point if needed.

### DIFF
--- a/ipld/hamt/src/node.rs
+++ b/ipld/hamt/src/node.rs
@@ -534,6 +534,36 @@ where
             let mut new_path = Vec::from(curr_path);
             new_path.push(i);
 
+            // if new_path is larger at any point, don't skip
+            let mut skip = None;
+            for (new, gate) in new_path.iter().zip(start_at.iter()) {
+                match new.cmp(gate) {
+                    std::cmp::Ordering::Less => {
+                        // if new_path is smaller at any point, skip
+                        skip = Some(true);
+                        break;
+                    }
+                    std::cmp::Ordering::Equal => {}
+                    std::cmp::Ordering::Greater => {
+                        // if new_path is greater at any point, definitely explore
+                        skip = Some(false);
+                        break;
+                    }
+                }
+            }
+
+            match skip {
+                // told to skip
+                Some(true) => {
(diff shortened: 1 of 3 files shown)


ipld/hamt/src/node.rs
```diff
@@ -507,4 +507,126 @@ where
...
+                // told to skip
+                Some(true) => {
+                    println!("Skipping {:?}", new_path);
+                    continue;
+                }
+                // told to not skip
+                Some(false) => {}
+                // matches so far, we are along the parent path so keep exploring
+                None => {}
+            }
```
alexytsu, line 565: needs to be refactored/tidied up and tested more thoroughly

ipld/hamt/src/node.rs
```diff
@@ -507,4 +507,126 @@ where
...
+    // returns number of values traversed in the subtree
+    pub(crate) fn for_each_while<S, F>(
+        &self,
+        store: &S,
+        start_at: &[u8],
+        curr_path: &[u8],
+        limit: u64,
+        conf: &Config,
+        f: &mut F,
+    ) -> Result<u64, Error>
```
alexytsu, line 520: needs to return the path of the final node traversed

ipld/hamt/src/node.rs
```diff
@@ -507,4 +507,126 @@ where
             Err(err) => Err(err),
         }
     }
+
+    // returns number of values traversed in the subtree
+    pub(crate) fn for_each_while<S, F>(
+        &self,
+        store: &S,
+        start_at: &[u8],
```
alexytsu, line 515: range start needs an opaque type

alexytsu:
While in proof-of-concept, I'll be working here https://github.com/helix-onchain/ref-fvm/pull/3

PR #1629, opened 2023-02-02
EVM: apply read-only early and reject events

### LINKED ISSUE #1624: FVM: Fail early when read-only
Currently, we fail when we actually try to update state. However, now that we're charging for actor reads/writes in various places, it would be better to fail _early_. #1521 exposes a `is_read_only` method on the call manager, and we should use that in:

1. `Kernel::set_root`.
2. `Kernel::emit_event`.
3. `Kernel::create_actor`.
4. `CallManager::send` when the value is greater than 0.
5. `CallManager::send` when auto-creating actors.

### LINKED ISSUE #1628: FIP: Reject events when read-only

Stebalien:
This moves read-only handling from inside the state-tree to the kernel, and rejects events when read-only.

I originally put it inside the state-tree to simplify things, but it meant that that the read-only state only got applied at the end, and somewhat inconsistently:

1. Non-zero value transfers to self technically worked.
2. Gas was charged up to the first time a state-update was attempted.

Now, read-only is applied in the kernel _except_ when automatically creating new actors on send (because we don't know if it exists yet).

fixes #1628
fixes #1624

### DIFF
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -174,6 +174,7 @@ where
         params: Option<Block>,
         value: &TokenAmount,
         gas_limit: Option<Gas>,
+        read_only: bool,
     ) -> Result<InvocationResult>
     where
         K: Kernel<CallManager = Self>,
@@ -196,8 +197,9 @@ where
             self.gas_tracker.push_limit(limit);
         }
 
-        let mut result =
-            self.with_stack_frame(|s| s.send_unchecked::<K>(from, to, method, params, value));
+        let mut result = self.with_stack_frame(|s| {
+            s.send_unchecked::<K>(from, to, method, params, value, read_only)
+        });
 
         // If we pushed a limit, pop it.
         if gas_limit.is_some() {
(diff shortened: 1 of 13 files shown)


fvm/src/call_manager/default.rs
```diff
@@ -617,12 +616,18 @@ where
             Some(addr) => addr,
             None => match to.payload() {
                 Payload::BLS(_) | Payload::Secp256k1(_) => {
+                    if read_only {
```
Stebalien, line 619: Checked after checking the payload type so we return "not found" before "read only".

fvm/src/kernel/default.rs
```diff
@@ -371,11 +388,11 @@ where
         }
 
         // Send.
-        let result = self
-            .call_manager
-            .with_transaction(flags.read_only(), |cm| {
-                cm.send::<Self>(from, *recipient, method, params, value, gas_limit)
-            })?;
+        let result = self.call_manager.with_transaction(|cm| {
+            cm.send::<Self>(
```
Stebalien, line 392: So, technically, instead of passing a "read-only" flag, we could change the kernel's type into a read-only kernel. But it just doesn't seem worth it.

arajasek (approved):

PR #1630, opened 2023-02-02
feat: remove negative gas charges everywhere

Stebalien:
This removes negative gas charges and makes gas a u64. Motivations:

1. We (finally) no longer have negative gas charges.
2. This makes it easier to reason about gas and ensure that gas never goes "backwards".

If the gas limit starts above i64::MAX _milligas_, we saturate it to i64::MAX.

### DIFF
--- a/fvm/src/gas/mod.rs
+++ b/fvm/src/gas/mod.rs
@@ -20,7 +20,7 @@ mod outputs;
 mod price_list;
 mod timer;
 
-pub const MILLIGAS_PRECISION: i64 = 1000;
+pub const MILLIGAS_PRECISION: u64 = 1000;
 
 /// A typesafe representation of gas (internally stored as milligas).
 ///
@@ -29,7 +29,7 @@ pub const MILLIGAS_PRECISION: i64 = 1000;
 ///   by gas).
 /// - Makes it harder to confuse gas and milligas.
 #[derive(Hash, Eq, PartialEq, Ord, PartialOrd, Copy, Clone, Default)]
-pub struct Gas(i64 /* milligas */);
+pub struct Gas(u64 /* milligas */);
 
 impl Debug for Gas {
     fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
@@ -60,37 +60,32 @@ impl Display for Gas {
 impl Gas {
     /// Construct a `Gas` from milligas.
     #[inline]
-    pub const fn from_milligas(milligas: i64) -> Gas {
+    pub const fn from_milligas(milligas: u64) -> Gas {
         Gas(milligas)
     }
 
-    /// Construct a `Gas` from gas, scaling up. If this exceeds the width of an i64, it saturates at
-    /// `i64::MAX` milligas.
+    /// Construct a `Gas` from gas, scaling up. If this exceeds the width of a u64, it saturates at
+    /// `u64::MAX` milligas.
     #[inline]
-    pub const fn new(gas: i64) -> Gas {
+    pub const fn new(gas: u64) -> Gas {
         Gas(gas.saturating_mul(MILLIGAS_PRECISION))
-    }
-
-    #[inline]
-    pub const fn is_saturated(&self) -> bool {
-        self.0 == i64::MAX
     }
 
     /// Returns the gas value as an integer, rounding the fractional part up.
     #[inline]
-    pub const fn round_up(&self) -> i64 {
+    pub const fn round_up(&self) -> u64 {
         milligas_to_gas(self.0, true)
     }
 
     /// Returns the gas value as an integer, truncating the fractional part.
     #[inline]
-    pub const fn round_down(&self) -> i64 {
+    pub const fn round_down(&self) -> u64 {
         milligas_to_gas(self.0, false)
     }
 
     /// Returns the gas value as milligas, without loss of precision.
     #[inline]
-    pub const fn as_milligas(&self) -> i64 {
+    pub const fn as_milligas(&self) -> u64 {
         self.0
     }
 }
@@ -137,30 +132,12 @@ impl Sub for Gas {
     }
 }
 
-impl Mul<i64> for Gas {
-    type Output = Gas;
-
-    #[inline]
-    fn mul(self, rhs: i64) -> Self::Output {
-        Self(self.0.saturating_mul(rhs))
-    }
-}
-
-impl Mul<i32> for Gas {
-    type Output = Gas;
-
-    #[inline]
-    fn mul(self, rhs: i32) -> Self::Output {
-        Self(self.0.saturating_mul(rhs.into()))
-    }
-}
-
 impl Mul<u64> for Gas {
     type Output = Gas;
 
     #[inline]
     fn mul(self, rhs: u64) -> Self::Output {
-        Self(self.0.saturating_mul(rhs.try_into().unwrap_or(i64::MAX)))
+        Self(self.0.saturating_mul(rhs))
     }
 }
 
@@ -178,7 +155,7 @@ impl Mul<usize> for Gas {
 
     #[inline]
     fn mul(self, rhs: usize) -> Self::Output {
-        Self(self.0.saturating_mul(rhs.try_into().unwrap_or(i64::MAX)))
+        Self(self.0.saturating_mul(rhs.try_into().unwrap_or(u64::MAX)))
     }
 }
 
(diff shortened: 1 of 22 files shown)


fvm/src/gas/mod.rs
```diff
@@ -60,37 +60,32 @@ impl Display for Gas {
...
+    /// Construct a `Gas` from gas, scaling up. If this exceeds the width of a u64, it saturates at
+    /// `u64::MAX` milligas.
     #[inline]
-    pub const fn new(gas: i64) -> Gas {
+    pub const fn new(gas: u64) -> Gas {
         Gas(gas.saturating_mul(MILLIGAS_PRECISION))
     }
 
-    #[inline]
-    pub const fn is_saturated(&self) -> bool {
```
Stebalien, line 75: This wasn't particularly reliable.

fvm/src/gas/outputs.rs
```diff
@@ -61,27 +59,33 @@ impl GasOutputs {
     }
 }
 
-fn compute_gas_overestimation_burn(gas_used: i64, gas_limit: i64) -> (i64, i64) {
-    const GAS_OVERUSE_NUM: i64 = 11;
-    const GAS_OVERUSE_DENOM: i64 = 10;
+fn compute_gas_overestimation_burn(gas_used: u64, gas_limit: u64) -> (u64, u64) {
```
Stebalien, line 62: TODO: write a test

fvm/src/gas/price_list.rs
```diff
@@ -323,24 +323,18 @@ pub(crate) struct StepCost(Vec<Step>);
...
-                break;
-            }
-            i += 1;
-        }
-        i -= 1;
-        if i < 0 {
-            return Gas::zero();
-        }
-        self.0[i as usize].cost
+    pub(crate) fn lookup(&self, x: u64) -> Gas {
```
Stebalien, line 331: TODO: also write a test. I changed this because it's easier to prove that the new code never panics or anything.

fvm/src/gas/outputs.rs
Stebalien: I changed it because I wasn't sure about wrap-around.

fvm/src/syscalls/mod.rs
```diff
@@ -47,7 +47,7 @@ pub struct InvocationData<K> {
     /// The last-set milligas limit. When `charge_for_exec` is called, we charge for the
     /// _difference_ between the current gas available (the wasm global) and the
     /// `last_milligas_available`.
-    pub last_milligas_available: i64,
+    pub last_gas_available: Gas,
```
Stebalien, line 50: Might as well use our gas type.

fvm/src/syscalls/mod.rs
```diff
@@ -87,32 +94,47 @@ pub fn charge_for_exec<K: Kernel>(
...
+    // `exec_gas_charge` is the number we want to charge. If, for some reason, `memory_gas_charge`
+    // exceeds `exec_gas_charge`, we just set `memory_gas_charge` to `exec_gas_charge`, and set
+    // `exec_gas_charge` to zero.
+    let mut memory_gas_charge = data.kernel.price_list().grow_memory_gas(memory_delta_bytes);
+    if memory_gas_charge <= exec_gas_charge {
+        exec_gas_charge -= memory_gas_charge;
+    } else {
+        memory_gas_charge = exec_gas_charge;
+        exec_gas_charge = Gas::zero();
+    }
```
Stebalien, line 130: I don't think this could have ever been an issue, but... we might as well be careful.

fvm/src/gas/mod.rs
```diff
@@ -194,10 +171,18 @@ pub struct GasTracker {
...
     /// Gas limit and gas used are provided in protocol units (i.e. full units).
     /// They are converted to milligas for internal canonical accounting.
-    pub fn new(gas_limit: Gas, gas_used: Gas, enable_tracing: bool) -> Self {
+    ///
+    /// - If the gas limit exceeds `i64::MAX` milligas, it's rounded down to `i64::MAX` milligas and
+    ///   treated as "unlimited".
+    /// - If the gas used exceeds the gas limit, it's capped at the gas limit.
+    pub fn new(mut gas_limit: Gas, mut gas_used: Gas, enable_tracing: bool) -> Self {
+        gas_limit = gas_limit.min(UNLIMITED_GAS);
+        gas_used = gas_used.min(gas_limit);
```
Stebalien, line 185: I changed this to be consistent. That way, if we have "too much" gas, we just don't burn gas. Otherwise, there were a lot of edge-cases around the maximums that were hard to reason about.

Stebalien:
This is ready for a preliminary review, but I need to track down an issue with the events test. I.e., there's probably a bug.

fvm/src/gas/outputs.rs
```diff
@@ -61,27 +59,106 @@ impl GasOutputs {
...
+    // This computes the fraction of the "remaining" gas to burn and will never be greater than 100%
+    // of the remaining gas.
+    let gas_to_burn = (gas_remaining * over) / gas_used;
+
+    // But... we use saturating sub, just in case.
+    let refund = gas_remaining.saturating_sub(gas_to_burn);
 
-    if over > gas_used {
-        over = gas_used;
+    (refund as u64, gas_to_burn as u64)
```
raulk, line 90: This is a ton clearer, thanks.

fvm/src/gas/outputs.rs
```diff
@@ -61,27 +59,106 @@ impl GasOutputs {
...
+}
+
+// Adapted from lotus.
+#[test]
+fn overestimation_burn_test() {
```
raulk, line 95: Out of curiosity -- would it be straightforward to do a one-off run of these tests against the old logic just to make sure we haven't inadvertently introduced a regression?

raulk (approved):

fvm/src/gas/outputs.rs
Stebalien: I just stole these tests from lotus, which have been running against the lotus implementation of this function.

fvm/src/syscalls/mod.rs
```diff
@@ -65,17 +63,27 @@ pub fn update_gas_available(
...
         .map_err(|e| Abort::Fatal(anyhow!("failed to set available gas global: {}", e)))?;
 
-    ctx.data_mut().last_milligas_available = avail_milligas;
-
-    // Also adjust the instant we use in `charge_for_exec` to measure wasm execution time.
-    ctx.data_mut().last_charge_time = GasTimer::start();
+    // Finally, update the last-seen values. We'll use these values in `charge_for_exec` below.
+    let data = ctx.data_mut();
+    data.last_gas_available = avail_gas;
+    data.last_memory_bytes = data.kernel.limiter_mut().memory_used();
```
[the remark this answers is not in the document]
Stebalien: Not explicitly, but it shouldn't have been an issue (we were doing this when we exited wasm). I just don't trust our code.

NOTE: I've also ensured that `last_memory_bytes` has no impact on actual gas usage (it's used for tracking memory allocation costs.

fvm/src/syscalls/mod.rs
```diff
@@ -87,49 +95,65 @@ pub fn charge_for_exec<K: Kernel>(
...
     let data = ctx.data_mut();
 
+    // abs_diff(0) is the simplest way to get the absolute value of an i64 as a u64 without
+    // overflows.
+    let milligas_available_wasm_abs = milligas_available_wasm.abs_diff(0);
+
+    // Get the exec gas to charge, taking negatives into account.
+    let mut exec_gas_charge = if milligas_available_wasm < 0 {
+        // If the gas remaining is negative, we charge for all remaining gas, plus `-remaining_gas`.
+        // That way we actually run out.
```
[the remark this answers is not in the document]
Stebalien: Yes. That's how things currently work.

(which I got wrong the first 3 times I tried this...)

fvm/src/syscalls/mod.rs
```diff
@@ -87,49 +95,65 @@ pub fn charge_for_exec<K: Kernel>(
...
+    // `exec_gas_charge` is the number we want to charge. If, for some reason, `memory_gas_charge`
+    // exceeds `exec_gas_charge`, we just set `memory_gas_charge` to `exec_gas_charge`, and set
+    // `exec_gas_charge` to zero.
+    let mut memory_gas_charge = data.kernel.price_list().grow_memory_gas(memory_delta_bytes);
+    if memory_gas_charge <= exec_gas_charge {
+        exec_gas_charge -= memory_gas_charge;
+    } else {
+        memory_gas_charge = exec_gas_charge;
+        exec_gas_charge = Gas::zero();
+    }
```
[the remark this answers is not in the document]
Stebalien: I'll update the comments. We've already charged (as part of the compute charge). We split it out for tracing/benchmarking purposes.

(really, I'd like to disable all that stuff when tracing is disabled, but I'm leaving that for a future patch)

fvm/src/syscalls/send.rs
```diff
@@ -28,7 +28,9 @@ pub fn send(
     let value = TokenAmount::from_atto((value_hi as u128) << 64 | value_lo as u128);
 
     // If that gas limit exceeds i64, treat it as infinity. u64::MAX is used to indicate "all gas".
-    let gas_limit = gas_limit.try_into().ok().map(Gas::new);
+    // Although really, this doesn't matter. Any gas exceeding the current gas available is treated
+    // as "all remaining gas".
+    let gas_limit = (gas_limit <= (i64::MAX as u64)).then(|| Gas::new(gas_limit));
```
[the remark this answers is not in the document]
Stebalien: Yes. I wrote it this way to be consistent with what we were already doing but... actually, I think I'll just change it to be `u64::MAX`.

(But as the comment says, it doesn't matter for consensus.)

PR #1635, opened 2023-02-03
Add gas parameters for event validation + calibration harnesses.

### LINKED ISSUE #1109: Events: parameterize gas for events
The events implementation has introduced the following gas formula:

```rs
#[inline]
pub fn on_actor_event(&self, evt: &ActorEvent) -> GasCharge {
    let (mut indexed_entries, mut total_bytes) = (0, 0);
    for evt in evt.entries.iter() {
        indexed_entries += evt
            .flags
            .intersection(Flags::FLAG_INDEXED_KEY | Flags::FLAG_INDEXED_VALUE)
            .bits()
            .count_ones();
        total_bytes += evt.key.len() + evt.value.bytes().len();
    }

    GasCharge::new(
        "OnActorEvent",
        self.event_emit_base_cost + (self.event_per_entry_cost * evt.entries.len() as i64),
        (self.event_entry_index_cost * indexed_entries as i64)
            + (self.event_per_byte_cost * total_bytes as i64),
    )
}
```

**In short:**

- event_emit_base_cost: base emission cost for an event.
- event_per_byte_cost: variable cost per byte.
- event_per_entry_cost: cost per event entry.
- event_entry_index_cost: variable cost per indexed entry (should probably account for number of bytes indexed).

The goal of this issue is to find the right pricing. And, of course, to determine if we want to enhance or refine this model in any manner.

raulk:
Closes #1109.

---

## Changes

- Sets the gas parameters for event validation starting from nv18. See below for more details on the choice of gas parameters.
- Adds a harness for event gas calibration under `testing/calibration`.
	- It offers two modes: (a) target a specific encoded event size, (b) generate specific event shapes determined by the number of entries, key length, value length.
	- We currently use (b) to calibrate gas parameters by generating events in the same shape as the EVM runtime actor does for LOG{0..4} opcodes.
		- That is, events with 1-5 entries, with 32 byte values, and with 2-char keys (although the `data` entry carries a 1-char key: `d`).
		- Note that the EVM runtime drops leading zeros in values, so our calibration targets the worst-case scenario where all topics and the payload have no leading zeros.
	- There's a case for simply dropping (a), but I've kept it because it may come in handy later stages of the FVM roadmap. It compiles but it never runs.
- Caps the maximum event length to 1024 bytes. This limit can be disarmed by activating the `gas-calibration` feature.

## Choice of gas parameters

These parameters hit the 10gas/ns baseline consistently, with a slight overestimation to cover for the worst case scenario reasonably well (just under 10gas/ns). These parameters do lead to a larger overestimation on the 1-entry case (LOG0 -- pretty rare anyway), but I'm happy living with that.

| <br/>entries | <br/>size | <br/>compute\_gas | count<br/> | mean<br/> | std<br/> | min<br/> | 25%<br/> | 50%<br/> | 75%<br/> | max<br/> |
| :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- | :--- |
| 1entries | 40 | 2846000 | 490.0 | 14.264802 | 2.191890 | 8.131429 | 14.978947 | 14.978947 | 15.811111 | 16.741176 |
| 2entries | 79 | 3914600 | 490.0 | 12.859819 | 1.223028 | 9.103721 | 12.233125 | 13.498621 | 13.498621 | 14.498519 |
| 3entries | 118 | 4983200 | 490.0 | 12.285889 | 1.019456 | 9.384557 | 12.458000 | 12.777436 | 12.777436 | 13.113684 |
| 4entries | 157 | 6051800 | 490.0 | 11.927226 | 0.987889 | 9.019076 | 12.103600 | 12.350612 | 12.607917 | 12.876170 |
| 5entries | 196 | 7120400 | 490.0 | 11.596974 | 0.932503 | 8.466587 | 11.847587 | 12.048054 | 12.048054 | 12.255422 |

Here is a visual depiction of the above as emitted from the calibration tools:

image

Here's a closer look at the gas distribution in a violin plot:

gas_distribution

## Future direction

This exercise has highlighted the fact that CBOR is not a good choice for an exchange format at the ABI level dealing with (future) untrusted inputs. We found that the costs scale with the number of entries (CBOR list items), which we can't learn prior to parsing the event in some way (we could do an O(n) prevalidation though). It's worth revising this choice going forward. Any changes to the ABI will not have externally visible impacts.


fvm/src/kernel/default.rs
```diff
@@ -985,18 +985,20 @@ where
...
+            let res = match panic::catch_unwind(|| {
+                fvm_ipld_encoding::from_slice(raw_evt).or_error(ErrorNumber::IllegalArgument)
+            }) {
+                Ok(v) => v,
+                Err(e) => {
+                    log::error!("panic when decoding event cbor from actor: {:?}", e);
+                    Err(syscall_error!(IllegalArgument; "panic when decoding event cbor from actor").into())
+                }
+            };
+            t.stop();
```
raulk, line 998: Now capturing a reading even on deserialization errors.

testing/calibration/contract/fil-gas-calibration-actor/src/lib.rs
```diff
@@ -244,14 +294,22 @@ fn random_mutations(data: &mut Vec<u8>, seed: u64, n: usize) {
...
 /// https://en.wikipedia.org/wiki/Linear_congruential_generator
-fn lcg64(mut seed: u64) -> impl Iterator<Item = u64> {
+fn lcg64(initial_seed: u64) -> impl Iterator<Item = u64> {
     let a = 6364136223846793005;
     let c = 1442695040888963407;
+    let mut seed = Wrapping(initial_seed);
     std::iter::repeat_with(move || {
-        seed = a * seed + c;
-        seed
+        seed = Wrapping(a) * seed + Wrapping(c);
```
raulk, line 311: I'm not entirely sure why, but the lack of wrapping arithmetics here caused weird side effects during ASCII string generation, which led to weird characters. This is more correct.

testing/calibration/src/lib.rs
```diff
@@ -24,7 +24,7 @@ use num_traits::Zero;
 use serde::Serialize;
 
 pub const WASM_COMPILED_PATH: &str =
-    "../../target/debug/wbuild/fil_gas_calibration_actor/fil_gas_calibration_actor.compact.wasm";
+    "../../target/release/wbuild/fil_gas_calibration_actor/fil_gas_calibration_actor.compact.wasm";
```
raulk, line 27: This was the culprit for some major headaches. We run the harness in `--release` profile, which means that we recompile contract changes under the `release` profile. Unfortunately the calibration harness was picking the actor up from the `debug` profile, which usually contains a stale version.

@Stebalien -- this might have or might have not affected your readings.

Stebalien: It shouldn't matter. I was timing FVM operations, not things within the actors.

fvm/src/kernel/default.rs
Stebalien: I'd be careful about that. We usually don't time _failures_ because it can mess up the number's we're looking for (e.g., we can stop half-way through parsing something).

testing/calibration/src/lib.rs
raulk: Correct, it should not affect the readings. However, in my case this ended up picking stale actor code more than once, leading to wrong conclusions. So wanted to give you a heads-up just in case.

testing/calibration/contract/fil-gas-calibration-actor/src/lib.rs
```diff
@@ -35,6 +38,8 @@ pub enum Method {
     OnRecoverSecpPublicKey,
     /// Measure sends
     OnSend,
+    /// Emit events variying the number of entries, and with variable lengths for keys and values.
```
aakoshh, line 41:
```suggestion
    /// Emit events varying the number of entries, and with variable lengths for keys and values.
```

raulk: Thanks!

fvm/src/kernel/default.rs
raulk: I think this makes sense here as long as the test harness is controlling for when it's sending malformed input and when it isn't. This allows us to calibrate the costs for invalid CBOR input, although that will be less relevant if/when we move away from the CBOR-based ABI.

testing/calibration/src/lib.rs
Stebalien: Oh.... Damn, I see. I think rust-analyzer's background "let me recompile everything for you and melt a hole in your lap" helped there.

fvm/src/kernel/default.rs
Stebalien: Yeah, fair enough. And the gas charges here are specifically about parsing.

fvm/src/gas/price_list.rs
```diff
@@ -269,13 +269,25 @@ lazy_static! {
...
         },
 
-        // TODO::PARAM
+        // The protocol does not currently mandate indexing work, so these are
+        // left at zero. Once we start populating and committing indexing data
+        // structures, these costs will need to reflect the computation and
+        // storage costs of such actions.
         event_accept_per_index_element: ScalingCost {
             flat: Zero::zero(),
             scale: Zero::zero(),
```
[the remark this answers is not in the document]
Stebalien: Or we can just charge up-front in validation, I guess.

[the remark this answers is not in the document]
Stebalien: But this is important as it factors into the cost to "commit" events.

Stebalien (approved):

fvm/src/gas/price_list.rs
```diff
@@ -890,17 +902,28 @@ impl PriceList {
...
+        // and allocaing into the AMT's buffer on finish.
+        let alloc = self.block_allocate.apply(stamped_event_size) * 2u32;
+
+        // Charge for the hashing on AMT insertion.
+        let hash = self.hashing_cost[&SupportedHashes::Blake2b256].apply(stamped_event_size);
 
         GasCharge::new(
             "OnActorEventAccept",
-            (memcpy * 3u32) + alloc,
+            memcpy + alloc + hash,
```
Stebalien, line 926: To make the benchmarks work out, we should:

1. Put a single memcpy + alloc in the first field.
2. Put the second two copies + alloc + hashing in the second field, as these costs are _deferred_, so the benchmarking won't cover them.

raulk:
I've updated the benchmarks to cover power of two byte array lengths in the range 32..8KiB in the last entry in the event, which in EVM land will correspond to the event data (which is of arbitrary length). The longer that value gets, the higher the overestimation. We will retain these parameters because the award us good security, and are not too oversized for reasonable inputs (and we prefer to encourage reasonable inputs!).

validate

Furthermore, @Stebalien and I revised the theoretical accept costs and made some adjustments (increased to 2 memory allocs, and added cost of hashing). These costs are also somewhat overestimated but award good security (this diagram shows only 1 memcpy and 1 alloc; hiding all deferred costs that happen when committing the events).

accept

fvm/src/gas/price_list.rs
raulk: Done, and updated the benchmarks.

Issue #1514, opened 2023-01-19
Paginated traversal of AMT / HAMT structures


alexytsu:
New methods should be added to the (H)AMT libraries in order to support paginated traversal. The first motivating use case is the `ListTokens` [method](https://github.com/filecoin-project/FIPs/discussions/492#discussioncomment-4150663) and other enumerated methods on [FRC-0053](https://github.com/filecoin-project/FIPs/blob/master/FRCs/frc-0053.md).

Whilst it is possible for the predicate passed into `for_each_while` to terminate traversal, it is not currently possible to then resume traversal at a specific index/range. This makes it infeasible to implement methods such as `ListTokens` due to the gas costs of expanding every node in an AMT.

alexytsu:
A proof-of-concept `for_range_while` method is demonstrated here: which allows traversal to be limited within `[range_start, range_end)` for AMTs.

https://github.com/helix-onchain/ref-fvm/pull/2

alexytsu:
https://github.com/filecoin-project/ref-fvm/pull/1673
https://github.com/filecoin-project/ref-fvm/pull/1665

PR #1636, opened 2023-02-03
Ranged iteration on HAMTs

alexytsu:
With reference to the motivations discussed in https://github.com/filecoin-project/ref-fvm/issues/1514, this is the first part implementing ranged iteration functionality for HAMTs. The same basic concepts here can be applied to AMTs. 

A design tradeoff here is that it is up to the caller of `for_each_ranged` to guarantee that HAMT has not changed since the previous iteration. Actors which expose iteration over HAMTs are recommended to flush HAMTs before traversal and compare the root Cid to ensure contents have not changed.


anorth (approved):
Thank you @alexytsu! This looks good to me, but we should of course wait for some wider review from maintainers of these structures.

Stebalien:
Why not just use the "next key"? This patch adds a new concept of "paths" which exposes internal structure. It should be pretty easy to seek to a specific key and start reading there.

> The same basic concepts here can be applied to AMTs.

AMTs are ordered, so they should _definitely_ just seek to the next index (i.e., don't apply the same concept).

alexytsu:
Thanks for the review and suggestions @Stebalien 

> AMTs are ordered, so they should definitely just seek to the next index (i.e., don't apply the same concept).

Agreed. A method without externally exposed 'paths' was spiked in https://github.com/helix-onchain/ref-fvm/pull/2/files. I think this is more natural where keys are ordered. 

> Why not just use the "next key"? 

I like this idea because it additionally allows immediate feedback that the HAMT is exhausted. In the current algorithm, where the returned cursor points to the last item, it's not possible to tell if there are zero or more items left.

It's probably possible to hide the "path" concept from external users I'll put up a comparative implementation in a separate PR.

alexytsu:
Replaced by https://github.com/filecoin-project/ref-fvm/pull/1665

PR #1658, opened 2023-02-13
Alex Wade's solidity test framework

vyzo:
Transplanted from https://github.com/filecoin-project/fvm-bench/pull/10

### DIFF
--- a/tools/solidity-tests/libraries/FilUtils.sol
+++ b/tools/solidity-tests/libraries/FilUtils.sol
@@ -1,0 +1,182 @@
+// SPDX-License-Identifier: UNLICENSED
+pragma solidity ^0.8.0;
+
+library FilUtils {
+
+    // keccak([])
+    bytes32 constant EVM_EMPTY_CODEHASH = 0xc5d2460186f7233c927e7db2dcc703c0e500b653ca82273b7bfad8045d85a470;
+    // keccak([0xFE])
+    bytes32 constant FIL_NATIVE_CODEHASH = 0xbcc90f2d6dada5b18e155c17a1c0a55920aae94f39857d39d0d8ed07ae8f228b;
+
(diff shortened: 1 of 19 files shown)


Stebalien:
Are these running in CI? If not, can we run them in CI?

vyzo:
Not yet.
Sure, but the difficulty is in having an fvm-bench build and an actors bundle.

Also there is the compile thr contracts each time thing, that i dont quite like.
I think we should compilr them oncr and get done with it, as this does less work and makes things deterministic.

If we resolve these we could perhaps merge CIfying this with the gas stress thingie.

Stebalien (approved):

tools/solidity-tests/libraries/FilUtils.sol
```diff
@@ -0,0 +1,182 @@
...
+    address constant SYSTEM_ACTOR = 0xfF00000000000000000000000000000000000000;
+    address constant INIT_ACTOR = 0xff00000000000000000000000000000000000001;
+    address constant REWARD_ACTOR = 0xff00000000000000000000000000000000000002;
+    address constant CRON_ACTOR = 0xFF00000000000000000000000000000000000003;
+    address constant POWER_ACTOR = 0xFf00000000000000000000000000000000000004;
+    address constant MARKET_ACTOR = 0xff00000000000000000000000000000000000005;
+    address constant VERIFIED_REGISTRY_ACTOR = 0xFF00000000000000000000000000000000000006;
+    address constant DATACAP_TOKEN_ACTOR = 0xfF00000000000000000000000000000000000007;
+    address constant EAM_ACTOR = 0xfF0000000000000000000000000000000000000a;
+    // address constant CHAOS_ACTOR = 0xFF00000000000000000000000000000000000000; // 98
```
wjmelements, line 25: What is the chaos actor?

Issue #1489, opened 2023-01-17
Test: Gas limit fuzzing


Stebalien:
We should "fuzz" our gas limits by:

1. Creating a simple actor that makes several recursive calls, writes some state, etc. Ideally, this actor should never terminate naturally.
2. Executing this actor with different gas limits.
3. Making sure we always run out of gas and never run into any other error.

PR #1659, opened 2023-02-13
Gas Stress/Fuzzing Contracts

vyzo:
See https://github.com/filecoin-project/ref-fvm/issues/1489


Stebalien:
Next step is to use this to try to find bugs. Specifically:

1. Use backtraces and/or gas tracing to try to find the exact amount of gas required to cause a recursive call to fail.
2. Test gas values around this point, ensuring that we always run out of gas.

Stebalien (approved):

Stebalien:
We're going to merge this and continue in a new issue.

PR #1673, opened 2023-02-16
Ranged iteration for AMTs

alexytsu:
Follow up to https://github.com/filecoin-project/ref-fvm/pull/1665/files applying the same principles to AMTs

Since keys on AMTs are definitely clone-able, we don't introduce any new trait bounds with the new method `Node::for_each_while_ranged`. Therefore, we could merge some code paths and have `Amt::for_each_while` call `Node::for_each_while_ranged` passing `None` for start_at and max. This of course incurs a slight performance penalty for the extra bounds checks during complete traversals.

### DIFF
--- a/ipld/amt/src/amt.rs
+++ b/ipld/amt/src/amt.rs
@@ -383,6 +383,51 @@ where
             .map(|_| ())
     }
 
+    pub fn for_each_ranged<F>(
+        &self,
+        start_at: Option<u64>,
+        max: Option<u64>,
+        mut f: F,
+    ) -> Result<(u64, Option<u64>), Error>
+    where
+        F: FnMut(u64, &V) -> anyhow::Result<bool>,
+    {
+        let (_, num_traversed, next_index) = self.root.node.for_each_while_ranged(
+            &self.block_store,
+            start_at,
+            max,
+            self.height(),
+            self.bit_width(),
+            0,
+            &mut |i, v| {
+                f(i, v)?;
+                Ok(true)
+            },
+        )?;
+        Ok((num_traversed, next_index))
+    }
+
+    pub fn for_each_while_ranged<F>(
+        &self,
+        start_at: Option<u64>,
+        max: Option<u64>,
+        mut f: F,
+    ) -> Result<(u64, Option<u64>), Error>
+    where
+        F: FnMut(u64, &V) -> anyhow::Result<bool>,
+    {
+        let (_, num_traversed, next_index) = self.root.node.for_each_while_ranged(
+            &self.block_store,
+            start_at,
+            max,
+            self.height(),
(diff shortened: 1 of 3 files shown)


ipld/amt/src/node.rs
```diff
@@ -553,6 +553,7 @@ where
         Ok((true, did_mutate))
     }
 
+    #[allow(clippy::too_many_arguments)]
```
alexytsu, line 556: if this is ok for an internally used function I'll leave it as is, otherwise could refactor `start_at and limit` into one tuple or compress other parameters

anorth (commented):
This looks pretty good to me, but I would want @Stebalien to review the details too.

We need some tests for an AMT with gaps, including at the start.

ipld/amt/src/amt.rs
```diff
@@ -383,6 +383,90 @@ where
...
+    /// }).unwrap();
+    /// assert_eq!(&values, &[(4, "Four".to_owned()), (5, "Five".to_owned()), (6, "Six".to_owned())]);
+    /// assert_eq!(num_traversed, 3);
+    /// assert_eq!(next_idx, Some(10));
+    /// ```
+    pub fn for_each_ranged<F>(
+        &self,
+        start_at: Option<u64>,
+        max: Option<u64>,
+        mut f: F,
```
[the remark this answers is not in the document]
alexytsu: This follows the pattern of the difference between `for_each` and `for_each_while`. 

The difference is that the predicate here doesn't have a boolean return value and so doesn't terminate iteration until the entire range has been traversed (or an error is returned). 

The boolean value in `for_each_while` gives the caller a chance to terminate iteration early by returning `Ok(false)` from the predicate.

If this code feels too repeated, I'd be ok with removing the `for_each_ranged` entirely.

ipld/amt/src/node.rs
```diff
@@ -552,6 +552,96 @@ where
...
+        S: Blockstore,
+    {
+        let mut traversed_count = 0_u64;
+        match self {
+            Node::Leaf { vals } => {
+                let start_idx = start_at.map_or(0, |s| s.saturating_sub(offset));
+                let mut keep_going = true;
+                for (i, v) in (start_idx..).zip(vals[start_idx as usize..].iter()) {
+                    let idx = offset + i;
+                    if let Some(v) = v {
```
Stebalien, line 582: NOTE: you could write this as:

```suggestion
                for (idx, v) in iter::zip(offset.., vals)
                    .skip(start_idx as usize)
                    // skip empty entries
                    .filter_map(|(i, v)| v.as_ref().map(|v| (i, v)))
                {
```

Stebalien (approved):
LGTM!

Issue #1115, opened 2022-11-17
Test: Check exit codes when we run out of memory


Stebalien:
We need a test-case that covers all possible OOM cases.

1. If we run out of memory while executing (I think we have one?).
2. If we run out of memory because we can't allocate a new wasm instance (we definitely don't have one).

maciejwitowski:
@aakoshh is there any chance you'll be able to help with this?

vyzo:
Not sure how to test for 1 -- what kind of test do you have in mind?

PR #1674, opened 2023-02-17
test actor OOM conditions

vyzo:
See https://github.com/filecoin-project/ref-fvm/issues/1115
This exercises condition 1, whereby an actor can run OOM either by allocating one big chunk or many small ones.


vyzo:
the security audit CI check is spewing nonsense.

Stebalien (approved):
❤️

Issue #1667, opened 2023-02-14
Test: Gas Fuzzing V2


Stebalien:
Now that we have #1489, let's use those contracts to try to find bugs automatically.

Specifically, let's:

1. Run the contract with some reasonable gas limit.
2. Look at the _gas trace_.
3. Use the gas trace to set limits and force the contract to run out of gas just before/after various points in the trace.
4. Ensure that we always fail with "out of gas" and that nothing else happens.

Jacarte:
Well, this is actually what we are doing in https://github.com/filecoin-project/Fuzzing-FVM  right?

Stebalien:
@Jacarte yes, but this one is targeted and deterministic. I.e., regression testing (this one) versus bug finding (general fuzzing).

Jacarte:
Gotcha

PR #1677, opened 2023-02-20
test: gas stress fuzzing

vyzo:
See #1667 

Note: In the very first fuzzpoint, execution fails with `SYS_SENDER_STATE_INVALID` instead of `SYS_OUT_OF_GAS`.
Not sure if this is a bug, ie execution should fail with out of gas or that's correct.

If it _is_ a bug, we should fix it and update the test to not skip the first fuzzpoint.
[EDIT: not a bug in ref-fvm, but rather in the test; see comment below]

Note2: The test exercises 96 fuzzpoints and takes 67.36s on my laptop.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2601,6 +2601,7 @@ dependencies = [
  "fvm_ipld_encoding 0.3.3",
  "fvm_ipld_hamt 0.6.1",
  "fvm_shared 3.0.0-alpha.20",
+ "hex",
  "lazy_static",
  "libsecp256k1",
  "multihash",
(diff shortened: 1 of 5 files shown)


vyzo:
After some debugging, here is what is happening:
- low fuzz fails with OutOfGas as expected
- testkit increments nonce, but tester does not (it shoudnt!, the first chage is message inclusion).
- hi fuzz uses increment nonce and fails with invalid state (nonce mismatch).

So it is not a bug in ref-fvm, but rather a but in the test handing of nonces.

Stebalien (approved):

PR #1686, opened 2023-02-28
ci: github actions maintenance

Stebalien:
- Update actions/checkout
- Switch away from unmaintained actions-rs actions.
- Use our shared sccache action (also used by the builtin actors).


Stebalien:
Blocked on https://github.com/pl-strflt/rust-sccache-action/pull/12.

Issue #1238, opened 2022-12-08
Runtime: use `SupportedHashes` instead of `multihash::Code`


mriise:
We have our own table already defined in the FVM so instead of using multihash's default table (and potentially accidently hashing inside of WASM) we can use something that _cant_ do that

PR #1694, opened 2023-03-03
Refactor the blockstore

Stebalien:
First, remove the multihash "code" from the arguments and use a plain u64. This will let us compile the actors without importing hashing libraries.

fixes https://github.com/filecoin-project/ref-fvm/issues/1238
fixes https://github.com/filecoin-project/ref-fvm/issues/504

Second, replace the concrete `Block` type with a trait. We can then implement this trait for, e.g., `fvm_ipld_encoding::IpldBlock`.

Finally, make the blockstore object safe. This will let us work with the blockstore as `&dyn Blockstore` in most cases.


ipld/blockstore/src/lib.rs
```diff
@@ -34,17 +76,7 @@ pub trait Blockstore {
...
-    fn put<D>(&self, mh_code: multihash::Code, block: &Block<D>) -> Result<Cid>
-    where
-        Self: Sized,
-        D: AsRef<[u8]>,
-    {
-        let k = block.cid(mh_code);
-        self.put_keyed(&k, block.as_ref())?;
-        Ok(k)
-    }
+    fn put(&self, mh_code: u64, block: &dyn Block) -> Result<Cid>;
```
Stebalien, line 79: - I'm using a trait for `Block` to allow abstracting over blocks that own their data, and blocks that borrow their data.
- After implementing this, I think we need to replace `mh_code` with some form of `MultihashType { size: u32, code: u64 }`.

Stebalien:
@anorth can I get feedback on the blockstore trait (and the block trait)? You can ignore everything else.

ipld/blockstore/src/lib.rs
Stebalien: Alternatively, we _could_ simply remove the multihash-code parameter and say that the blockstore gets to decide.

anorth: I don't think removing it is right at this base level. Currying it in to a wrapper store could make sense though.

anorth (commented):
It feels like there's two different interfaces for storing blocks to express here:
- The core one of `(CID, [u8])`
- A convenience method in terms of `(MHCode, Codec, [u8])` that computes the CID

In both cases the codec is assumed to match the data format. In the first one also the hash is assumed to have been computed with the MH code in the CID. The point being that the blockstore isn't checking anything.

Given that, it doesn't make much sense to me why in `put()` the MH code is separate from the `Block`, but the codec isn't. A more convenient method might compute both the serialisation _and_ the CID, from appropriate parameters. But this probably would be less good than the pattern of wrapping, like `CborStore`. 

So forgive me if I've overstepped the scope here, but I don't really get why the core `Blockstore` trait and implementations should do hashing and CID construction. Instead they'd be simpler as a pure `(CID, data)` store abstraction, and then wrappers can curry in either or both the hashing function and serialisation method, no neither need to be passed around. How bad would it be to remove (deprecate) `put`? I think there are few call sites, since most code uses a wrapper like `CborStore`.

[Unfortunately I can see one piece of friction in that most actor code declares to use a `Blockstore` but in fact uses `CborStore` via `impl<T: Blockstore> CborStore for T {}`, so we'd probably need to (a) restrict all those declarations and (b) make CborStore wrap rather than extend BlockStore.]

An alternative intent might be that the blockstore _should_ be responsible for the correctness of hashes corresponding to the data (even if it can't verify encoding), but in that case `put_keyed()` should not exist, and the API look more like the FVM storage syscalls.

In that latter case, or if my earlier thoughts are off the rails, then yes I think these traits look about right.

Stebalien:
> So forgive me if I've overstepped the scope here, but I don't really get why the core Blockstore trait and implementations should do hashing and CID construction. Instead they'd be simpler as a pure (CID, data) store abstraction, and then wrappers can curry in either or both the hashing function and serialisation method, no neither need to be passed around. How bad would it be to remove (deprecate) put? I think there are few call sites, since most code uses a wrapper like CborStore.

The issue is that I'm to support two distinct use-cases:

- FVM -> Client: The FVM will already know the CID so I need a way to "put" with a pre-computed CID (`put_keyed`).
- Builtin actors -> FVM: For security, the FVM needs to compute the CID internally. I.e., it can't trust a CID passed by the user.

I can _probably_ drop the `FVM -> Client` use-case here, honestly. I.e., get rid of `put_keyed`, leaving only `put`.

anorth:
The built-in-actors test VM is another example with the two use cases: untrusted use from within the actors, but trusted use from the surrounding test infrastructure code. However, we can probably support these with two different traits.

rjan90:
Hey! 👋

As part of our cleanup to kick off the year, I'm reviewing all open non-draft pull requests. Could you please do one of the following for your PR?

**1. Close it:** If it's no longer needed.
**2. Mark as Draft:** If it needs more work, and add the next steps.
**3. Ready for Review:** If it's good to go, let me know, and I'll assign a reviewer.

If there's no response in a week, I'll assume it's option 1 and close the PR. If you have any questions, just let me know.

Thanks for your help in keeping things organized, and I appreciate your contributions!

rjan90:
> If there's no response in a week, I'll assume it's option 1 and close the PR.

Closing.

PR #1696, opened 2023-03-07
add standard exit code for non-payable functions

alexytsu:
Adds a standard exit code that allows actors to indicate they are not prepared to receive funds: https://github.com/filecoin-project/builtin-actors/issues/836

Example usage in builtin-actors: https://github.com/filecoin-project/builtin-actors/pull/1241


anorth (commented):
This works for me. Let's wait for @Stebalien as he suggested it.

Stebalien:
It works for me too. My only reservations are:

1. We didn't reserve enough of these...
2. FEVM actors won't use this exit code unless we _liberally_ apply some magic sauce.

But I think we should have this.

Stebalien (approved):

PR #1706, opened 2023-03-10
fix: correctly format syscall error messages

Stebalien:
This makes string interpolation ("message: {some_variable}") work.

Previously, we'd apply string formatting to syscall error messages if and only if we had one or more format arguments. Now, we apply the formatting whenever the message is a string literal.

found by @fridrik01 in https://github.com/filecoin-project/ref-fvm/pull/1698

### DIFF
--- a/fvm/src/kernel/error.rs
+++ b/fvm/src/kernel/error.rs
@@ -11,13 +11,13 @@ pub type Result<T> = std::result::Result<T, ExecutionError>;
 /// Convenience macro for generating Actor Errors
 #[macro_export]
 macro_rules! syscall_error {
-    // Error with only one stringable expression
-    ( $code:ident; $msg:expr ) => { $crate::kernel::SyscallError::new(fvm_shared::error::ErrorNumber::$code, $msg) };
-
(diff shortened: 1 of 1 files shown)


fvm/src/kernel/error.rs
```diff
@@ -11,14 +11,14 @@ pub type Result<T> = std::result::Result<T, ExecutionError>;
 /// Convenience macro for generating Actor Errors
 #[macro_export]
 macro_rules! syscall_error {
-    // Error with only one stringable expression
-    ( $code:ident; $msg:expr ) => { $crate::kernel::SyscallError::new(fvm_shared::error::ErrorNumber::$code, $msg) };
```
fridrik01, line 15: ah so we were always matching this pattern first when passing only a string literal since it taken as expression, that's why you moved it down

fridrik01 (approved):
looks good!

fvm/src/kernel/error.rs
Stebalien: Yep.

Issue #1710, opened 2023-03-13
Fix Github integration failures due to running out of disk space


fridrik01:
I have noticed a failing integration test on my PR (https://github.com/filecoin-project/ref-fvm/pull/1698) that is due running out of disk space on the hosted github CI.

The GH [hardware specs](https://docs.github.com/en/actions/using-github-hosted-runners/about-github-hosted-runners#supported-runners-and-hardware-resources) for these runners list 14GB SSD space - however sometime they _can_ have more disk space than that )

Since our default cargo build and tests generate build artifacts that are awfully close to this limit or even higher then this seems easy to generate flaky tests.

We should update our testing pipeline to minimize binary artifact sizes. For example, there are several huge test binaries  (on_verify_signature, on_event_evm_shapes, etc) that each over 700MB each which we can try strip in our CI tests at least.

Stebalien:
FYI, https://github.com/filecoin-project/ref-fvm/issues/1705 is also likely related.

PR #1711, opened 2023-03-13
fix: Github integration tests failed due to running out of disk space

fridrik01:
This commit overrides the development profile for our fvm_gas_calibration integration test so it strips all the generated binaries (6 in total) of debug symbols.

This reduces disk space usage by over 4GB so it should give us enough headroom for running integration tests on Gibhug (which only has 12GB SSD).

fixes: https://github.com/filecoin-project/ref-fvm/issues/1710

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -20,3 +20,7 @@ overflow-checks = true
 lto = true
 opt-level = "z"
 #strip = true
+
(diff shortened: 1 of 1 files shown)


Stebalien:
also see https://github.com/filecoin-project/ref-fvm/pull/1712

Stebalien (approved):

PR #1712, opened 2023-03-13
ci: disable incremental build in CI

Stebalien:
- This causes build failures in CI (at least with sccache). We had disabled this before, but I re-enabled it when I upgraded CI.
- This is likely increasing cache sizes, and just isn't worth it.


Stebalien:
Ok, I'm not sure if this actually fixes the macos build (https://github.com/pl-strflt/rust-sccache-action/pull/14 may also be related). But it's a step in the right direction (I think)

PR #1716, opened 2023-03-16
Refactor calibration binaries into integration tests

fridrik01:
This commit removes all the calibration test binaries and replaces them with integration tests with same exact functionality.

This reduces the build artifact size by 0.5GB and makes it easier to run these tests in the future (they are left disabled for now in CI behind the features flag "calibration)


fridrik01:
Rebased

Stebalien (approved):

PR #1719, opened 2023-03-16
test crate renames and small fixes

Stebalien:
This required some "let's see what builds" development, so it was easier to do it as a patch than as a review:

- Rename wasm builder to "test actors", and move the actors into that crate. Otherwise, it looks like this crate is a general-purpose wasm builder, but it's not. It's a pseudo-bundle for the test actors.
- Ensure that we can continue publishing the integration test suite. Importantly, don't depend on git repos.
- Remove the cargo dependencies on the actors, it doesn't actually help anything and just results in warnings.
- Mark test crates as "not for publishing".
- Remove build hack from the conformance tester, and make it depend on the test actors crate.


testing/calibration/Cargo.toml
```diff
@@ -32,6 +33,7 @@ thiserror = "1.0.30"
 bls-signatures = { version = "0.13", default-features = false }
 blake2b_simd = "1.0.0"
 wat = "1.0.52"
+fvm_test_actors = { path = "../test_actors" }
```
Stebalien, line 36: This ensures that we actually build the calibration test actor.

testing/calibration/build.rs
```diff
@@ -1,9 +0,0 @@
-// Copyright 2021-2023 Protocol Labs
-// SPDX-License-Identifier: Apache-2.0, MIT
-
-fn main() {
-    // intentionally empty here since fil_gas_calibration_actor is already built by
```
Stebalien, line 5: Now that we've specified the correct dependencies, we don't need this.

testing/integration/Cargo.toml
```diff
@@ -34,26 +33,11 @@ default-features = false
 features = ["cranelift", "parallel-compilation"]
 
 [dev-dependencies]
-fvm_test_wasm_builder = { path = "../wasm_builder" }
+actors-v10 = { package = "fil_builtin_actors_bundle", git = "https://github.com/filecoin-project/builtin-actors", branch = "next", features = ["m2-native"] }
+fvm_test_actors = { path = "../test_actors" }
 wat = "1.0.51"
 hex = "0.4.3"
 
-[target.'cfg(target_arch = "wasm32")'.dependencies]
```
Stebalien, line 41: Turns out these weren't actually doing anything for us, so I've removed them.

testing/test_actors/Cargo.toml
```diff
@@ -1,8 +1,9 @@
 [package]
-name = "fvm_test_wasm_builder"
+name = "fvm_test_actors"
```
Stebalien, line 2: Renamed to make it clear that this isn't a general-purpose "wasm builder".

Stebalien:
Pushed.

PR #1726, opened 2023-03-24
feat: introduce network versions 19 and 20

arajasek:
As in the title.

### DIFF
--- a/shared/src/version/mod.rs
+++ b/shared/src/version/mod.rs
@@ -11,6 +11,7 @@ use serde::{Deserialize, Serialize};
 #[repr(transparent)]
 #[serde(transparent)]
 pub struct NetworkVersion(u32);
+
 impl NetworkVersion {
     /// genesis (specs-actors v0.9.3)
     pub const V0: Self = Self(0);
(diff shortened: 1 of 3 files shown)


shared/src/version/mod.rs
```diff
@@ -50,6 +51,10 @@ impl NetworkVersion {
     pub const V17: Self = Self(17);
     /// Hygge (builtin-actors v10)
     pub const V18: Self = Self(18);
+    /// Lightning (builtin-actors v11)
+    pub const V19: Self = Self(19);
+    /// Lightning (builtin-actors v11)
```
ZenGround0, line 56: Idk how we're thinking about actors versioning now but wouldn't this be v12?

ZenGround0 (approved):

shared/src/version/mod.rs
arajasek: No, because FIP-0061 proposes a 2-stage upgrade, with a rollover period.

Stebalien: Call the second one Thunder, maybe?

Stebalien (approved):

shared/src/version/mod.rs
arajasek: That's such a good idea, it's worth re-running CI.

PR #1727, opened 2023-03-24
feat: FIP-0061: Add new PoSt proof types

arajasek:
Needs #1726 and #1728

### DIFF
--- a/fvm/Cargo.toml
+++ b/fvm/Cargo.toml
@@ -27,7 +27,7 @@ serde_tuple = "0.5"
 lazy_static = "1.4.0"
 derive_more = "0.99.17"
 replace_with = "0.1.7"
-filecoin-proofs-api = { version = "13", default-features = false }
+filecoin-proofs-api = { version = "14", default-features = false }
 rayon = "1"
 num_cpus = "1.13.0"
 log = "0.4.14"
--- a/shared/src/sector/registered_proof.rs
+++ b/shared/src/sector/registered_proof.rs
@@ -108,6 +108,11 @@ pub enum RegisteredPoStProof {
     StackedDRGWindow512MiBV1,
     StackedDRGWindow32GiBV1,
     StackedDRGWindow64GiBV1,
+    StackedDRGWindow2KiBV1P1,
+    StackedDRGWindow8MiBV1P1,
+    StackedDRGWindow512MiBV1P1,
+    StackedDRGWindow32GiBV1P1,
+    StackedDRGWindow64GiBV1P1,
     Invalid(i64),
 }
 
@@ -116,11 +121,21 @@ impl RegisteredPoStProof {
     pub fn sector_size(self) -> Result<SectorSize, String> {
         use RegisteredPoStProof::*;
         match self {
-            StackedDRGWindow2KiBV1 | StackedDRGWinning2KiBV1 => Ok(SectorSize::_2KiB),
-            StackedDRGWindow8MiBV1 | StackedDRGWinning8MiBV1 => Ok(SectorSize::_8MiB),
-            StackedDRGWindow512MiBV1 | StackedDRGWinning512MiBV1 => Ok(SectorSize::_512MiB),
-            StackedDRGWindow32GiBV1 | StackedDRGWinning32GiBV1 => Ok(SectorSize::_32GiB),
-            StackedDRGWindow64GiBV1 | StackedDRGWinning64GiBV1 => Ok(SectorSize::_64GiB),
+            StackedDRGWindow2KiBV1P1 | StackedDRGWindow2KiBV1 | StackedDRGWinning2KiBV1 => {
+                Ok(SectorSize::_2KiB)
+            }
+            StackedDRGWindow8MiBV1P1 | StackedDRGWindow8MiBV1 | StackedDRGWinning8MiBV1 => {
+                Ok(SectorSize::_8MiB)
+            }
+            StackedDRGWindow512MiBV1P1 | StackedDRGWindow512MiBV1 | StackedDRGWinning512MiBV1 => {
+                Ok(SectorSize::_512MiB)
+            }
+            StackedDRGWindow32GiBV1P1 | StackedDRGWindow32GiBV1 | StackedDRGWinning32GiBV1 => {
+                Ok(SectorSize::_32GiB)
+            }
+            StackedDRGWindow64GiBV1P1 | StackedDRGWindow64GiBV1 | StackedDRGWinning64GiBV1 => {
+                Ok(SectorSize::_64GiB)
+            }
             Invalid(i) => Err(format!("unsupported proof type: {}", i)),
         }
     }
@@ -138,7 +153,12 @@ impl RegisteredPoStProof {
             | StackedDRGWindow8MiBV1
             | StackedDRGWindow512MiBV1
             | StackedDRGWindow32GiBV1
-            | StackedDRGWindow64GiBV1 => Ok(192),
+            | StackedDRGWindow64GiBV1
+            | StackedDRGWindow2KiBV1P1
+            | StackedDRGWindow8MiBV1P1
+            | StackedDRGWindow512MiBV1P1
+            | StackedDRGWindow32GiBV1P1
+            | StackedDRGWindow64GiBV1P1 => Ok(192),
             Invalid(i) => Err(format!("unsupported proof type: {}", i)),
         }
     }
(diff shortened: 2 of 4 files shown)


shared/src/sector/registered_proof.rs
```diff
@@ -130,19 +145,19 @@ impl RegisteredPoStProof {
     pub fn registered_seal_proof(self) -> Result<RegisteredSealProof, String> {
         use RegisteredPoStProof::*;
         match self {
-            StackedDRGWindow64GiBV1 | StackedDRGWinning64GiBV1 => {
+            StackedDRGWindow64GiBV1P1 | StackedDRGWindow64GiBV1 | StackedDRGWinning64GiBV1 => {
```
ZenGround0, line 148: Not sure what's going on here but it seems off that we're not returning V1P1 seal proofs

shared/src/sector/registered_proof.rs
```diff
@@ -172,16 +192,58 @@ impl RegisteredPoStProof {
...
+            Ok(RegisteredPoStProof::StackedDRGWindow8MiBV1)
+        }
+        RegisteredPoStProof::StackedDRGWindow512MiBV1P1 => {
+            Ok(RegisteredPoStProof::StackedDRGWindow512MiBV1)
+        }
+        RegisteredPoStProof::StackedDRGWindow32GiBV1P1 => {
+            Ok(RegisteredPoStProof::StackedDRGWindow32GiBV1)
+        }
+        RegisteredPoStProof::StackedDRGWindow64GiBV1P1 => {
+            Ok(RegisteredPoStProof::StackedDRGWindow64GiBV1P1)
```
ZenGround0, line 229: Typo `StackedDRGWindow64GiBV1`

shared/src/sector/registered_proof.rs
```diff
@@ -233,11 +295,11 @@ impl RegisteredSealProof {
...
-            Self::StackedDRG64GiBV1 | Self::StackedDRG64GiBV1P1 => Ok(StackedDRGWindow64GiBV1),
-            Self::StackedDRG32GiBV1 | Self::StackedDRG32GiBV1P1 => Ok(StackedDRGWindow32GiBV1),
-            Self::StackedDRG2KiBV1 | Self::StackedDRG2KiBV1P1 => Ok(StackedDRGWindow2KiBV1),
-            Self::StackedDRG8MiBV1 | Self::StackedDRG8MiBV1P1 => Ok(StackedDRGWindow8MiBV1),
-            Self::StackedDRG512MiBV1 | Self::StackedDRG512MiBV1P1 => Ok(StackedDRGWindow512MiBV1),
+            Self::StackedDRG64GiBV1 | Self::StackedDRG64GiBV1P1 => Ok(StackedDRGWindow64GiBV1P1),
+            Self::StackedDRG32GiBV1 | Self::StackedDRG32GiBV1P1 => Ok(StackedDRGWindow32GiBV1P1),
+            Self::StackedDRG2KiBV1 | Self::StackedDRG2KiBV1P1 => Ok(StackedDRGWindow2KiBV1P1),
+            Self::StackedDRG8MiBV1 | Self::StackedDRG8MiBV1P1 => Ok(StackedDRGWindow8MiBV1P1),
+            Self::StackedDRG512MiBV1 | Self::StackedDRG512MiBV1P1 => Ok(StackedDRGWindow512MiBV1P1),
```
ZenGround0, line 302: ✅

shared/src/version/mod.rs
```diff
@@ -50,6 +51,10 @@ impl NetworkVersion {
     pub const V17: Self = Self(17);
     /// Hygge (builtin-actors v10)
     pub const V18: Self = Self(18);
+    /// Lightning (builtin-actors v11)
+    pub const V19: Self = Self(19);
+    /// Lightning (builtin-actors v11)
+    pub const V20: Self = Self(20);
```
ZenGround0, line 57: I don't understand why we need to introduce 20 right now, afaik the fip 61 PR just references nv19, but it seems ok to do so.

shared/src/sector/registered_proof.rs
ZenGround0: Can't find any callers in builtin actors, is it possible we can just kill this function?

ZenGround0 (approved):
You'll probably want someone who knows ref-fvm a little better but once the typo is fixed this LGTM.

If you have time please explore removing `registered_seal_proof` or updating it to use current seal proofs.  It would also be helpful to get an explanation for why we need n20 specified here.

shared/src/version/mod.rs
ZenGround0: Reason is tests

fvm/Cargo.toml
```diff
@@ -31,7 +31,7 @@ lazy_static = "1.4.0"
 derive-getters = "0.2.0"
 derive_more = "0.99.17"
 replace_with = "0.1.7"
-filecoin-proofs-api = { version = "13", default-features = false }
+filecoin-proofs-api = { version = "14", default-features = false }
```
Stebalien, line 34: Is this network breaking? I.e., will updating to version 14 cause a hard-fork?

shared/src/sector/registered_proof.rs
arajasek: Opened https://github.com/filecoin-project/ref-fvm/pull/1728 for this

shared/src/sector/registered_proof.rs
```diff
@@ -108,6 +108,11 @@ pub enum RegisteredPoStProof {
     StackedDRGWindow512MiBV1,
     StackedDRGWindow32GiBV1,
     StackedDRGWindow64GiBV1,
+    StackedDRGWindow2KiBV1P1,
```
Stebalien, line 111: Nit but... can't we just call these V2?

shared/src/sector/registered_proof.rs
```diff
@@ -172,16 +168,58 @@ impl RegisteredPoStProof {
...
+            StackedDRGWinning512MiBV1 | StackedDRGWindow512MiBV1 | StackedDRGWindow512MiBV1P1 => {
+                Ok(2)
+            }
             Invalid(i) => Err(format!("unsupported proof type: {}", i)),
         }
     }
 }
 
+/// Convert the v1_1 PoSt Proof type to the older v1 types (used in nv18 and below)
+pub fn convert_window_post_proof_v1p1_to_v1(
```
Stebalien, line 188: Is there any reason to put this here rather in, e.g., the builtin actors? I'd like to shrink the amount of proofs logic in this crate in general.

shared/src/sector/registered_proof.rs
```diff
@@ -172,16 +168,58 @@ impl RegisteredPoStProof {
...
+        }
+        RegisteredPoStProof::StackedDRGWindow64GiBV1P1 => {
+            Ok(RegisteredPoStProof::StackedDRGWindow64GiBV1P1)
+        }
+        i => Err(format!("not a v1p1 proof type: {:?}", i)),
+    }
+}
+
+/// Convert the v1_1 PoSt Proof type to the older v1 types (used in nv18 and below)
+pub fn is_window_post_proof_v1p1(rpp: RegisteredPoStProof) -> bool {
```
Stebalien, line 212: Ditto above.

fvm/Cargo.toml
arajasek: The answer _should_ be no, but I will let @cryptonemo confirm.

shared/src/sector/registered_proof.rs
arajasek: We can, and honestly we should -- I was following the pattern we have the versioned seal proofs above (hence the V1P1 name too), but I'm quite happy to break the pattern instead.

shared/src/sector/registered_proof.rs
arajasek: E A G L E E Y E S

shared/src/sector/registered_proof.rs
Stebalien: Ah... Well, if we already have the pattern of versions and are using them in the crypto code... it's probably not worth it?

fvm/Cargo.toml
cryptonemo: The answer is *yes*, this is `nv19` scope and beyond.

cryptonemo: To clarify, it's yes once the new PoSt Proof types are used/accepted.  If you continue using the old PoSt proof types in the new version of proofs, it's compatible.

shared/src/sector/registered_proof.rs
arajasek: I'm gonna keep the naming pattern for now, because I think diverging there will be confusing. But I will move the helpers to actors code, because they're only used there, and I don't want anything else depending on it.

Stebalien (approved):

shared/src/sector/registered_proof.rs
```diff
@@ -371,6 +402,11 @@ impl TryFrom<RegisteredPoStProof> for filecoin_proofs_api::RegisteredPoStProof {
             StackedDRGWindow512MiBV1 => Ok(Self::StackedDrgWindow512MiBV1),
             StackedDRGWindow32GiBV1 => Ok(Self::StackedDrgWindow32GiBV1),
             StackedDRGWindow64GiBV1 => Ok(Self::StackedDrgWindow64GiBV1),
+            StackedDRGWindow2KiBV1P1 => Ok(Self::StackedDrgWindow2KiBV1),
+            StackedDRGWindow8MiBV1P1 => Ok(Self::StackedDrgWindow8MiBV1),
+            StackedDRGWindow512MiBV1P1 => Ok(Self::StackedDrgWindow512MiBV1),
+            StackedDRGWindow32GiBV1P1 => Ok(Self::StackedDrgWindow32GiBV1),
+            StackedDRGWindow64GiBV1P1 => Ok(Self::StackedDrgWindow64GiBV1),
```
0x5459, line 409:
```diff
            StackedDRGWindow2KiBV1P1 => Ok(Self::StackedDrgWindow2KiBV1_2),
            StackedDRGWindow8MiBV1P1 => Ok(Self::StackedDrgWindow8MiBV1_2),
            StackedDRGWindow512MiBV1P1 => Ok(Self::StackedDrgWindow512MiBV1_2),
            StackedDRGWindow32GiBV1P1 => Ok(Self::StackedDrgWindow32GiBV1_2),
            StackedDRGWindow64GiBV1P1 => Ok(Self::StackedDrgWindow64GiBV1_2),
```
typo?

0x5459: @Stebalien

PR #1728, opened 2023-03-27
refactor: drop unused registered_seal_proof method

arajasek:
Seems unused across both ref-fvm and builtin-actors.


Stebalien:
@lemmih are you using this?

arajasek:
> @lemmih are you using this?

Note that the value returned right now is (arguably) wrong, since it should probably be returning the V1P1 seal proof types.

lemmih (approved):
Not used by Forest, either.

PR #1729, opened 2023-03-27
chore: update and cleanup deps

Stebalien:
1. Update all deps. We haven't done this in a while.
2. Replace `pretty_env_logger` with `env_logger`. The former depends on a very outdated version of `env_logger` and doesn't look maintained.


fridrik01 (approved):
LGTM

PR #1732, opened 2023-03-29
feat: add new PoSt proof types to price list

arajasek:


Stebalien (approved):
Note to future readers: this is safe because the actors themselves restrict the allowed post versions before calling this API.

Issue #979, opened 2022-10-18
Measure test coverage in integration tests in CI


Stebalien:
We took some steps towards implementing it, but never integrated it with CI. We need to run the `integration` test with code coverage enabled.

Stebalien:
Current status: we can get part of the way there by testing with:

```bash
cargo llvm-cov \
              --coverage-target-only \
              --target x86_64-unknown-linux-gnu
              --package fvm_integration_tests \
              --package "*actor" \
              --lcov --output-path itest-lcov.info
```

Unfortunately, substrate-wasm-builder seems to _insist_ on picking up the build flags from the rest of the project, even though it shouldn't.

I've tried updating substrate-wasm-builder, but that pulls in wasm-opt which makes the build times impossibly slow (it has to build wasm-opt from scratch, and seems to be doing it over and over and over again.

So, I believe the first step here will be to ditch substrate-wasm-builder. We definitely don't need it to build our test actors and might as well go with something a bit simpler.

PR #1734, opened 2023-04-03
Enable code coverage in integration tests

fridrik01:
This PR enables code coverage for our integration tests, it seems to bump up our coverage from to 55% to 75% ([link](https://app.codecov.io/github/filecoin-project/ref-fvm/commit/167dc3b4de9c1086e64f4f67a640e4497c2e2887))

Fixes: https://github.com/filecoin-project/ref-fvm/issues/979

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -53,9 +53,10 @@ jobs:
             command: llvm-cov
             args: --all --exclude fvm --exclude fvm_conformance_tests --exclude fvm_integration_tests --exclude "*actor" --lcov --output-path lcov.info
           - name: integration
-            key: v3
-            command: test
-            args: --package fvm_integration_tests --package "*actor"
+            key: v3-cov
+            covname: itest-lcov.info
+            command: llvm-cov
+            args: --package fvm_integration_tests --package "*actor" --lcov --output-path itest-lcov.info
           - name: conformance
             key: v3
             command: test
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -102,7 +103,6 @@ jobs:
       uses: codecov/codecov-action@d9f34f8cd5cb3b3eb79b3e4b5dae3a16df499a70
       with:
         files: ${{ matrix.covname }}
-        token: ${{ secrets.CODECOV_TOKEN }}
```
fridrik01, line 105: Removed as we should not need a token for public repo

.github/workflows/ci.yml
```diff
@@ -103,6 +104,8 @@ jobs:
       with:
         files: ${{ matrix.covname }}
         token: ${{ secrets.CODECOV_TOKEN }}
+        fail_ci_if_error: true
+        verbose: true
```
fridrik01, line 108: I would prefer to fail the CI if the code coverage fails (which _should_ be rare), and if it does then we have extra verbose logs

.github/workflows/ci.yml
Stebalien: 🤷‍♂️

.github/workflows/ci.yml
Stebalien: It's unfortunately not too rare (codecov randomly fails to accept coverage reports). I'm happy to try re-enabling this, but we may need to turn it back off in the future.

Stebalien (approved):
🥳

PR #1735, opened 2023-04-03
feat: kernel: use specified proof type when verifying PoSts

arajasek:
We no longer have only one PoSt proof type corresponding to each Seal proof type. As a result, when verifying PoSts, we must supply the proof type that the proof "announces" as the public Replica's PoSt proof type. This is still secure as proofs code checks that all replicas have the same proof type, and actors code confirms that the proof type of the first PoSt is valid.

If we wanted to, we could revert this change in FVM4, at which point we'll have a one PoSt per Seal restriction again. If we _really_ wanted to, we could make this code network version dependent, and only apply this change to nv19, but I think that belongs in actors code.

### DIFF
--- a/fvm/src/kernel/default.rs
+++ b/fvm/src/kernel/default.rs
@@ -18,7 +18,7 @@ use fvm_shared::econ::TokenAmount;
 use fvm_shared::error::ErrorNumber;
 use fvm_shared::event::ActorEvent;
 use fvm_shared::piece::{zero_piece_commitment, PaddedPieceSize};
-use fvm_shared::sector::SectorInfo;
+use fvm_shared::sector::{RegisteredPoStProof, SectorInfo};
 use fvm_shared::sys::out::vm::ContextFlags;
 use fvm_shared::{commcid, ActorID};
 use lazy_static::lazy_static;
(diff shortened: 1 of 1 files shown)


Stebalien (approved):

fvm/src/kernel/default.rs
```diff
@@ -1173,8 +1162,10 @@ fn verify_post(verify_info: &WindowPoStVerifyInfo) -> Result<bool> {
     // Necessary to be valid bls12 381 element.
     randomness[31] &= 0x3f;
 
+    let proof_type = proofs[0].post_proof;
```
Stebalien, line 1165: I'd slightly prefer to validate that all of the proof types here are the same.

fvm/src/kernel/default.rs
```diff
@@ -1114,18 +1107,14 @@ fn get_required_padding(
...
 ) -> Result<BTreeMap<SectorId, PublicReplicaInfo>> {
     let replicas = src
         .iter()
         .map::<core::result::Result<(SectorId, PublicReplicaInfo), String>, _>(
             |sector_info: &SectorInfo| {
                 let commr = commcid::cid_to_replica_commitment_v1(&sector_info.sealed_cid)?;
-                let proof = match typ {
-                    ProofType::Winning => sector_info.proof.registered_winning_post_proof()?,
-                    ProofType::Window => sector_info.proof.registered_window_post_proof()?,
-                };
```
Stebalien, line 1127: Is there some way to quickly check that the proof type is valid for the given sectors?

Stebalien (approved):

Stebalien:
My concern is that... that code is _very_ far away from this code. I'd rather just have the extra checks, honestly.

arajasek:
@Stebalien Yup, happy to do so.

Stebalien (approved):
Ah, I see why you weren't thrilled about this. But I think it's safer to have for now...

Issue #358, opened 2022-03-07
ref-fvm version philosophy 


jennijuju:
Starting a discussion for the ref-fvm version philosophy.

Things to consider:
-  do we associate FVM version w/ network upgrade versions?
- what's the approach that may support multiple fvm version in the future.
- what's the FVM version would be for network v16 M1 FVM?

raulk:
There are roughly two approaches here:

1. One FVM release per network version, with client importing all and selecting which version of the FVM to instantiate depending on the network version.
    - Pros: theoretically cleaner; every network version is self contained.
    - Cons: the integration of through FFI, etc. is going to be quite hard; higher complexity; binary bloat.
2. Single FVM import, with support for network versions added through version guards (like today).
    - Pros: much simpler; easier to reason about; easier to develop / manage the codebase; smaller binary sizes; no FFI drama.
    - Cons: less clean; no isolation/self-containment.

I am more agreeable to (2), and it's going to take quite a bit of effort to convince me that (1) is the better approach.

raulk:
@Stebalien Please add your input here, and let's aim to close this issue with a final decision.

Stebalien:
First, because I'm pedantic and want us to be on the same page:

**Multiple FVMs:**

- Bloat: This isn't correct. It'll all be in a single crate so the only "bloat" will be from the FVM code itself, not the dependencies. The FVM logic itself is pretty tiny and many of the types actually live in "shared".
- Integration in the FFI: this is pretty easy. We'll need a small abstraction layer, but that comes down to: a constructor, an "apply message" method, and a "flush" method. We'd also likely want to break out the "externs" interface into a new crate, but that's not difficult.
- High complexity: I'm not sure how not having to support old network versions in the FVM will lead to _higher_ complexity.

**Single FVM:**

- Simpler: See my previous comment.
- Easier to reason about: Multiple FVM versions means less complexity in each version, which should be easier to reason about. Unless you mean "easier to reason about what changed", in which case I agree.
- Easier to develop/manage: I agree.

So really, the question is development. If we have multiple versions:

1. Backporting fixes becomes annoying.
2. The "new" version won't be stress-tested on the network until the network upgrade.
3. It can get harder to see what was "introduced" in the new network version.

But I think I have a reasonably balanced solution:

1. Develop the "next" version with version conditionals.
2. After the network upgrade, release a new FVM version that _doesn't_ support previous network versions.

This has some advantages:

[...]

PR #1738, opened 2023-04-04
chore: release fvm/fvm_shared/fvm_sdk 3.2

Stebalien:
For nv19.


arajasek:
@Stebalien We're skipping SDK 3.1.0?

Stebalien:
Yes. I try to keep the releases aligned where possible (I probably should have released an empty 3.1.0, but it didn't seem necessary?).

arajasek:
Makes sense -- might be worth writing down as a general hint in the contribution guidelines.

Stebalien:
I'm hoping to address https://github.com/filecoin-project/ref-fvm/issues/358 before that comes up.

Issue #1699, opened 2023-03-09
Fix broken calibration test


fridrik01:
The `on_event_evm_shapes` test binary in `testing/calibration` is failing to pass tests:

```
cargo run --release --bin on_event_evm_shapes
thread 'main' panicked at 'message execution failed: message failed with backtrace:
00: f010000 (method 6) -- error running method: illegal argument (22)
--> caused by: event::emit_event -- event total values exceeded max size: {total_value_size} > {MAX_DATA} (1: illegal argument)
', /Users/fridrik/workspace/ref-fvm/testing/calibration/src/lib.rs:70:13
```
We should look into the validation added in https://github.com/filecoin-project/ref-fvm/pull/1638 and update the test accordingly.

Stebalien:
Hm. We should probably run those tests in CI as well. Nice catch!

PR #1739, opened 2023-04-05
Add calibration tests to CI

fridrik01:
This PR runs the calibration tests in our CI pipeline. 

Running these tests creates output artifacts (png, json) many of which are marked as changes/new files to the git repo. I don't think we want to store these in git so added all of the `measurements/out` as .gitignore, this also makes your local repo clean after running these tests locally.

Note that there is one failing tests in the calibration suite, fixing that is tracked in https://github.com/filecoin-project/ref-fvm/issues/1699

Fixes: https://github.com/filecoin-project/ref-fvm/issues/1740


Stebalien:
LGTM but let's fix or disable the failing test first. Red-CI tends to get... redder.

fridrik01:
> LGTM but let's fix or disable the failing test first. Red-CI tends to get... redder.

Disabled the test in this PR, but pushed a new PR which addressed the failing test, see https://github.com/filecoin-project/ref-fvm/pull/1743

PR #1744, opened 2023-04-07
chore: update cache action

Stebalien:
This new version cleans the cache.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -102,15 +102,11 @@ jobs:
       with:
         tool: cargo-llvm-cov@0.4.5
     - name: Setting up cache
-      uses: pl-strflt/rust-sccache-action@v1
-      env:
-        SCCACHE_CACHE_SIZE: 2G
-        SCCACHE_DIR: ${{ github.workspace }}/.cache/sccache
-        CACHE_SKIP_SAVE: ${{ matrix.push == '' || matrix.push == 'false' }}
+      uses: Swatinem/rust-cache@6fd3edff6979b79f87531400ad694fb7f2c84b1f
       with:
-        version: v0.2.15
+        save-if: ${{ matrix.push == 'true' }}
         # change this to invalidate sccache for this job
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -102,15 +102,11 @@ jobs:
       with:
         tool: cargo-llvm-cov@0.4.5
     - name: Setting up cache
-      uses: pl-strflt/rust-sccache-action@v1
-      env:
-        SCCACHE_CACHE_SIZE: 2G
-        SCCACHE_DIR: ${{ github.workspace }}/.cache/sccache
-        CACHE_SKIP_SAVE: ${{ matrix.push == '' || matrix.push == 'false' }}
+      uses: Swatinem/rust-cache@6fd3edff6979b79f87531400ad694fb7f2c84b1f
```
galargh, line 105: FYI, I removed the deprecated GitHub Actions usage (https://github.com/pl-strflt/rust-sccache-action/pull/16) and added autoclean feature (https://github.com/pl-strflt/rust-sccache-action/pull/18) to `rust-sccache-action`.

So if we wanted to stick with incremental builds we could go with:

```suggestion
      uses: pl-strflt/rust-sccache-action@v1
      env:
        SCCACHE_CACHE_SIZE: 2G
        SCCACHE_DIR: ${{ github.workspace }}/.cache/sccache
      with:
        version: v0.2.15
        # change this to invalidate sccache for this job
        shared-key: ${{ matrix.key }}
        skip-save: ${{ matrix.push == '' || matrix.push == 'false' }}
        autoclean: true
```

Otherwise, this looks good 👍

galargh (approved):

.github/workflows/ci.yml
Stebalien: Ah, wonderful. I'll try that first.

Stebalien: Ah, nevermind. That's what we were already doing, but it doesn't seem to _actually_ keep the cache clean. It just grows and grows.

Issue #1198, opened 2022-12-05
Upgrade Wasmtime


Stebalien:

maciejwitowski:
Will be closed by https://github.com/filecoin-project/ref-fvm/pull/1412

maciejwitowski:
@Stebalien I see this is being addressed by https://github.com/filecoin-project/ref-fvm/pull/1672 but that PR seems abandoned. What's the status?

Stebalien:
Basically, we had a patch but changes in wasmtime made it difficult to distinguish between different error types. Right now the only error types are "illegal instruction" and "fatal", so we can probably safely fold them into "illegal instruction", but we still need to write some careful tests.

Stebalien:
Needs backport to v2.

Stebalien:
Current status:

1. Re-enabling conformance tests (v3).
2. Need to cut a blockstore release (v3).
3. Backport conformance tests fixes to v2.
4. Remove integration tests from v2.
5. Backport this to v2.

Stebalien:
Ok, this has been released and bubbled all the way to the FVM (and tested by syncing both pre-hygge and post-hygge). I'm declaring it done.

PR #1745, opened 2023-04-07
chore: update to wasmtime 7.0.0

Stebalien:
Now that we've shipped FEVM, we should get this done.

- This new version _should_ be compatible with mainnet.
- We can no longer be precise when handling instantiation errors, but we can still detect "type checking" versus "instantiation" errors. We treat the former as fatal, and the latter as an "illegal instruction" (which also covers out of memory cases).

Fixes https://github.com/filecoin-project/ref-fvm/issues/1198


Stebalien:
This PR seems to be syncing the network just fine.

Stebalien:
Still syncing just fine.

fvm/src/syscalls/error.rs
```diff
@@ -50,59 +50,37 @@ impl Abort {
...
-    fn wrap(a: Abort) -> Self {
-        Self {
-            inner: Mutex::new(Some(a)),
+                // I think this is fatal? But I'm not sure.
+                | Trap::StackOverflow => Abort::Exit(
+                    ExitCode::SYS_ILLEGAL_INSTRUCTION,
+                    trap.to_string(),
+                    NO_DATA_BLOCK_ID,
+                ),
+                _ => Abort::Fatal(anyhow!("unexpected wasmtime trap: {}", trap)),
```
magik6k, line 78: Nit: Do we actually want a default case here? I feel like it may be better to be very explicit here, and make things not build if we don't handle all traps explicitly.

magik6k (approved):

fvm/src/syscalls/error.rs
Stebalien: I'd love to... but the enum is marked as "non exhaustive".

PR #1748, opened 2023-04-23
fix: correctly convert `RegisteredPoStProof` type

0x5459:


hunjixin:
@Stebalien @raulk @anorth

arajasek (approved):

anorth:
This suffix style, especially `V1P1`, makes it very hard to visually spot differences here. I suggest following up here and/or in the proofs library to find something a little more distinctive.

zhiqiangxu:
> This suffix style, especially `V1P1`, makes it very hard to visually spot differences here. I suggest following up here and/or in the proofs library to find something a little more distinctive.

It would be even better if the codebase can stick to **the single point of reference principle**, say each module doesn't redefine what's essentially the same, but reference to the same one, this will eliminate such issue in the future and make the code more readable.

arajasek:
@Stebalien No, we don't, they've landed in the right branches. Thanks SO MUCH again to @0x5459 and @hunjixin for the fix.

PR #1750, opened 2023-04-23
fix: correctly convert RegisteredPoStProof type

arajasek:
#1748, but on the release branch


raulk (approved):
I'm not familiar with the actual new proof enum types on the fil proofs library, but this fix looks correct given that we were returning the old values before.

PR #1718, opened 2023-03-16
Switch to Rust stable

fridrik01:
Since we have now removed substrate-wasm-builder dependency we are able to switch to rust stable channel.

I needed to update some of the integration tests that dependended on unstable nightly features.

Fixes: https://github.com/filecoin-project/ref-fvm/issues/1717

PR #1751, opened 2023-04-23
ignore clippy warns happening with rust 1.69.0

raulk:
We switched to Rust stable in https://github.com/filecoin-project/ref-fvm/pull/1718. The new release of Rust stable introduced new clippy lints that started causing CI failures on our end.

Ignoring these clippy lints for now, let's discuss options here. If we want to keep the dynamic Rust stable usage, we should probably add a nightly build on master and release branches, to catch breakages sooner and avoid later surprises.


Stebalien:
> Note: If we want to keep the dynamic Rust stable usage, we should probably add a nightly build on master and release branches, to catch breakages sooner and avoid later surprises.

Honestly, I think this is a good enough reason to "freeze" to a specific rust release.

Stebalien (approved):

PR #1755, opened 2023-04-26
feat: reject V1P1 proofs on calibrationnet before cutoff

arajasek:
This resolves the inconsistent state on calibrationnet that resulted from the bug fixed in #1750. It's unfortunate to have to have a patch like this, but we can drop it when we graduate to FVM4.

### DIFF
--- a/fvm/src/kernel/default.rs
+++ b/fvm/src/kernel/default.rs
@@ -12,12 +12,14 @@ use fvm_ipld_blockstore::Blockstore;
 use fvm_ipld_encoding::{bytes_32, IPLD_RAW};
 use fvm_shared::address::Payload;
 use fvm_shared::bigint::Zero;
+use fvm_shared::chainid::ChainID;
 use fvm_shared::consensus::ConsensusFault;
 use fvm_shared::crypto::signature;
 use fvm_shared::econ::TokenAmount;
 use fvm_shared::error::ErrorNumber;
 use fvm_shared::event::ActorEvent;
 use fvm_shared::piece::{zero_piece_commitment, PaddedPieceSize};
+use fvm_shared::sector::RegisteredPoStProof::{StackedDRGWindow32GiBV1, StackedDRGWindow32GiBV1P1};
 use fvm_shared::sector::{RegisteredPoStProof, SectorInfo};
 use fvm_shared::sys::out::vm::ContextFlags;
 use fvm_shared::{commcid, ActorID};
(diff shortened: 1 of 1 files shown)


fvm/src/kernel/default.rs
```diff
@@ -539,6 +541,19 @@ where
             .call_manager
             .charge_gas(self.call_manager.price_list().on_verify_post(verify_info))?;
 
+        #[allow(clippy::collapsible_if)]
+        if self.call_manager.context().network.chain_id == ChainID::from(314159) {
+            if self.call_manager.context().epoch <= 498691
+                && verify_info.proofs[0].post_proof == StackedDRGWindow32GiBV1P1
+            {
+                let mut verify_info_copy = verify_info.clone();
```
arajasek, line 549: Other efforts to fail early, such as returning Err here, or panicking without calling `verify_post` lead to slightly different gas usage. This surprises me, since these operations seem to be uncharged, but :shrug:

fvm/src/kernel/default.rs
```diff
@@ -539,6 +541,19 @@ where
             .call_manager
             .charge_gas(self.call_manager.price_list().on_verify_post(verify_info))?;
 
+        #[allow(clippy::collapsible_if)]
+        if self.call_manager.context().network.chain_id == ChainID::from(314159) {
```
Stebalien, line 545: Let's define a constant for the chain ID to make it clear where this applies.

fvm/src/kernel/default.rs
```diff
@@ -539,6 +541,19 @@ where
             .call_manager
             .charge_gas(self.call_manager.price_list().on_verify_post(verify_info))?;
 
+        #[allow(clippy::collapsible_if)]
```
Stebalien, line 544: Let's add a comment explaining why we have this check.

fvm/src/kernel/default.rs
arajasek: I looked at the code and thoroughly convinced myself it did not, but...you're right. That's much nicer.

fvm/src/kernel/default.rs
```diff
@@ -539,6 +541,22 @@ where
...
             .charge_gas(self.call_manager.price_list().on_verify_post(verify_info))?;
 
+        // Due to a bug on calibrationnet, we have some _valid_ StackedDRGWindow32GiBV1P1
+        // proofs that were deemed invalid on chain. This was fixed WITHOUT a network version check.
+        // As a result, we need to explicitly consider all such proofs invalid, ONLY on calibrationnet,
+        // and ONLY before epoch 498691,
+        let calibnet_chain_id = ChainID::from(314159);
+
+        #[allow(clippy::collapsible_if)]
+        if self.call_manager.context().network.chain_id == calibnet_chain_id {
```
Stebalien, line 551: Any reason not to collapse these checks? `rustfmt` should align everything nicely. Also, we can get a reference to the context first:

```rust
let ctx = self.call_manager.context();
if ctx.network.chain_id == calibnet_chain_id
    && ctx.epoch <= 498691
    && ... {
```

Stebalien (approved):
One nit, lgtm.

fvm/src/kernel/default.rs
arajasek: @Stebalien I wanted to explicitly have it wrapped in a "ARE YOU CALIBNET" top-level block. But we are biking all over this shed.

Agreed with getting context first.

arajasek:
Unfortunately, switching to just returning `Ok(false)` doesn't work because there _are_ 3 V1P1 PoSts that succeeded on-chain, despite being verified as V1 proofs.

I don't fully understand this, asking the Proofs folks for insight. Regardless, we can just go back to cloning and mutating the infos.

fvm/src/kernel/default.rs
Stebalien: I don't _really_ care. I just mildly prefer appeasing the linter rather than fighting it.

Stebalien: But you're right, it makes sense to have a top-level check.

Stebalien (approved):
LGTM!

Stebalien (changes requested):
Wait, no, still need to check the index in `proofs`.

arajasek:
What do you want to check? I thought you just meant that the index 0 exists (which I have done)

Stebalien (approved):
nvm. Steven isn't looking

PR #1771, opened 2023-05-04
chore: update proofs (v2)

Stebalien:
Update the proofs library for the FVM v2. This will allow us to significantly shrink our binary size.


Stebalien:
I've validated that this syncs mainnet pre-hygge.

Issue #1779, opened 2023-05-18
Generics and transmutation make replacing Kernel very difficult


anorth:
@alexytsu and I are improving tooling for benchmarking and profiling native actors on the FVM (see https://github.com/anorth/fvm-workbench and https://github.com/filecoin-project/builtin-actors/issues/1236). We've run into great difficulty attempting to fake out expensive VM intrinsic operations like proof verification, the `CryptoOps` trait of `Kernel`.

We can the `DefaultKernel` and delegate most methods...
```
pub struct BenchKernel<B: Blockstore + 'static> {
    kernel: DefaultKernel<DefaultCallManager<DefaultMachine<B, FakeExterns>>>,
    // ...
}
let executor = DefaultExecutor::<BenchKernel<B>>::new(EnginePool::new_default(engine_conf)?, machine)?;
```
... but this  is only effective for the top-level call. [During a sub-call](https://github.com/filecoin-project/ref-fvm/blob/f4f3f340ba29b3800cd8272e34023606def23855/fvm/src/kernel/default.rs#L395), the wrapped `DefaultKernel` specifies a static `Self` as a type parameter to `CallManger::send`. No instance reference is passed (and indeed the kernel later consumes itself).

```
    let result = self.call_manager.with_transaction(|cm| {
        cm.send::<Self>(
            from, *recipient, method, params, value, gas_limit, read_only,
        )
    })?;
```

The `Kernel` trait specifies a `new()` constructor, and the call manager later uses that to construct a new kernel for the subcall, but it always constructs `DefaultKernel`.

With reference to some code in the conformance testing package, we tried also wrapping the CallManger so as to specify our own `BenchKernel` as `Self` instead, also replacing `with_transaction` to do a [funky transmutation](https://github.com/filecoin-project/ref-fvm/blob/f4f3f340ba29b3800cd8272e34023606def23855/testing/conformance/src/vm.rs#L236-L250). This also isn't doing the job. When the `DefaultCallManager` instantiates a Kernel inside send_resolved it passes itself in. Therefore, when the first Kernel is created, we get a `BenchKernel<DefaultCallManager>` rather than a `BenchKernel<BenchCallManager>` as needed. It looks like for the `BenchCallManager` to intercept this it would need to re-implement that entire call stack
`CallManager::send -> DefaultCallManager::send_unchecked -> DefaultCallManager::send_resolved`.

[...]

PR #1780, opened 2023-05-19
Add type parameter to Kernel::send specifying type for receiving kernel

anorth:
This type parameter supports a kernel to be wrapped, and have the wrapped type also be instantiated on the other side of the send. Something like this is necessary to support injecting kernel behaviour for testing.

@alexytsu and I couldn't get the TestKernel/TestCallManager pattern to work from github.com/anorth/fvm-workbench (where we need to fake out proof verification, and long term wish to be able to inject arbitrary kernel behaviour). This new scheme avoids the coupled TestCallManager and associated funky transmutation.

It's limited in that, unless we can figure out the right type constraints, a TestCallManager itself can't be wrapped, because it ignores the type parameter provided to it.

This is a partial resolution of #1779 (we're unblocked for now). But as noted there, still very limited because there's no facility for dynamic behaviour of an injected kernel. The kernel is constructed internally and short-lived. Passing a kernel construction method (i.e. factory) at runtime would be far more flexible.

### DIFF
--- a/fvm/src/kernel/default.rs
+++ b/fvm/src/kernel/default.rs
@@ -111,6 +111,70 @@ where
     fn machine(&self) -> &<Self::CallManager as CallManager>::Machine {
         self.call_manager.machine()
     }
+
+    fn send<K: Kernel<CallManager = C>>(
+        &mut self,
+        recipient: &Address,
+        method: MethodNum,
+        params_id: BlockId,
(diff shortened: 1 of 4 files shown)


Stebalien (approved):
I wish there were a better way... but this definitely makes testing easier.

PR #1802, opened 2023-06-28
feat: generate test actor "bundle"

Stebalien:
Instead of hard-coding it. This way, we can specify all actors in one place.


fridrik01 (approved):
LGTM

Issue #1637, opened 2023-02-03
Improved event syscall API


Stebalien:
Currently, events are emitted as CBOR objects. Unfortunately, this makes it hard to accurately charge gas for parsing.

**Proposal:**

Emit three buffers:

- `entries`: An array of `[(flags: u64, codec: u64, key_len: u32, value_len: u32), ...]`  (packed). I.e., 24 bytes per event.
- `keys`: A buffer of all keys, concatenated.
- `values`: A buffer of all values, concatenated.

The FVM will:

1. Compute the number of entries as `entries.len() / 24` (or return an error if not a multiple of 24).
2. Charge gas based on the number of entries, the length of `keys`, and the length of `values`.
3. Reconstruct the events from these parameters.

Importantly, this means the FVM can accurately charge gas before parsing anything.

aashidham:
Does this have any product implications with the emission of events? What about event listeners? 

Will this generally make sending events to the event log more or less expensive?

Stebalien:
This won't have any user-facing implications.

This issue is purely security/performance related. Although it may make emitting events less expensive as it may be less expensive to validate them? But that's unclear.

PR #1807, opened 2023-07-05
[WIP] feat: Improved event syscall API 

fridrik01:
Fixes: https://github.com/filecoin-project/ref-fvm/issues/1637

## Context 
TODO

## Test plan
TODO


fvm/src/kernel/default.rs
```diff
@@ -1006,41 +1007,102 @@ impl<C> EventOps for DefaultKernel<C>
...
+                Some(f) => f,
+                None => return Err(syscall_error!(IllegalArgument, "invalid flag").into()),
             };
-            t.stop();
-            res
-        }?;
-        validate_actor_event(&actor_evt)?;
+            let key_len = BigEndian::read_u32(&view[8..12]);
+            let codec = BigEndian::read_u64(&view[12..20]);
+            let val_len = BigEndian::read_u32(&view[20..24]);
```
Stebalien, line 1059: Given that we use little-endian for return values, we should likely do the same here. I.e., we can literally transmute the input array into an array of packed structs.

Stebalien:
```rust

// define this in shared
#[repr(C, packed)]
struct EventMeta {
    flags: Flags, // repr ut64
    codec: u64, // order matters for packing.
    key_len: u32,
    value_len: u32,
} 
```

Although... we'll likely need to transmute one-by-one and copy as the input slice _may_ not be aligned.

Stebalien: This is less critical on the FVM side, but it'll save time and avoid swapping around bytes in wasm.

fvm/src/kernel/default.rs
```diff
@@ -1006,41 +1007,102 @@ impl<C> EventOps for DefaultKernel<C>
...
-                Err(e) => {
-                    log::error!("panic when decoding event cbor from actor: {:?}", e);
-                    Err(syscall_error!(IllegalArgument; "panic when decoding event cbor from actor").into())
-                }
+        let total_buf_len = raw_evt.len() + raw_key.len() + raw_val.len();
+
+        let _t = self.call_manager.charge_gas(
+            self.call_manager
+                .price_list()
+                .on_actor_event_validate(total_buf_len),
```
Stebalien, line 1030: we'll likely need to revisit the costs here, but we can do that in a followup PR.

fvm/src/kernel/default.rs
Stebalien: Note: `Flags` will have to be decorated with `#[repr(transparent)]` for this to work.

fvm/src/kernel/default.rs
```diff
@@ -1006,41 +1006,99 @@ impl<C> EventOps for DefaultKernel<C>
...
+                );
+            }
+
+            // parse the variable sized fields from the raw_key/raw_val buffers
+            //
+            let key = match std::str::from_utf8(
+                &raw_key[key_offset..key_offset + entry_fixed.key_len as usize],
+            ) {
+                Ok(s) => s,
+                Err(_) => return Err(syscall_error!(IllegalArgument, "event key").into()),
```
Stebalien, line 1077:
```suggestion
                Err(e) => return Err(syscall_error!(IllegalArgument, "invalid event key: {}", e).into()),
```

(or something like that)

sdk/src/event.rs
```diff
@@ -1,12 +1,56 @@
...
+    let mut v = vec![0u8; evt.entries.len() * BYTES_PER_ENTRY];
+    for i in 0..evt.entries.len() {
+        let e = &evt.entries[i];
+        let offset = i * BYTES_PER_ENTRY;
+
+        let view = &mut v[offset..offset + BYTES_PER_ENTRY];
+        LittleEndian::write_u64(&mut view[..8], e.flags.bits());
+        LittleEndian::write_u64(&mut view[8..16], e.codec);
+        LittleEndian::write_u32(&mut view[16..20], e.key.len() as u32);
+        LittleEndian::write_u32(&mut view[20..24], e.value.len() as u32);
```
Stebalien, line 24: We should be able to use transmutes here as well. As a matter of fact, we should be able to just:

1. Make an array of `EntryFixed`.
2. Transmute by-reference.

....

Actually, we _should_ be able to just take `*mut EventFixed` (pointer to an _array_) in the syscall ABI itself and let rust do the transmute for us. Welcome to C!

PR #1808, opened 2023-07-05
Add support for HamtV0 from forest(commit hash: b622af5a6)

creativcoder:
More context: https://github.com/ChainSafe/forest/pull/3060#pullrequestreview-1503177969

### DIFF
--- a/ipld/hamt/src/hamt.rs
+++ b/ipld/hamt/src/hamt.rs
@@ -4,7 +4,6 @@
 
 use std::borrow::Borrow;
 use std::marker::PhantomData;
-use std::str::Bytes;
 
 use cid::Cid;
 use forest_hash_utils::BytesKey;
@@ -37,7 +36,7 @@ use crate::{version, Config, Error, Hash, HashAlgorithm, Sha256};
 /// ```
 pub type Hamt<BS, V, K = BytesKey> = HamtImpl<BS, V, version::V3, K, Sha256>;
 /// Legacy amt V0
-pub type Hamtv0<BS, V, K = BytesKey> = HamtImpl<V, BS, version::V0, K, Sha256>;
+pub type Hamtv0<BS, V, K = BytesKey> = HamtImpl<BS, V, version::V0, K, Sha256>;
 
 #[derive(Debug)]
 pub struct HamtImpl<BS, V, Ver, K = BytesKey, H = Sha256> {
--- a/ipld/hamt/src/hamt.rs
+++ b/ipld/hamt/src/hamt.rs
@@ -18,29 +18,6 @@ use crate::node::Node;
 use crate::pointer::version::Version;
 use crate::{version, Config, Error, Hash, HashAlgorithm, Sha256};
 
-// pub mod version {
-//     #[derive(PartialEq, Eq, Debug)]
-//     pub struct V0;
-//     #[derive(PartialEq, Eq, Debug)]
-//     pub struct V3;
-
-//     pub trait Version {
-//         const NUMBER: usize;
-//     }
-
-//     impl Version for V0 {
-//         const NUMBER: usize = 0;
-//     }
-
-//     impl Version for V3 {
-//         const NUMBER: usize = 3;
-//     }
-// }
-
-pub type Hamt<BS, V, K = BytesKey> = HamtImpl<BS, V, version::V3, K, Sha256>;
-/// Legacy amt V0
-pub type Hamtv0<BS, V, K> = HamtImpl<V, BS, version::V0, K, Sha256>;
-
 /// Implementation of the HAMT data structure for IPLD.
 ///
 /// # Examples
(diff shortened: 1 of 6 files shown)


ipld/hamt/src/lib.rs
```diff
@@ -23,12 +23,15 @@ pub use forest_hash_utils::{BytesKey, Hash};
...
-pub use self::hamt::Hamt;
+pub use self::hamt::{Hamt, Hamtv0};
 pub use self::hash::*;
 pub use self::hash_algorithm::*;
+pub use self::pointer::version;
 
 /// Default bit width for indexing a hash at each depth level
 const DEFAULT_BIT_WIDTH: u32 = 8;
+/// Default bit width for indexing a hash at each depth level for Hamt v0
+pub const DEFAULT_BIT_WIDTH_V0: u32 = 5;
```
Stebalien, line 34: Let's make this private unless we need it to be public for some reason.

ipld/hamt/src/pointer.rs
```diff
@@ -81,26 +133,57 @@ where
...
+        match Ver::NUMBER {
+            0 => {
+                let ipld = Ipld::deserialize(deserializer)?;
+                let (_key, value) = match ipld {
+                    Ipld::Map(map) => map
+                        .into_iter()
+                        .next()
+                        .ok_or("Expected at least one element".to_string()),
+                    other => Err(format!("Expected `Ipld::Map`, got {:#?}", other)),
+                }
```
Stebalien, line 155: We should decode this as a tagged enum instead of using `Ipld` and looking at the types.

```rust
#[derive(Deserialize, Serialize)]
enum PointerDe<K, V> {
    #[serde(name = "l")]
    Link(Cid),
    #[serde(name = "v")]
    Values(KeyValuePair<K, V>),
}
```

ipld/hamt/src/hamt.rs
```diff
@@ -33,21 +34,26 @@ use crate::{Config, Error, Hash, HashAlgorithm, Sha256};
...
 /// let cid = map.flush().unwrap();
 /// ```
+pub type Hamt<BS, V, K = BytesKey> = HamtImpl<BS, V, version::V3, K, Sha256>;
+/// Legacy amt V0
+pub type Hamtv0<BS, V, K = BytesKey> = HamtImpl<BS, V, version::V0, K, Sha256>;
+
 #[derive(Debug)]
-pub struct Hamt<BS, V, K = BytesKey, H = Sha256> {
-    root: Node<K, V, H>,
+pub struct HamtImpl<BS, V, Ver, K = BytesKey, H = Sha256> {
```
Stebalien, line 42: Does this need to be public? And if so, can we hide it in the docs?

ipld/hamt/src/pointer.rs
```diff
@@ -15,18 +15,37 @@ use super::node::Node;
 use super::{Error, Hash, HashAlgorithm, KeyValuePair};
 use crate::Config;
 
+pub mod version {
```
Stebalien, line 18: Let's make this crate-private if possible, or `#[doc(hidden)]` otherwise.

ipld/hamt/src/pointer.rs
creativcoder: i had to change l and v to 0 and 1 respectively to read genesis header in forest. 
Changes made: 
```
    #[derive(Serialize)]
    pub(super) enum PointerSer<'a, K, V> {
        #[serde(rename = "0")]
        Vals(&'a [KeyValuePair<K, V>]),
        #[serde(rename = "1")]
        Link(&'a Cid),
    }

    #[derive(Deserialize, Serialize)]
    pub enum PointerDe<K, V> {
        #[serde(rename = "0")]
        Link(Cid),
        #[serde(rename = "1")]
        Vals(Vec<KeyValuePair<K, V>>),
    }
```
But ipld_hamt_v0 tests fail with:
```
thread 'test_default::test_hamtv0' panicked at 'called `Result::unwrap()` on an `Err` value: Dynamic(Serialization error for Cbor protocol: TypeMismatch { name: "CBOR tag", byte: 1 })', ipld/hamt/tests/hamt_tests.rs:1022:69
```

Stebalien: Huh. I thought it was k/v. I must be thinking of IPFS HAMTs.

> But Hamtv0 tests in this repo fails with:

Hm. Something funky is going on there. Can you push the code?

Stebalien: (or push it to a new branch)

ipld/hamt/src/pointer.rs
```diff
@@ -37,25 +56,58 @@ impl<K: PartialEq, V: PartialEq, H> PartialEq for Pointer<K, V, H> {
...
+
+    use crate::KeyValuePair;
+
+    use super::Pointer;
+
+    #[derive(Serialize)]
+    pub(super) enum PointerSer<'a, K, V> {
+        Vals(&'a [KeyValuePair<K, V>]),
+        Link(&'a Cid),
+    }
```
Stebalien, line 71: I'm a bit confused at how this works. For v0, we want to serialize to a map of "0"/"1" to a cid/list of links. As far as I know, this here will serialize use "Vals" and "Link" as the keys?

ipld/hamt/src/pointer.rs
creativcoder: I've pushed the changes with failing hamtv0 tests.

ipld/hamt/src/lib.rs
Stebalien: I've removed this field due to #1828 and https://github.com/filecoin-project/builtin-actors/issues/1346. Basically, the user should just pass the bit-width they need and I'm going to deprecate the "default" methods as they're useless.

As far as I can tell, the v0 bit-width was also 8 by default. We just set it to 5 everywhere.

Stebalien (approved):

ipld/hamt/src/pointer.rs
Stebalien: It was a sneaky bug. The Vals/Pointer variants were swapped.

creativcoder: Ah i see, interesting! thanks

PR #1810, opened 2023-07-07
refactor: fixup fvm IPLD flush logic

Stebalien:
In preparation for reachability analysis (we're going to re-use this same code).

- Make it less generic (performance).
- Remove blocks from the write buffer as we write them to avoid duplicate writes.
- Simplify some of the checks around what is allowed. For example, I'm now allowing CBOR + Identity hash which should have been allowed previously but wasn't (we don't use it but still, it should have been allowed).
- Remove the explicit 100 byte CID length check. The `Cid` type already validates that the digest can be no longer than 64 bytes.
- Be less strict on DagCBOR validation. Counter-intuitively, being overly strict here is dangerous as it gives us more points where implementations can disagree and fork. Instead, we enforce the following rules for DAG_CBOR:
  1. Blocks must have a valid CBOR structure, but _values_ aren't validated. E.g., no utf-8 validation, no float validation, no minimum encoding requirements, no canonical ordering requirements, etc.
  2. All CBOR values tagged with 42 must be valid CIDs. I.e., a CBOR byte string starting with a 0x0 byte followed by a valid CID with at most a 64 byte digest.
- Remove recursion. We're likely going to implement a max IPLD graph depth anyways, but we might as well be safe. In terms of memory usage, this logic will allocate `O(blocks)` for blocks already in memory, so there should be no danger of running out of memory.

Working towards https://github.com/filecoin-project/fvm-pm/issues/715

### DIFF
--- a/fvm/src/blockstore/buffered.rs
+++ b/fvm/src/blockstore/buffered.rs
@@ -4,13 +4,12 @@
 
 use std::cell::RefCell;
 use std::collections::HashMap;
-use std::io::{Cursor, Read, Seek};
+use std::io::Read;
 
 use anyhow::{anyhow, Result};
-use byteorder::{BigEndian, ByteOrder, ReadBytesExt};
 use cid::Cid;
 use fvm_ipld_blockstore::{Blockstore, Buffered};
-use fvm_ipld_encoding::{CBOR, DAG_CBOR};
+use fvm_ipld_encoding::{CBOR, DAG_CBOR, IPLD_RAW};
 use fvm_shared::commcid::{FIL_COMMITMENT_SEALED, FIL_COMMITMENT_UNSEALED};
 
 /// Wrapper around `Blockstore` to limit and have control over when values are written.
@@ -43,15 +42,10 @@ where
 {
     /// Flushes the buffered cache based on the root node.
     /// This will recursively traverse the cache and write all data connected by links to this
-    /// root Cid. Calling flush will not reset the write buffer.
+    /// root Cid, moving the reachable blocks from the write buffer to the backing store.
     fn flush(&self, root: &Cid) -> Result<()> {
-        let mut buffer = Vec::new();
-        let s = self.write.borrow();
-        copy_rec(&s, *root, &mut buffer)?;
-
-        self.base.put_many_keyed(buffer)?;
-
-        Ok(())
+        self.base
+            .put_many_keyed(take_reachable(&mut self.write.borrow_mut(), root)?)
     }
 }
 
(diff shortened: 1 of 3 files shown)


fvm/src/blockstore/buffered.rs
```diff
@@ -62,86 +56,63 @@ where
...
             6 => {
                 // Check if the tag refers to a CID
                 if extra == 42 {
-                    let (maj, extra) = cbor_read_header_buf(buf, &mut scratch)?;
+                    let (maj, extra) = cbor_read_header_buf(&mut buf)?;
                     // The actual CID is expected to be a byte string
                     if maj != 2 {
                         return Err(anyhow!("expected cbor type byte string in input"));
                     }
-                    if extra > 100 {
```
Stebalien, line 139: This was unnecessary. If the CID is invalid and/or too big, `Cid::try_from` will catch that.

fvm/src/blockstore/buffered.rs
```diff
@@ -180,53 +147,53 @@ fn copy_rec<'a>(
...
+        match (k.hash().code(), k.hash().size()) {
+            // Allow non-truncated blake2b-256 and identity hashes.
+            (BLAKE2B_256, BLAKE2B_LEN) | (IDENTITY, _) => (),
+            // Reject everything else.
+            (hash, length) => {
+                return Err(anyhow!(
+                    "cid {k} has unexpected multihash (code={hash}, len={length})"
+                ))
+            }
         }
```
Stebalien, line 173: The previous logic was shorter but hard to reason about.

fvm/src/blockstore/buffered.rs
```diff
@@ -43,15 +42,10 @@ where
...
     fn flush(&self, root: &Cid) -> Result<()> {
-        let mut buffer = Vec::new();
-        let s = self.write.borrow();
-        copy_rec(&s, *root, &mut buffer)?;
-
-        self.base.put_many_keyed(buffer)?;
-
-        Ok(())
+        self.base
+            .put_many_keyed(take_reachable(&mut self.write.borrow_mut(), root)?)
```
Stebalien, line 48: I changed this to "take" blocks so we wouldn't end up writing them multiple times. The alternative would have been to keep some form of "written" set.

For some history, we used to clear the _entire_ buffer on flush. However, _that_ ended up causing some issues if we, e.g., wanted to flush some intermediate state. Now we take a middle-path of deleting the written blocks but leaving everything else.

fvm/src/blockstore/buffered.rs
```diff
@@ -62,86 +56,63 @@ where
...
-        Err(anyhow!("invalid header cbor_read_header_buf"))
-    }
+    let val = match low {
+        ..=23 => low.into(),
+        24 => read_fixed::<1>(br)?[0].into(),
+        25 => u16::from_be_bytes(read_fixed(br)?).into(),
+        26 => u32::from_be_bytes(read_fixed(br)?).into(),
+        27 => u64::from_be_bytes(read_fixed(br)?),
+        _ => return Err(anyhow!("invalid header cbor_read_header_buf")),
+    };
```
fridrik01, line 77: Great refactor for improving readability! 

Regarding performance, IIUC this is now doing 2 extra allocations for each call now instead of using a scratch buffer as before?

fvm/src/blockstore/buffered.rs
fridrik01: Will this then end up saving disk space?

fridrik01 (approved):
LGTM

fvm/src/blockstore/buffered.rs
Stebalien: Stack allocations so it should basically be free. Technically, it might cause some zeroing, but that shouldn't matter (and may not even happen if rust can optimize it).

PR #1812, opened 2023-07-13
fix(fvm/syscall): don't assume that return pointers are aligned

Stebalien:
This isn't an issue in the builtin actors as deployed today, but would definitely be an issue with custom actors that may not pass aligned "out" pointers. We fix this by using the special "write unaligned" method to avoid ever dereferencing the destination pointer.


Stebalien:
An example of how this can go wrong:

```rust
fn main() {
    unsafe {
        struct Foo(u64);
        let mut x = [0u8; 9];
        let ptr = x.as_mut_ptr().offset(0);
        assert!(
            (ptr as usize) % std::mem::align_of::<Foo>() > 0,
            "should not be aligned to {}",
            std::mem::align_of::<Foo>(),
        );
        *(ptr as *mut Foo) = Foo(0xffff);
        println!("Results: {x:?}")
    }
}
```

Stebalien:
Er, I'm going to add a quick integration test before merging.

Stebalien:
Ok, so... the test-case is "ok", but... due to our use of repr-packed, this isn't even an issue anyways (technically). Because rust treats repr-packed structs as having 1-alignment anyways.

But we should still do this to be safe (might as well).

fridrik01 (approved):

PR #1820, opened 2023-07-20
steb's review changes

Stebalien:
(we're going to squash anyways so there's little point in breaking this into multiple commits)

- Pass event slice length instead of slice of byte buffer.
- Rework gas costs for the new call and remove the indexing costs because we don't actually charge anything anyways. This lets us charge once at the top.
- Add an additional "deferred" memory allocation charge per-event which likely should have been in the initial spec (will need to go into the FIP).
- Be extra careful about overflows. Unecessary, but we do this _just_ in case.
- _Do_ cast the event entry slice but assert that the alignment is the same as u8 (bytes). Given `repr(packed)`, this is guaranteed. - Don't cap the total value length at 8kib. The total cap is 8kib * 255.
- Reduce the max key length & max number of entries by one each so we can better reason about estimated CBOR sizes. I.e., 255 fits in 2 bytes (major type plus one byte for the integer), 256 does not. - Isolate the unsafe logic into the syscall API.

And some other misc cleanups. I did this in a PR instead of just reviewing because, well, it required quite a bit of hands-on "does this work".


fvm/src/gas/price_list.rs
```diff
@@ -27,6 +26,21 @@ use crate::kernel::SupportedHashes;
...
+//   2: Entry array overhead (max size 255)
+const EVENT_OVERHEAD: u64 = 12;
+// The maximum overhead (in bytes) of a single event entry when encoded into CBOR.
+//
+// 1: CBOR tuple with 4 fields
+//   1: Flags (will adjust as more flags are added)
+//   2: Key major type + length (up to 255 bytes)
+//   2: Codec major type + value (codec should be <= 255)
+//   3: Value major type + length (up to 8192 bytes)
+const EVENT_ENTRY_OVERHEAD: u64 = 9;
```
Stebalien, line 42: In an ideal world, we'd just serialize then charge the exact amounts. But we don't form the AMT containing events till everything is done.

In theory, we could defer the charge till the end but... that would break per-call gas limits.

fvm/src/gas/price_list.rs
```diff
@@ -468,14 +471,14 @@ pub struct PriceList {
 
     /// Gas cost to validate an ActorEvent as soon as it's received from the actor, and prior
     /// to it being parsed.
-    pub(crate) event_validation_cost: ScalingCost,
-
-    /// Gas cost of every indexed element, scaling per number of bytes indexed.
-    pub(crate) event_accept_per_index_element: ScalingCost,
```
Stebalien, line 474: Removed because it was unused.

fvm/src/gas/price_list.rs
```diff
@@ -468,14 +471,14 @@ pub struct PriceList {
...
+    pub(crate) event_per_entry: ScalingCost,
 
     /// Gas cost of doing lookups in the builtin actor mappings.
     pub(crate) builtin_actor_manifest_lookup: Gas,
 
+    /// Gas cost of utf8 parsing.
+    pub(crate) utf8_validation: ScalingCost,
```
Stebalien, line 480: Split out as we should benchmark this separately.

fvm/src/gas/price_list.rs
```diff
@@ -919,58 +922,33 @@ impl PriceList {
...
-            if evt.flags.contains(Flags::FLAG_INDEXED_KEY) {
-                indexed_bytes += evt.key.len();
-                indexed_elements += 1;
-            }
-            if evt.flags.contains(Flags::FLAG_INDEXED_VALUE) {
-                indexed_bytes += evt.value.len();
-                indexed_elements += 1;
-            }
-        }
+    pub fn on_actor_event(&self, entries: usize, keysize: usize, valuesize: usize) -> GasCharge {
```
Stebalien, line 925: Renamed as we now charge for everything all at once. _Later_, we may charge for indexing after validation. But we can cross that bridge later...

fvm/src/gas/price_list.rs
```diff
@@ -919,58 +922,33 @@ impl PriceList {
...
-                + self.event_accept_per_index_element.scale * indexed_bytes
-                + memcpy * 2u32 // deferred cost, placing here to hide from benchmark
-                + alloc // deferred cost, placing here to hide from benchmark
-                + hash, // deferred cost, placing here to hide from benchmark
+            "OnActorEvent",
+            // Charge for validation/storing events.
+            mem + validate_entries + validate_utf8,
+            // Charge for forming the AMT and returning the events to the client.
+            // one copy into the AMT, one copy to the client.
+            hash + (mem * 2u32),
```
Stebalien, line 951: We used to charge for only one allocation here but I'm now charging for two (which we probably should have done since the start).

fvm/src/kernel/default.rs
```diff
@@ -1006,113 +1006,93 @@ impl<C> EventOps for DefaultKernel<C>
 where
     C: CallManager,
 {
-    fn emit_event(&mut self, raw_evt: &[u8], raw_key: &[u8], raw_val: &[u8]) -> Result<()> {
-        const MAX_NR_ENTRIES: usize = 256;
-        const MAX_KEY_LEN: usize = 32;
+    fn emit_event(
+        &mut self,
+        event_headers: &[fvm_shared::sys::EventEntry],
```
Stebalien, line 1011: Taking this as a slice of event entries to avoid unsafe logic in the kernel where possible.

fvm/src/kernel/default.rs
```diff
@@ -1006,113 +1006,93 @@ impl<C> EventOps for DefaultKernel<C>
...
+    ) -> Result<()> {
+        const MAX_NR_ENTRIES: usize = 255;
+        const MAX_KEY_LEN: usize = 31;
         const MAX_VALUE_LEN: usize = 8 << 10;
 
         if self.read_only {
             return Err(syscall_error!(ReadOnly; "cannot emit events while read-only").into());
         }
 
-        if raw_val.len() > MAX_VALUE_LEN {
```
Stebalien, line 1018: Removed this check as we allow _each_ value to be 8kib.

fvm/src/kernel/default.rs
```diff
@@ -1006,113 +1006,93 @@ impl<C> EventOps for DefaultKernel<C>
...
+            let flags = header.flags;
             if Flags::from_bits(flags.bits()).is_none() {
                 return Err(
                     syscall_error!(IllegalArgument; "event flags are invalid: {}", flags.bits())
                         .into(),
                 );
             }
-            if entry_fixed.key_len > MAX_KEY_LEN as u32 {
-                let tmp = entry_fixed.key_len;
+            if header.key_len > MAX_KEY_LEN as u32 {
```
Stebalien, line 1048: simple rename to make it clear what it is.

fvm/src/kernel/default.rs
```diff
@@ -1006,113 +1006,93 @@ impl<C> EventOps for DefaultKernel<C>
...
-                &raw_key[key_offset..key_offset + entry_fixed.key_len as usize],
-            ) {
-                Ok(s) => s,
-                Err(e) => {
-                    return Err(syscall_error!(IllegalArgument, "invalid event key: {}", e).into())
-                }
-            };
-            let value = &raw_val[val_offset..val_offset + entry_fixed.val_len as usize];
+            let key = &event_keys
+                .get(key_offset..key_offset + header.key_len as usize)
```
Stebalien, line 1066: bounds checking.

No need to check for overflows as we're limiting the sizes of the key/value lengths.

shared/src/sys/mod.rs
```diff
@@ -67,6 +67,16 @@ impl SendFlags {
     }
 }
 
+/// A fixed sized struct for serializing an [event `Entry`](crate::event::Entry) separately from the
+/// key/value bytes.
+#[repr(C, packed)]
+pub struct EventEntry {
```
Stebalien, line 73: Moved as this is where we generally stash all of our shared "syscall" structs.

Stebalien:
Failing tests are expected for now.

fvm/src/syscalls/event.rs
```diff
@@ -32,8 +33,21 @@ pub fn emit_event(
...
+        static_assertions::assert_eq_align!(fvm_shared::sys::EventEntry, u8);
+        let size = event_len
+            .checked_mul(EVENT_SIZE)
+            .context("events index out of bounds")
+            .or_illegal_argument()?;
+        let buf = context.memory.try_slice(event_off, size)?;
+        std::slice::from_raw_parts(
+            buf.as_ptr() as *const fvm_shared::sys::EventEntry,
+            event_len as usize,
+        )
```
fridrik01, line 48: Yeah, it makes sense doing as much condition checking here rather in the kernel.

fvm/src/kernel/default.rs
fridrik01: oh, that is interesting, it has up to now always been 8kib in total over _all_ values. Should we remove this check ([relevant code](https://github.com/filecoin-project/ref-fvm/blob/master/fvm/src/kernel/default.rs#L1086-L1090)) in master as well?

fvm/src/kernel/default.rs
```diff
@@ -1006,113 +1006,93 @@ impl<C> EventOps for DefaultKernel<C>
...
-        const MAX_NR_ENTRIES: usize = 256;
-        const MAX_KEY_LEN: usize = 32;
+    fn emit_event(
+        &mut self,
+        event_headers: &[fvm_shared::sys::EventEntry],
+        event_keys: &[u8],
+        event_values: &[u8],
+    ) -> Result<()> {
+        const MAX_NR_ENTRIES: usize = 255;
+        const MAX_KEY_LEN: usize = 31;
```
fridrik01, line 1016: was going to ask about this, but then read the PR summary again, makes sense!

fridrik01 (approved):
Looks great!

fvm/src/kernel/default.rs
Stebalien: Uh. No, no, you were right. I'm not sure what I was looking at (and I wrote the code + spec so I'm not sure what I was thinking either...).

Stebalien: Fixed and I've:

1. Renamed the constant so I don't mess this up again.
2. Added a comment to the second check to explain why we have two.

Issue #1793, opened 2023-06-26
Trace requests (for Internal Transaction support in Blockscout)


eshon:
To enable Internal Transactions in the new FEVM block explorer Blockscout, the following trace API is needed:

https://docs.blockscout.com/for-developers/information-and-settings/node-tracing-json-rpc-requirements#enable-tracing-to-fetch-internal-transactions

eshon:
To enable Internal Transactions in the new FEVM block explorer Blockscout, the following trace API is needed:

https://docs.blockscout.com/for-developers/information-and-settings/node-tracing-json-rpc-requirements#enable-tracing-to-fetch-internal-transactions

eshon:
Dune Analytics and ngram also need this for integration.

eshon:
Update from Dune on the requests they need:

Head - eth_blockNumber
Block - eth_getBlockByNumber
Traces - debug_traceTransaction or trace_block (Needed for decoding)
Receipts - eth_getTransactionReceipt (one call per transaction)
Health - net_version

cc: @fridrik01

Stebalien:
Status:

- We landed an implementation of `trace_block`, but need another pass to resolve some TODOs.
- `eth_getBlockByNumber` and `eth_blockNumber` are already implemented. However, the former will still stumble a bit on "native" messages. `trace_block` has some very special logic for handling native transactions, but we haven't ported those over to `eth_getBlockByNumber` yet.
- `ethGetTransactionReceipt` exists.
- `net_version` exists.

At the moment, @fridrik01 is working on other things. We'll get back to this, but we all needed a break as it turned out that mapping Filecoin semantics onto Ethereum was... very painful.

Stebalien:
Also, I thought we _weren't_ doing Dune because they wanted 500k. Did that change again?

eshon:
Ah sorry - @Stebalien Can you plz delete your comment? Price negotiations like that are normally confidential / internal to an organization.

PR #1823, opened 2023-07-27
feat: add new fields to message trace for supporting new trace api

fridrik01:
Related: https://github.com/filecoin-project/ref-fvm/issues/1793 and https://github.com/filecoin-project/fvm-pm/issues/613

This PR adds the following fields to `ExecutionEvent` which are required to confirm to the trace api [spec](https://docs.alchemy.com/reference/trace-api):
- `gas_limit`: The gas provided by the sender
- `read_only`: Whether the call is read only or not

### DIFF
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -186,6 +186,13 @@ where
                 method,
                 params: params.as_ref().map(Into::into),
                 value: value.clone(),
+                gas_limit: std::cmp::min(
+                    gas_limit
+                        .unwrap_or(Gas::from_milligas(u64::MAX))
+                        .as_milligas(),
+                    self.gas_tracker.gas_available().as_milligas(),
+                ),
+                read_only,
             });
         }
 
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -187,10 +187,8 @@ where
                 params: params.as_ref().map(Into::into),
                 value: value.clone(),
                 gas_limit: std::cmp::min(
-                    gas_limit
-                        .unwrap_or(Gas::from_milligas(u64::MAX))
-                        .as_milligas(),
-                    self.gas_tracker.gas_available().as_milligas(),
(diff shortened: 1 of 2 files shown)


fvm/src/trace/mod.rs
```diff
@@ -25,7 +26,10 @@ pub enum ExecutionEvent {
         method: MethodNum,
         params: Option<IpldBlock>,
         value: TokenAmount,
+        gas_limit: u64,
+        read_only: bool,
     },
     CallReturn(ExitCode, Option<IpldBlock>),
     CallError(SyscallError),
+    InvokeActor(Cid),
```
anorth, line 34: What's `InvokeActor`, and how does it differ from `Call`?

anorth (commented):
FYI @alexytsu

fvm/src/trace/mod.rs
fridrik01: Its the code CID of the invoked actor after resolving its address.

Stebalien: Basically:

1. Call means actor A calls actor B, but actor B may not yet exist (and/or may never exist).
2. InvokeActor means that we actually successfully invoked actor B.

anorth: So we'll see both in the trace for most successful calls? And there's no InvokeActorEnd? 

This all sounds fine, but please consider documenting it here.

arajasek (approved):

Stebalien (approved):

fvm/src/call_manager/default.rs
```diff
@@ -186,6 +186,13 @@ where
                 method,
                 params: params.as_ref().map(Into::into),
                 value: value.clone(),
+                gas_limit: std::cmp::min(
```
Stebalien, line 189: Ah, actually, this should just be "gas". We only use milligas internally. Just round-up.

Stebalien (changes requested):
One change, but then feel free to dismiss my review and merge it.

Stebalien (approved):

PR #1829, opened 2023-08-01
refactor: remove `with_transaction` and move "return" gas charge

Stebalien:
(ipld reachability pre-factor)

In #1824, we'll need to charge for tracking reachable links on return. However, we'd _like_ to apply the gas limit specified by the caller so the callee can't "go over" their gas limit. That means we need a return-value charge inside the call manager instead of inside the caller (executor/kernel).

So, I figured I'd just apply a refactor that's been a long-time coming and remove the `with_transaction` dance. Instead, I put _all_ of the transaction logic into `CallManager::send`. This makes the function a bit heftier but, IMO, it's easier to reason about this way. It also removes an implementation detail (`with_transaction`) from the `CallManager` trait.

### DIFF
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -194,14 +194,46 @@ where
             self.gas_tracker.push_limit(limit);
         }
 
-        let mut result = self.with_stack_frame(|s| {
-            s.send_unchecked::<K>(from, to, method, params, value, read_only)
-        });
-
-        // If we pushed a limit, pop it.
-        if gas_limit.is_some() {
-            self.gas_tracker.pop_limit()?;
+        if self.call_stack_depth >= self.machine.context().max_call_depth {
+            let sys_err = syscall_error!(LimitExceeded, "message execution exceeds call depth");
+            if self.machine.context().tracing {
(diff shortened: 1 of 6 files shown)


fvm/src/call_manager/default.rs
```diff
@@ -194,14 +194,46 @@ where
...
+            Ok(v) => (!v.exit_code.is_success(), Ok(v)),
+            Err(e) => (true, Err(e)),
+        };
+
+        self.call_stack_depth -= 1;
+        // Return the _first_ error (if any). We don't expect any errors here anyways as all error
+        // cases are fatal.
+        if let Some(err) = [
+            // End all transactions
+            self.state_access_tracker.end_transaction(revert).err(),
```
Stebalien, line 224: The previous logic with early returns was technically safe, but always made me pause and bite my nails whenever I looked at it. This version should be a bit more robust.

fridrik01 (approved):
LGTM

fvm/src/call_manager/default.rs
fridrik01: Oh this took me a few seconds to read correctly but then I noticed the square bracket so it started to make sense. But yeah, these early returns looked to me like a bug honestly when I read it.

Issue #1828, opened 2023-08-01
Deprecate Hamt::load


Stebalien:
See https://github.com/filecoin-project/builtin-actors/issues/1346.

This method is a lie as it'll just automatically use a bit-width of 8 without checking anything.

PR #1831, opened 2023-08-02
chore: deprecate default bitwidths in the KAMT/AMT

### LINKED ISSUE #1346: Gas: Reprice state read/write
We need to account for the cost of looking actors up in the state-tree, and writing them back.

Stebalien:
1. We don't intentionally use these anywhere.
2. We _do_ unintentionally use these where we shouldn't (https://github.com/filecoin-project/builtin-actors/issues/1346).

fixes https://github.com/filecoin-project/ref-fvm/issues/1828

### DIFF
--- a/ipld/kamt/benches/kamt_benchmark.rs
+++ b/ipld/kamt/benches/kamt_benchmark.rs
@@ -9,9 +9,15 @@ use std::borrow::Cow;
 use criterion::{black_box, criterion_group, criterion_main, Criterion};
 use fvm_ipld_blockstore::MemoryBlockstore;
 use fvm_ipld_encoding::tuple::*;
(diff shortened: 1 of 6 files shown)


ipld/kamt/benches/kamt_benchmark.rs
```diff
@@ -56,7 +62,7 @@ fn insert(c: &mut Criterion) {
     c.bench_function("KAMT bulk insert (no flush)", |b| {
         b.iter(|| {
             let db = fvm_ipld_blockstore::MemoryBlockstore::default();
-            let mut a = BKamt::new(&db);
+            let mut a = BKamt::new_with_config(&db, Config::default());
```
anorth, line 65: TEST_CONFIG here too?

anorth (approved):

Issue #286, opened 2022-01-20
Use external iteration for hamt


Stebalien:
These currently use internal iteration but this makes them difficult to use with iterator adapters.

Unfortunately, HAMT  iteration is recursive so we'll need some form of stack. However, that stack usually has a fixed depth that we can allocate up-front.

Stebalien:
Opening a new issue for KAMT and AMT.

PR #1833, opened 2023-08-04
feat(hamt): implement external iteration

Stebalien:
Previously, we only supported "internal" iteration with a callback. This switches to an external iterator which:

1. Allows us to make the APIs a little more composable.
2. Allows HAMT iterators to be used in rust for loops.
3. Avoids the awkward error wrapping when bubbling up errors from `for_each` calls.

Usage:

```rust
let mut hamt = Hamt::new_with_bit_width(&bs, 4);

// ...

for item in &hamt {
    let (k, v) = item?; // error handling is a bit annoying.
}

// ranged iteration
for item in hamt.iter_from(start_key).take(50) {
    // ...
}

```

part of https://github.com/filecoin-project/ref-fvm/issues/286

### DIFF
--- a/ipld/hamt/src/iter.rs
+++ b/ipld/hamt/src/iter.rs
@@ -3,7 +3,6 @@
 use std::borrow::Borrow;
 use std::iter::FusedIterator;
 
-use anyhow::anyhow;
 use forest_hash_utils::BytesKey;
 use fvm_ipld_blockstore::Blockstore;
 use fvm_ipld_encoding::de::DeserializeOwned;
@@ -91,7 +90,7 @@ where
                                 stack,
                                 current: values[offset..].iter(),
                             }),
-                            None => Err(anyhow!("key not found").into()),
+                            None => Err(Error::StartKeyNotFound),
                         }
                     }
                 },
--- a/ipld/hamt/src/iter.rs
+++ b/ipld/hamt/src/iter.rs
@@ -113,36 +113,36 @@ where
             return Some(Ok((v.key(), v.value())));
         }
         loop {
-            if let Some(next) = self.stack.last_mut()?.next() {
-                match next {
-                    Pointer::Link { cid, cache } => {
-                        let node = if let Some(cached_node) = cache.get() {
-                            cached_node
-                        } else {
-                            let node = match self.store.get_cbor::<Node<K, V, H, Ver>>(cid) {
-                                Ok(Some(node)) => node,
-                                #[cfg(not(feature = "ignore-dead-links"))]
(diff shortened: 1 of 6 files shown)


Stebalien:
@anorth would you be interested in something like this? This basically means you wouldn't need to call `for_each` and plumb through the error. Additionally, you can stop iteration at any time without any special stop error.

Stebalien:
(draft because it likely needs some more tests and docs)

Stebalien:
Benchmarks: no perf change.

anorth:
Yeah, something like this would be great.

Stebalien:
Ok, docs up with a doctest. Functional tests should already be covered by the existing `for_each` and `for_each_ranged` tests (which delegate to the new iterators).

Stebalien:
Note: there's one _tiny_ breaking change: I've extended the error enum. So I guess I'll cut a minor release just in case. Although it shouldn't actually affect anyone.

anorth:
I'll review, but I don't feel confident enough in my knowledge of HAMT internals to be any great auditor for the iterator implementation. I've requested review from @Kubuxu too.

anorth (approved):
I'm a little concerned that existing tests only test one internal shape of HAMT: 200 integers. I'd be a little happier if they extended to test, say, every HAMT size from 0 to 200, and some different bitwidths. These are all degress though, short of hacking the hash function to enable explicit control over branching and shape.

Stebalien:
I'll see if I can make it a bit more parametric.

(latest push was just a small cleanup to reduce indentation a bit)

Stebalien:
@creativecoder does this PR make sense?

Stebalien:
(not sure if edits summon..., so I'll try again @creativcoder)

Stebalien:
@filecoin-project/forest anyone feel comfortable reviewing this code?

Stebalien:
Np. Nobody is waiting on this, I just want to get it off the table so to speak.

ipld/hamt/src/iter.rs
```diff
@@ -0,0 +1,158 @@
...
+        K: Borrow<Q>,
+        Q: Hash + Eq,
+    {
+        let hashed_key = H::hash(key);
+        let mut hash = HashBits::new(&hashed_key);
+        let mut node = root;
+        let mut stack = Vec::new();
+        loop {
+            let idx = hash.next(conf.bit_width)?;
+            stack.push(node.pointers[node.index_for_bit_pos(idx)..].iter());
```
creativcoder, line 62: is it possible for the indexing to panic?

creativcoder (approved):
This looks cleaner than for_each, thanks!

Stebalien:
Thanks for the review!

PR #1836, opened 2023-08-08
feat: add syscall for bls aggregate signature verification

### LINKED ISSUE #1349: Replace verify_signature with verify_bls_aggregate
We currently have:

- `verify_signature` for verifying arbitrary signatures.
- `recover_secp_public_key` for recovering a secp public key from a secp signature.

Unfortunately, the first one is a bit limited:

- It does hashing internally, which restricts the user to a specific hash function.
- It needs the public key to do verification in the first place, which is why we needed `recover_secp_public_key`.
- It can't, e.g., validate bls signature aggregates.

Proposal:

1. Add a new `verify_bls_aggregate` (where the aggregate may have one element).
2. Remove `verify_signature`.
3. Move all hashing to userspace.

DrPeterVanNostrand:
This PR adds a syscall and associated kernel operation for BLS aggregate signature validation.

Work remaining for this draft PR:
- [x] Remove `verify_signature` syscall
- [ ] Calculate gas charge for bls aggregate signature validation (maybe remove `on_verify_signature` from `PriceList`)

Closes #1349

### DIFF
--- a/fvm/src/gas/price_list.rs
+++ b/fvm/src/gas/price_list.rs
@@ -583,7 +583,7 @@ impl PriceList {
     }
 
     /// Returns gas required for BLS aggregate signature verification.
-    pub fn on_verify_aggregate_signature(&self, num_sigs: usize) -> GasCharge {
+    pub fn on_verify_aggregate_signature(&self, num_sigs: usize, data_len: usize) -> GasCharge {
         // TODO: the gas cost calculated below is only an estimate; its actual value is yet to be
         // benchmarked.
 
@@ -594,11 +594,21 @@ impl PriceList {
         // Note that `bls_signatures` rearranges the textbook verifier equation (containing
         // `num_sigs + 1` full pairings) into a more efficient equation containing `num_sigs + 1`
         // Miller loops and one final exponentiation.
-        let num_pairings = num_sigs as u64 + 1;
-        let gas_two_pairings = self.sig_cost[&SignatureType::BLS].flat.as_milligas();
-        let gas_one_pairing = gas_two_pairings / 2;
-        let gas = Gas::from_milligas(num_pairings * gas_one_pairing);
-        GasCharge::new("OnVerifySignature", gas, Zero::zero())
+        let (num_pairings, data_len) = (num_sigs as u64 + 1, data_len as u64);
+
+        let ScalingCost {
+            flat: gas_two_pairings,
+            scale: gas_per_byte,
+        } = self.sig_cost[&SignatureType::BLS];
+        let gas_one_pairing = gas_two_pairings.as_milligas() / 2;
+        let gas_pairings = num_pairings * gas_one_pairing;
+        let gas_hashing = data_len * gas_per_byte.as_milligas();
+
+        GasCharge::new(
+            "OnVerifySignature",
+            Gas::from_milligas(gas_pairings + gas_hashing),
+            Zero::zero(),
+        )
     }
 
     /// Returns gas required for recovering signer pubkey from signature
--- a/fvm/src/gas/price_list.rs
+++ b/fvm/src/gas/price_list.rs
@@ -582,6 +582,22 @@ impl PriceList {
         GasCharge::new("OnVerifySignature", gas, Zero::zero())
     }
 
+    /// Returns gas required for BLS aggregate signature verification.
+    pub fn on_verify_aggregate_signature(&self, num_sigs: usize) -> GasCharge {
+        // When `num_sigs` BLS signatures are aggregated into a single signature, the aggregate
+        // signature verifier must perform `num_sigs + 1` expensive pairing operations (one
+        // pairing on the aggregate signature, and one pairing for each signed plaintext's digest).
+        //
+        // Note that `bls_signatures` rearranges the textbook verifier equation (containing
+        // `num_sigs + 1` full pairings) into a more efficient equation containing `num_sigs + 1`
+        // Miller loops and one final exponentiation.
+        let num_pairings = num_sigs as u64 + 1;
+        let gas_two_pairings = self.sig_cost[&SignatureType::BLS].flat.as_milligas();
+        let gas_one_pairing = gas_two_pairings / 2;
+        let gas = Gas::from_milligas(num_pairings * gas_one_pairing);
+        GasCharge::new("OnVerifySignature", gas, Zero::zero())
+    }
+
     /// Returns gas required for recovering signer pubkey from signature
     #[inline]
     pub fn on_recover_secp_public_key(&self) -> GasCharge {
--- a/fvm/src/gas/price_list.rs
+++ b/fvm/src/gas/price_list.rs
@@ -584,6 +584,9 @@ impl PriceList {
 
     /// Returns gas required for BLS aggregate signature verification.
     pub fn on_verify_aggregate_signature(&self, num_sigs: usize) -> GasCharge {
+        // TODO: the gas cost calculated below is only an estimate; its actual value is yet to be
+        // benchmarked.
+
         // When `num_sigs` BLS signatures are aggregated into a single signature, the aggregate
         // signature verifier must perform `num_sigs + 1` expensive pairing operations (one
         // pairing on the aggregate signature, and one pairing for each signed plaintext's digest).
--- a/fvm/src/gas/price_list.rs
+++ b/fvm/src/gas/price_list.rs
@@ -7,7 +7,6 @@ use std::ops::Mul;
 
 use anyhow::Context;
 use fvm_shared::crypto::signature::SignatureType;
-use fvm_shared::event::{ActorEvent, Flags};
 use fvm_shared::piece::PieceInfo;
 use fvm_shared::sector::{
     AggregateSealVerifyProofAndInfos, RegisteredPoStProof, RegisteredSealProof, ReplicaUpdateInfo,
@@ -26,6 +25,21 @@ use crate::kernel::SupportedHashes;
 // Each element reserves a `usize` in the table, so we charge 8 bytes per pointer.
 // https://docs.rs/wasmtime/2.0.2/wasmtime/struct.InstanceLimits.html#structfield.table_elements
 const TABLE_ELEMENT_SIZE: u32 = 8;
+
+// The maximum overhead (in bytes) of a single event when encoded into CBOR.
+//
+// 1: CBOR tuple with 2 fields (StampedEvent)
+//   9: Emitter ID
+//   2: Entry array overhead (max size 255)
+const EVENT_OVERHEAD: u64 = 12;
+// The maximum overhead (in bytes) of a single event entry when encoded into CBOR.
+//
+// 1: CBOR tuple with 4 fields
+//   1: Flags (will adjust as more flags are added)
+//   2: Key major type + length (up to 255 bytes)
+//   2: Codec major type + value (codec should be <= 255)
+//   3: Value major type + length (up to 8192 bytes)
+const EVENT_ENTRY_OVERHEAD: u64 = 9;
 
 /// Create a mapping from enum items to values in a way that guarantees at compile
 /// time that we did not miss any member, in any of the prices, even if the enum
@@ -293,26 +307,15 @@ lazy_static! {
             memory_fill_per_byte_cost: Gas::from_milligas(400),
         },
 
-        // These parameters are specifically sized for EVM events. They will need
-        // to be revisited before Wasm actors are able to emit arbitrary events.
-        //
-        // Validation costs are dependent on encoded length, but also
-        // co-dependent on the number of entries. The latter is a chicken-and-egg
-        // situation because we can only learn how many entries were emitted once we
-        // decode the CBOR event.
-        //
-        // We will likely need to revisit the ABI of emit_event to remove CBOR
-        // as the exchange format.
-        event_validation_cost: ScalingCost {
+        // TODO(#1817): Per-entry event validation cost. These parameters were benchmarked for the
+        // EVM but haven't been revisited since revising the API.
+        event_per_entry: ScalingCost {
             flat: Gas::new(1750),
             scale: Gas::new(25),
         },
 
-        // The protocol does not currently mandate indexing work, so these are
-        // left at zero. Once we start populating and committing indexing data
-        // structures, these costs will need to reflect the computation and
-        // storage costs of such actions.
-        event_accept_per_index_element: ScalingCost {
+        // TODO(#1817): Cost of validating utf8 (used in event parsing).
+        utf8_validation: ScalingCost {
             flat: Zero::zero(),
             scale: Zero::zero(),
         },
@@ -468,13 +471,13 @@ pub struct PriceList {
 
     /// Gas cost to validate an ActorEvent as soon as it's received from the actor, and prior
     /// to it being parsed.
-    pub(crate) event_validation_cost: ScalingCost,
-
-    /// Gas cost of every indexed element, scaling per number of bytes indexed.
-    pub(crate) event_accept_per_index_element: ScalingCost,
+    pub(crate) event_per_entry: ScalingCost,
 
     /// Gas cost of doing lookups in the builtin actor mappings.
     pub(crate) builtin_actor_manifest_lookup: Gas,
+
+    /// Gas cost of utf8 parsing.
+    pub(crate) utf8_validation: ScalingCost,
 
     /// Gas cost of accessing the network context.
     pub(crate) network_context: Gas,
@@ -921,59 +924,33 @@ impl PriceList {
     }
 
     #[inline]
-    pub fn on_actor_event_validate(&self, data_size: usize) -> GasCharge {
-        let memcpy = self.block_memcpy.apply(data_size);
-        let alloc = self.block_allocate.apply(data_size);
-        let validate = self.event_validation_cost.apply(data_size);
+    pub fn on_actor_event(&self, entries: usize, keysize: usize, valuesize: usize) -> GasCharge {
+        // Here we estimate per-event overhead given the constraints on event values.
 
-        GasCharge::new(
-            "OnActorEventValidate",
-            memcpy + alloc + validate,
-            Zero::zero(),
-        )
-    }
+        let validate_entries = self.event_per_entry.apply(entries);
+        let validate_utf8 = self.utf8_validation.apply(keysize);
 
-    #[inline]
-    pub fn on_actor_event_accept(&self, evt: &ActorEvent, serialized_len: usize) -> GasCharge {
-        let (mut indexed_bytes, mut indexed_elements) = (0usize, 0u32);
-        for evt in evt.entries.iter() {
-            if evt.flags.contains(Flags::FLAG_INDEXED_KEY) {
-                indexed_bytes += evt.key.len();
-                indexed_elements += 1;
-            }
-            if evt.flags.contains(Flags::FLAG_INDEXED_VALUE) {
-                indexed_bytes += evt.value.len();
-                indexed_elements += 1;
-            }
-        }
+        // Estimate the size, saturating at max-u64. Given how we calculate gas, this will saturate
+        // the gas maximum at max-u64 milligas.
+        let estimated_size = EVENT_OVERHEAD
+            .saturating_add(EVENT_ENTRY_OVERHEAD.saturating_mul(entries as u64))
+            .saturating_add(keysize as u64)
+            .saturating_add(valuesize as u64);
 
-        // The estimated size of the serialized StampedEvent event, which
-        // includes the ActorEvent + 8 bytes for the actor ID + some bytes
-        // for CBOR framing.
-        const STAMP_EXTRA_SIZE: usize = 12;
-        let stamped_event_size = serialized_len + STAMP_EXTRA_SIZE;
-
-        // Charge for 3 memory copy operations.
-        // This includes the cost of forming a StampedEvent, copying into the
-        // AMT's buffer on finish, and returning to the client.
-        let memcpy = self.block_memcpy.apply(stamped_event_size);
-
-        // Charge for 2 memory allocations.
-        // This includes the cost of retaining the StampedEvent in the call manager,
-        // and allocaing into the AMT's buffer on finish.
-        let alloc = self.block_allocate.apply(stamped_event_size);
+        // Calculate the cost per copy (one memcpy + one allocation).
+        let mem =
+            self.block_memcpy.apply(estimated_size) + self.block_allocate.apply(estimated_size);
 
         // Charge for the hashing on AMT insertion.
-        let hash = self.hashing_cost[&SupportedHashes::Blake2b256].apply(stamped_event_size);
+        let hash = self.hashing_cost[&SupportedHashes::Blake2b256].apply(estimated_size);
 
         GasCharge::new(
-            "OnActorEventAccept",
-            memcpy + alloc,
-            self.event_accept_per_index_element.flat * indexed_elements
-                + self.event_accept_per_index_element.scale * indexed_bytes
-                + memcpy * 2u32 // deferred cost, placing here to hide from benchmark
-                + alloc // deferred cost, placing here to hide from benchmark
-                + hash, // deferred cost, placing here to hide from benchmark
+            "OnActorEvent",
+            // Charge for validation/storing events.
+            mem + validate_entries + validate_utf8,
+            // Charge for forming the AMT and returning the events to the client.
+            // one copy into the AMT, one copy to the client.
+            hash + (mem * 2u32),
         )
     }
 }
--- a/fvm/src/kernel/default.rs
+++ b/fvm/src/kernel/default.rs
@@ -439,14 +439,14 @@ where
         &self,
         aggregate_sig: &[u8; BLS_SIG_LEN],
         pub_keys: &[[u8; BLS_PUB_LEN]],
-        digests: &[[u8; BLS_DIGEST_LEN]],
+        plaintexts: &[&[u8]],
     ) -> Result<bool> {
         let num_signers = pub_keys.len();
 
-        if num_signers != digests.len() {
+        if num_signers != plaintexts.len() {
             return Err(syscall_error!(
                 IllegalArgument;
-                "unequal numbers of bls public keys and digests"
+                "unequal numbers of bls public keys and plaintexts"
             )
             .into());
         }
@@ -454,11 +454,14 @@ where
         let t = self.call_manager.charge_gas(
             self.call_manager
                 .price_list()
-                .on_verify_aggregate_signature(num_signers),
+                .on_verify_aggregate_signature(
+                    num_signers,
+                    plaintexts.iter().map(|msg| msg.len()).sum(),
+                ),
         )?;
 
         t.record(
-            signature::ops::verify_bls_aggregate(aggregate_sig, pub_keys, digests).or(Ok(false)),
+            signature::ops::verify_bls_aggregate(aggregate_sig, pub_keys, plaintexts).or(Ok(false)),
         )
     }
 
--- a/fvm/src/kernel/default.rs
+++ b/fvm/src/kernel/default.rs
@@ -439,7 +439,7 @@ where
         &self,
         aggregate_sig: &[u8; BLS_SIG_LEN],
         pub_keys: &[[u8; BLS_PUB_LEN]],
-        mut plaintexts_concat: &[u8],
+        plaintexts_concat: &[u8],
         plaintext_lens: &[u32],
     ) -> Result<bool> {
         let num_signers = pub_keys.len();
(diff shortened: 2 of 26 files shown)


fvm/src/kernel/default.rs
```diff
@@ -466,6 +466,37 @@ where
...
+
+        let gas_required = {
+            let gas_two_pairings = self.call_manager.price_list().sig_cost[&SignatureType::BLS]
+                .flat
+                .as_milligas();
+            let gas_one_pairing = gas_two_pairings / 2;
+            let num_pairings = digests.len() as u64 + 1;
+            let gas_required = Gas::from_milligas(num_pairings * gas_one_pairing);
+            GasCharge::new("OnVerifySignature", gas_required, Zero::zero())
+        };
```
DrPeterVanNostrand, line 487: Currently, this is just an estimate of the gas cost for aggregate signature verification after hashing has been removed. I tried using the gas calibration test to measure what this should actually be, but wasn't able to get it working.

DrPeterVanNostrand:
Work remaining for this draft PR:
- [ ] Calculate gas charge for signature validation
- [ ] Remove `verify_signature` entirely

DrPeterVanNostrand:
Hi @Stebalien, I'm a bit stuck on how to get the gas charge for aggregate signature verification, is there a standard way of doing this?

Also to clarify, `verify_signature` should be totally replaced with BLS signature verification (i.e. Secp256k1 signatures should be removed)?

fvm/src/kernel/default.rs
Stebalien: That's fine, we can set it to zero for now and measure later.

fvm/src/kernel/default.rs
```diff
@@ -466,6 +466,37 @@ where
...
+        let sig = Signature::new_bls(aggregate_signature.to_vec());
+        let pub_keys: Vec<&[u8]> = pub_keys.iter().map(AsRef::as_ref).collect();
+        let digests: Vec<&[u8]> = digests.iter().map(AsRef::as_ref).collect();
+
+        let gas_required = {
+            let gas_two_pairings = self.call_manager.price_list().sig_cost[&SignatureType::BLS]
+                .flat
+                .as_milligas();
+            let gas_one_pairing = gas_two_pairings / 2;
+            let num_pairings = digests.len() as u64 + 1;
```
Stebalien, line 484: Can we add some comments on why `+1`?

fvm/src/kernel/default.rs
```diff
@@ -466,6 +466,37 @@ where
...
 
+    fn verify_bls_aggregate(
+        &self,
+        aggregate_signature: &[u8; BLS_SIG_LEN],
+        pub_keys: &[[u8; BLS_PUB_LEN]],
+        digests: &[[u8; BLS_DIGEST_LEN]],
+    ) -> Result<bool> {
+        let sig = Signature::new_bls(aggregate_signature.to_vec());
+        let pub_keys: Vec<&[u8]> = pub_keys.iter().map(AsRef::as_ref).collect();
+        let digests: Vec<&[u8]> = digests.iter().map(AsRef::as_ref).collect();
```
Stebalien, line 477: Let's do this after charging gas (payment up front!).

fvm/src/kernel/default.rs
```diff
@@ -466,6 +466,37 @@ where
...
+
+        let gas_required = {
```
Stebalien, line 479: We generally do this by defining a method on the `PriceList`. In this case, we'd pass in `digests.len()` and return the gas charge.

fvm/src/syscalls/crypto.rs
```diff
@@ -47,6 +48,50 @@ pub fn verify_signature(
...
+    let digests: Vec<[u8; BLS_DIGEST_LEN]> = context
+        .memory
+        .try_slice(digests_off, digests_len)?
+        .chunks(BLS_DIGEST_LEN)
+        .map(|digest_bytes| {
+            digest_bytes
+                .try_into()
+                .expect("bls digest bytes slice-to-array conversion should not fail")
+        })
+        .collect();
```
Stebalien, line 87: We should do this without copying (the user isn't paying for that copy yet). Unfortunately, that means re-implementing https://doc.rust-lang.org/std/primitive.slice.html#method.as_chunks (unstable), but that's not the end of the world.

fvm/src/syscalls/crypto.rs
```diff
@@ -47,6 +48,50 @@ pub fn verify_signature(
...
 
+pub fn verify_bls_aggregate(
+    context: Context<'_, impl Kernel>,
+    num_signers: u32,
+    sig_off: u32,
+    pub_keys_off: u32,
+    digests_off: u32,
+) -> Result<i32> {
+    let pub_keys_len = num_signers * BLS_PUB_LEN as u32;
+    let digests_len = num_signers * BLS_DIGEST_LEN as u32;
```
Stebalien, line 59: We need to be careful about integer overflows here.

Stebalien:
> Hi @Stebalien, I'm a bit stuck on how to get the gas charge for aggregate signature verification, is there a standard way of doing this?

We benchmark, but that's a whole process and kind of annoying. We usually just set the charges to zero then figure it out before the upgrade.

> Also to clarify, verify_signature should be totally replaced with BLS signature verification (i.e. Secp256k1 signatures should be removed)?

Yes. But this means we'll need to implement:

https://github.com/filecoin-project/ref-fvm/blob/f31c6d3a64278f98270e5a13fc6e8be11e5c534e/sdk/src/crypto.rs#L24-L44

In terms of `recover_secp_public_key` and this new syscall.

Stebalien:
@DrPeterVanNostrand could you rebase this PR against the `next` branch as this introduces network breaking changes?

fvm/src/kernel/default.rs
DrPeterVanNostrand: Sounds good, thanks.

fvm/src/syscalls/crypto.rs
DrPeterVanNostrand: Overflowing would mean 44739242 signatures are being aggregated, I'm not sure if that's expected. But, I can add some `checked_mul`s.

Stebalien: It's not expected, but that's a parameter from a potentially untrusted user so we can't assume anything.

DrPeterVanNostrand:
Hi @Stebalien, I've made most of the requested changes locally, however I have one remaining question. When you say:

> we'll need to implement [`sdk::crypto::verify_signature`] in terms of `recover_secp_public_key` and this new syscall [`verify_bls_aggregate`]

I don't quite understand why ECDSA public key recovery for curve Secp256k1 would be needed for BLS signature validation (which uses different curve groups and signing algorithm)?

Stebalien:
It's not. `sdk::crypto::verify_signature` verifies a Filecoin signature which can either be a secp256k1 signature _or_ a BLS signature.

- Previously, we provided a single syscall for "verifying signatures".
- Now, we're providing two lower-level syscalls for the cryptographic primitives (recover secp256k1 and verify bls aggregate).

So we need to re-implement the high-level signature verification in terms of those syscalls. Basically, move the logic from the `verify_signature` syscall you're removing _into_ the SDK.

BlocksOnAChain:
@DrPeterVanNostrand Hey, how are we looking on this PR? I wanted to check the current estimation to close it?

DrPeterVanNostrand:
@BlocksOnAChain I've pushed the requested changes and am just working on making CI turn green now

DrPeterVanNostrand:
Hi @Stebalien, I'm having some trouble with CI. Four of the five failing actions seem to be caused by the cargo registry not finding a dependency (I've noticed that other PRs have had similar CI failures), I'm not sure if you have a standard way of handling this?

```
cargo:warning="  Unable to update https://github.com/filecoin-project/ref-fvm?branch=next#af426971"
cargo:warning=""
cargo:warning="Caused by:"
cargo:warning="  object not found - no match for id (af4269718ba99ec286ef33ec61960e5c5d82c657); class=Odb (9); code=NotFound (-3)"
```

I think that the fifth failing action`ubuntu-latest - conformance` is being caused by some of the conformance test vectors expecting the SDK to expose an ABI for the `verify_signature` syscall (which was removed).

Stebalien:
Yeah, that's... cyclic dependency issues and entirely my fault. I'll handle it.

fvm/src/kernel/default.rs
```diff
@@ -436,34 +435,28 @@ impl<C> CryptoOps for DefaultKernel<C>
...
-            }
-        };
-
-        // Verify signature, catching errors. Signature verification can include some complicated
-        // math.
-        t.record(catch_and_log_panic("verifying signature", || {
-            Ok(signature::verify(sig_type, signature, plaintext, &signing_addr).is_ok())
-        }))
+        t.record(
+            signature::ops::verify_bls_aggregate(aggregate_sig, pub_keys, digests).map_err(|e| {
```
Stebalien, line 451: We'd generally just map all errors here to "false", reserving "illegal argument" errors for cases where, e.g., the caller passed arguments that just don't make sense. Is there any way to distinguish between the two?

E.g., maybe first (before we even charge for gas) check if `pub_keys.len() == digests.len()` and return an illegal argument in that case. Otherwise, verify the signature and return `false` for any error case.


For context, "illegal argument" usually means that there's a bug in the smart contract somewhere (i.e., it's the smart contract's fault).

fvm/src/syscalls/crypto.rs
```diff
@@ -4,46 +4,76 @@ use std::cmp;
...
+
+    // Check that the provided number of signatures aggregated does not cause `u32` overflow.
+    let (pub_key_refs_len, digest_refs_len) = num_signers
+        .checked_mul(pub_key_ref_len as u32)
+        .zip(num_signers.checked_mul(digest_ref_len as u32))
+        .ok_or(syscall_error!(
+            IllegalArgument;
+            "number of signatures aggregated ({}) exceeds limit",
+            num_signers
+        ))?;
```
Stebalien, line 45: Ah, sorry. I meant:

1. Keep the reference to a slice of fixed-size arrays types.
2. Avoid copying those arrays.

I.e.:

```rust 
let pub_keys: &[[u8; BLS_PUB_LEN]] = unsafe {
    // (checking for overflows)
    let buf = context.memory.try_slice(pub_keys_off, num_signers * BLS_PUB_LEN)?;
    std::slice::from_raw_parts(pub_keys.as_ptr(), num_signers)
};
```

(we can also debug/static assert the layout/alignment, but that shouldn't be necessary)

Stebalien: We could also make this "nicer" by adding a helper to the `Memory` type. I.e.:

```rust
fn try_slice_chunks<'a, const S: usize>(&'a self, offset: u32, len: u32) -> &'a [[u8; S]];
```

Stebalien (changes requested):

sdk/src/crypto.rs
```diff
@@ -24,30 +24,103 @@ use crate::{status_code_to_bool, sys, SyscallResult};
...
+    let num_signers = signers.len();
+    assert_ne!(
+        num_signers, 0,
+        "no signing addresses were provided for signature verification"
+    );
+    assert_eq!(
+        num_signers,
+        digests.len(),
+        "unequal numbers of signers and digests"
+    );
```
Stebalien, line 41: Please reserve panics for things that are statically impossible, or well documented edge-cases where returning an error would add too much noise.

In this case, we're likely handling user input so panicing is almost never the right answer.

Stebalien (changes requested):

fvm/src/kernel/default.rs
DrPeterVanNostrand: `shared`'s `verify_bls_aggregate` returns an `Err(String)` if any of the byte array arguments (sig, pub keys, digests) represent an invalid curve point or if `pub_keys.len() != digests.len()`. I returned `IllegalArgument` because those were user-supplied errors, but I can also just map those to `false`. The `shared` function returns `Ok(false)` on signature verification failure (including if some of `bls_signatures` security requirements are unmet, e.g. repeated digests being aggregated, or single signer aggregation with the public key `0`).

Stebalien: I would say: invalid curve points are just "invalid signature". Technically, the public keys having invalid curve points could be considered an "invalid argument", but that makes documenting the invalid argument error case more difficult so I'm going to optimize for simpler documentation...

DrPeterVanNostrand: Sounds good. Also, parsing the `String` error returned by `shared::[...]::verify_bls_aggregate` for exclusively curve point errors could get kludgy.

fvm/src/syscalls/crypto.rs
DrPeterVanNostrand: I changed the keys/digests parameters from slices of fixed-size arrays, e.g. `&[[u8; BLS_DIGEST_LEN]]`, to slices of byte slices `&[&[u8]]` because the SDK's `verify_signature` needs to operate on digests/public keys of variable lengths depending on BLS vs Secp signatures. So, in order to to pass the variable-size digests/public keys from the SDK to the Kernel as fixed-size arrays, I would need to perform copies at the SDK layer.

I can revert this change so that the `verify_bls_aggregate` syscall expects `&[[u8; SOME_LEN]]`, however I do think copies at the SDK layer would be required (I'm not sure if that's acceptable or not). If that's ok, then I can add the memory chunking helper.

Stebalien: Copies in the SDK are fine, copies in the kernel/FVM are less great. Basically, we have to explicitly charge for everything we do in the FVM itself, so I prefer to do things in "userland" (where the SDK lives in userland).

Stebalien: It also lets the user (in a future world where we allow arbitrary wasm actors) optimize this operation to avoid the copy, if they care about it. But once we pass from the SDK and into the syscall, the user can't change anything.

fvm/src/kernel/default.rs
Stebalien: Yeah, let's keep it simple. Anything else risks causing forks.

fvm/src/syscalls/crypto.rs
DrPeterVanNostrand: Cool, thanks. That's good info

fvm/src/gas/price_list.rs
```diff
@@ -582,6 +582,22 @@ impl PriceList {
...
+        // When `num_sigs` BLS signatures are aggregated into a single signature, the aggregate
+        // signature verifier must perform `num_sigs + 1` expensive pairing operations (one
+        // pairing on the aggregate signature, and one pairing for each signed plaintext's digest).
+        //
+        // Note that `bls_signatures` rearranges the textbook verifier equation (containing
+        // `num_sigs + 1` full pairings) into a more efficient equation containing `num_sigs + 1`
+        // Miller loops and one final exponentiation.
+        let num_pairings = num_sigs as u64 + 1;
+        let gas_two_pairings = self.sig_cost[&SignatureType::BLS].flat.as_milligas();
+        let gas_one_pairing = gas_two_pairings / 2;
```
Stebalien, line 596: no need to convert from/to `Gas`. The `Gas` type is defined to prevent errors:

- Handles overflow.
- Checks units.

fvm/src/gas/price_list.rs
```diff
@@ -582,6 +582,22 @@ impl PriceList {
         GasCharge::new("OnVerifySignature", gas, Zero::zero())
     }
 
+    /// Returns gas required for BLS aggregate signature verification.
```
Stebalien, line 585: nit: can you add a TODO here to re-benchmark and file an issue?

sdk/src/crypto.rs
```diff
@@ -23,21 +24,78 @@ use crate::{status_code_to_bool, sys, SyscallResult};
 /// NOTE: This only supports f1 and f3 addresses.
 pub fn verify_signature(
     signature: &Signature,
-    signer: &Address,
-    plaintext: &[u8],
+    signers: &[Address],
+    digests: &[&[u8]],
```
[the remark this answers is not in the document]
Stebalien: (unless there's a reason we can't keep it the same, but if there is, we may have a problem...)

shared/src/crypto/signature.rs
```diff
@@ -129,69 +131,129 @@ impl quickcheck::Arbitrary for Signature {
 #[cfg(feature = "crypto")]
 impl Signature {
     /// Checks if a signature is valid given data and address.
-    pub fn verify(&self, data: &[u8], addr: &crate::address::Address) -> Result<(), String> {
-        verify(self.sig_type, &self.bytes, data, addr)
+    pub fn verify(&self, digest: &[u8], addr: &crate::address::Address) -> Result<(), String> {
```
Stebalien, line 134: see below

shared/src/crypto/signature.rs
```diff
@@ -129,69 +131,129 @@ impl quickcheck::Arbitrary for Signature {
...
-            ))
-        }
+        Ok(bls_signatures::verify(&sig, &digests, &pub_keys))
     }
 
     /// Returns `String` error if a secp256k1 signature is invalid.
     pub fn verify_secp256k1_sig(
         signature: &[u8],
-        data: &[u8],
+        hash: &[u8],
```
Stebalien, line 256: see above

Stebalien: NOTE: we can always ass a `verify_secp256k1_sig_digest` function (if we need it), but I'd prefer to keep this one as-is as any downstream users won't notice that we've broken it.

Stebalien (commented):
Ok, I think we're getting down to the final pass.

Thanks for sticking with this!

sdk/src/crypto.rs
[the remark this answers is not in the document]
DrPeterVanNostrand: Oh alright, so `verify_signature` verifies either Secp or non-aggregated BLS signatures and `verify_bls_aggregate` is reserved for aggregated BLS.

fvm/src/gas/price_list.rs
DrPeterVanNostrand: I needed the conversion because `Gas` doesn't `impl Div`, however I can add the `impl Div for Gas` and remove this conversion; I just hadn't done that initially because `on_verify_aggregate_signature`'s internals would probably replaced in the future with looking up a hardcoded value in the `PriceList `.

Stebalien: Ah, I see... Let's leave it for now. We'll probably replace these entirely.

sdk/src/crypto.rs
```diff
@@ -23,21 +24,70 @@ use crate::{status_code_to_bool, sys, SyscallResult};
 /// NOTE: This only supports f1 and f3 addresses.
 pub fn verify_signature(
     signature: &Signature,
-    signer: &Address,
-    plaintext: &[u8],
+    addr: &Address,
+    digest: &[u8],
```
Stebalien, line 28: Can we revert the change to this signature as well?  I.e., call `sdk::crypto::hash` inside this function to compute the correct digest.

Stebalien: Oh dear. It looks like we're using a special hash function here.

We have two options:

1. One option is to define a multihash codec for this hash function, if it doesn't already exist. Then expose the hash function via the `crypto::hash` syscall.
2. Another option would be to pass the entire messages through instead of just the digests.

I guess the question is: can you think of a reason why a user might want to use a different hash function? Can they use a different hash function for BLS signatures? If not, we should go with option 2 (effectively treating "hashing" as part of signature verification).

We split hashing and verification in Secp256k1 because, in theory, the user may want to use a different hash function. But I _don't_ think that's the case here.

The _downside_ of option two is that the bls signature verification becomes annoying. We'd have to:

1. Concatenate all messages together into a single buffer.
2. Pass a pointer to this buffer along with a pointer to an array of message lengths.

E.g., something like what we do for events:

https://github.com/filecoin-project/ref-fvm/blob/5e743b9235eaaef09097fcd2c7af4e5eea51da96/sdk/src/event.rs#L6-L47

Stebalien: For some context: Smart contracts need to be able to verify signatures made by Filecoin wallets, and doing the hashing inside wasm is likely way too expensive.

Stebalien: Related: Is it possible to use different curves for BLS signatures? Does anyone actually _use_ a different curve?

DrPeterVanNostrand: > Can they use a different hash function for BLS signatures?

A BLS signer can use a different hash-to-curve algorithm than the default one used in `bls-signatures` (i.e. a signer may use any hash function that uniquely maps a string of bytes to a G2 point of curve BLS12-381), however hash-to-curve algorithms are generally considered a hard thing to do correctly and (in my experience) a single hash-to-curve algorithm is chosen and hardcoded into a protocol.

It looks like the IETF hash-to-curve standard suggests [just the one hash function for curve BLS12-381](https://datatracker.ietf.org/doc/html/draft-irtf-cfrg-hash-to-curve-12#name-bls12-381-g2).

> can you think of a reason why a user might want to use a different hash function? 

The only thing I can think of is a user wanting to use a gas cost optimized hash function, however I don't know if a secure hash function with that property exists.

DrPeterVanNostrand: > 1. Concatenate all messages together into a single buffer.
> 2. Pass a pointer to this buffer along with a pointer to an array of message lengths.

From the PR comments above, I was under the impression that SDK's `verify_signature` should operate on single-message signatures (i.e. Secp and non-aggregated BLS) and `verify_bls_aggregate` used for multi-message signatures. I think that you are referring to just `verify_bls_aggregate` here?

Stebalien: Ok, if it's standard, I'd say that we make "hashing" the responsibility of the syscall.

> From the PR comments above, I was under the impression that SDK's verify_signature should operate on single-message signatures (i.e. Secp and non-aggregated BLS) and verify_bls_aggregate used for multi-message signatures. I think that you are referring to just verify_bls_aggregate here?

Correct. Basically:

1. `verify_signature` must take the message, not the digest.
2. Given that `verify_signature` takes the message, how can we implement `verify_bls_aggregate`. I.e., `verify_bls_aggregate` needs to either (a) hash inside the SDK (actor) or (b) hash inside the FVM.

DrPeterVanNostrand: > 1. Concatenate all messages together into a single buffer.
> 2. Pass a pointer to this buffer along with a pointer to an array of message lengths.

Rather that copying the plaintexts at the SDK level (i.e. concatenating), couldn't the SDK create an array of plaintext slices `[&[u8]]` (a bytes slice is 16 bytes, which is probably smaller than each plaintext), then pass to the syscall a pointer `*const &[u8] as *const u8` to the allocated array of plaintext slices:

```rust
// SDK
pub fn verify_signature(
    signature: &Signature,
    addr: &Address,
    plaintext: &[u8],
) -> SyscallResult<bool> {
    match sig_type {
        SignatureType::BLS => {
            let sig = sig_bytes.try_into();
            let pub_key = addr.payload();
            // Pass slice of paintext slices to syscall.
            verify_bls_aggregate(sig, &[pub_key], &[plaintext])
        }
        SignatureType::Secp256k1 => {
             let sig = sig_bytes.try_into();
             let digest = hash_blake2b(plaintext);
             let addr_recovered =
                 recover_secp_public_key(digest, sig).map(Address::new_secp256k1)?;
            Ok(addr == &addr_recovered)
        }
}

// SDK
pub fn verify_bls_aggregate(
    sig: &[u8; BLS_SIG_LEN],
    pub_keys: &[[u8; BLS_PUB_LEN]],
    plaintexts: &[&[u8]],
) -> SyscallResult<bool> {
    unsafe {
        sys::crypto::verify_bls_aggregate(
            pub_keys.len() as u32,
            sig.as_ptr() as *const u8,
            pub_keys.as_ptr() as *const u8,
            // Cast `*const &[u8]` to `*const u8`.
            plaintexts.as_ptr() as *const u8,
        )
        .map(status_code_to_bool)
    }
}

// Syscall
pub fn verify_bls_aggregate(
    context: Context<'_, impl Kernel>,
    num_signers: u32,
    sig_off: u32,
    pub_keys_off: u32,
    plaintext_slices_off: u32,
) -> Result<i32> {
    let sig: &[u8; BLS_SIG_LEN] = context.try_slice(...);
    let pub_keys: &[[u8; BLS_PUB_LEN]] = context.try_chunks(...);

    // 16 bytes per slice
    let plaintext_slices_len = num_signers * std::mem::size_of::<&[u8]>();
    let plaintext_slices: &[&[u8]] = context
        .memory
        .try_slice(plaintext_slices_off, plaintext_slices_len)
        .map(|plaintext_slices| unsafe {
            std::slice::from_raw_parts(
                plaintext_slices.as_ptr() as *const &[u8],
                num_signers,
            )
        })?;

     context
        .kernel
        .verify_bls_aggregate(&sig, &pub_keys, plaintexts)
        .map(|v| if v { 0 } else { -1 })
}
```

then the kernel performs plaintext hashing in `CryptoOps::verify_bls_aggregate` via `shared::crypto::signature::verify_bls_aggregate`?

DrPeterVanNostrand: I did this technique in an earlier commit to pass a slice of digest array references `&[&[u8; BLS_DIGEST_LEN]]` from the SDK to syscall without copying the digests (at either the SDK or FVM levels).

Stebalien: Hm. So, the difference is that the "slices" here are fat pointers (that's why they're 16 bytes). And the lengths are dynamic. And, unfortunately, the layout is undefined IIRC so we can't just pass it through:

So, we _could_ likely map the input slice as:

```rust
plaintexts.map(|ptr| [ptr.as_ptr(), ptr.len()]).collect()
```

But this may be more annoying to charge for in terms of gas (and will require an allocation). That is, in the FVM, we'd have to:

1. In the _syscall binding_, we'd have to slice N times (and store those slices in a new vec that we'd have to allocate FVM-side). We can get away with charging nothing for a small constant amount of work (we can consider this to be "covered" by the base syscall gas fee) but, if we perform a variable amount of work, we need to charge for that.
2. Then, in the _kernel_, we'd have to charge for the actual hashing, etc.

If we instead pass a concatenated buffer through, this becomes a bit simpler:

1. We have to slice once in the syscall binding (and don't have to copy anything or charge any gas).
2. We can charge gas all at once in the kernel for hashing, etc. before we do any actual work.

The downside is that we have more copying in the actor, but there are bulk copy instructions in wasm which should compile down to some pretty-well optimized assembly, so it should be fast. We charge 0.4gas/byte copied, which is likely less than a percent of the charge of hashing.

Stebalien: @DrPeterVanNostrand have you had a chance to take a look at this?

DrPeterVanNostrand: Sorry, I had proofs' work last week then was afk for a couple of days. I picked it back up today and am currently adding the plaintext concatenation.

Stebalien: Thanks! And if this slips, it's not the end of the world compared to everything else coming in nv21. This can ship in nv22 without anyone complaining.

fvm/src/syscalls/crypto.rs
```diff
@@ -4,46 +4,66 @@ use std::cmp;
...
+    for len in plaintext_lens {
+        let plaintext = context.memory.try_slice(plaintexts_off, *len)?;
+        plaintexts.push(plaintext);
+        plaintexts_off += len;
+    }
 
     context
         .kernel
-        .verify_signature(sig_type, sig_bytes, &addr, plaintext)
+        .verify_bls_aggregate(sig, pub_keys, &plaintexts)
```
Stebalien, line 66: So, we really try to avoid `O(n)` work before charging for gas. In this case, I'd pass a `plaintext_lens` and `plaintexts` (buffer) directly into `verify_bls_aggregate`. Otherwise, the user could force us to do a lot of work even if they don't actually have the gas to pay for it.

Specifically:

```rust
fn verify_bls_aggregate(sig: &[u8; BLS_SIG_LEN], pub_keys: &[[u8; BLS_PUB_LEN]], plaintext: &[u8], plaintext_lens: &[u32])
```

This _also_ means we don't need to re-sum the lengths when charging for gas.

sdk/src/crypto.rs
```diff
@@ -23,21 +25,83 @@ use crate::{status_code_to_bool, sys, SyscallResult};
...
-            signer.as_ptr(),
-            signer.len() as u32,
-            plaintext.as_ptr(),
-            plaintext.len() as u32,
+        sys::crypto::verify_bls_aggregate(
+            num_signers,
+            sig.as_ptr() as *const u8,
+            pub_keys.as_ptr() as *const u8,
+            plaintext_lens.as_ptr() as *const u8,
+            plaintexts_concat.as_ptr() as *const u8,
```
Stebalien, line 104: I don't think we should be casting in any of these cases. We should be able to pass these pointers as-is (although we may need to change the syscall?).

Stebalien: Well, the `pub_keys` is more debatable. But we should definitely pass `plaintext_lens` as `*const u32`.

Stebalien (commented):
Ok, down to the final nits.

Stebalien:
@DrPeterVanNostrand I'm going to leave this PR here for the moment as it won't make it into nv21 at this point and we're preparing to merge everything in the "next" branch into master, hopefully next week.

Are you up to writing a FIP for this?

DrPeterVanNostrand:
@Stebalien
> Are you up to writing a FIP for this?

Ha, no I haven't, but I also wasn't asked to.

Stebalien:
I know, this is me asking if you're willing/able/have time to write a FIP.

Stebalien:
Hm. That and this will need a few tests.

DrPeterVanNostrand:
@Stebalien I'll push tests cases for the bad inputs this week. I can make some time to write the FIP draft, do you have an idea of when you would want the FIP drafted, or its discussion to start?

Stebalien:
Sooner is always better, although this won't fully land until the next
network upgrade anyways so no rush.

Issue #1837, opened 2023-08-10
Remove beneficiary from self_destruct syscall


Stebalien:
Replaces https://github.com/filecoin-project/ref-fvm/issues/736, see the discussion there.

Discussed in https://github.com/filecoin-project/FIPs/discussions/524. I'm implementing the version where we accept a flag to either burn remaining funds or fail. This forces the user to pick one or the other.

Stebalien:
(this came up when reviewing some APIs as a part of #1151)

PR #1838, opened 2023-08-10
feat: remove beneficiary from self destruct

Stebalien:
fixes https://github.com/filecoin-project/ref-fvm/issues/1837

### DIFF
--- a/fvm/src/kernel/default.rs
+++ b/fvm/src/kernel/default.rs
@@ -11,7 +11,6 @@ use filecoin_proofs_api::{self as proofs, ProverId, PublicReplicaInfo, SectorId}
 use fvm_ipld_blockstore::Blockstore;
 use fvm_ipld_encoding::{bytes_32, IPLD_RAW};
 use fvm_shared::address::Payload;
-use fvm_shared::bigint::Zero;
 use fvm_shared::chainid::ChainID;
 use fvm_shared::consensus::ConsensusFault;
 use fvm_shared::crypto::signature;
@@ -244,7 +243,7 @@ where
         t.record(Ok(self.get_self()?.map(|a| a.balance).unwrap_or_default()))
     }
 
-    fn self_destruct(&mut self) -> Result<()> {
+    fn self_destruct(&mut self, burn_unspent: bool) -> Result<()> {
         if self.read_only {
             return Err(syscall_error!(ReadOnly; "cannot self-destruct when read-only").into());
         }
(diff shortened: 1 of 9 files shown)


anorth (approved):
Changes LGTM, with disclaimer that I'm not familiar enough with FVM internals to know if there are other changes that should have been made too.

fvm/src/kernel/default.rs
```diff
@@ -224,35 +224,32 @@ where
...
 
-        // Idempotentcy: If the actor doesn't exist, this won't actually do anything. The current
+        // Idempotent: If the actor doesn't exist, this won't actually do anything. The current
         // balance will be zero, and `delete_actor_id` will be a no-op.
         let t = self
             .call_manager
             .charge_gas(self.call_manager.price_list().on_delete_actor())?;
 
+        // If there are remaining funds, burn them. We do this instead of letting the user to
+        // specify the beneficiary as:
```
fridrik01, line 239: One thing I don't get is If there are remaining funds, why don't we just fail the self destruct altogether requiring the user to explicitly transfer (or burn) the funds himself?

Stebalien: We considered that and I'm happy to go that way. I did it this way because it's most likely to "just work". E.g., you write a smart contract that never even expects to _have_ native tokens and someone sends you some anyways.

On the other hand, now that I revisit this... I'm more convinced we _should_ just fail. That ensures that fund transfers can only happen on "send" (and actor creation), not through other means.

Stebalien: Well, kind of. On the other hand, I'm not sure if that matters. Basically, when an actor self destructs, it does kind of make sense to destroy the funds automatically. _Technically_ it's sending them to the burnt funds actor, but that's more of an implementation detail than anything.

Stebalien:
@anorth I've changed this logic to implement your suggestion from https://github.com/filecoin-project/FIPs/discussions/524#discussioncomment-4140910. That way the user has to explicitly decide whether they want to burn remaining funds or fail.

fvm/src/kernel/default.rs
Kubuxu: Achieving zero funds can be hard in cases where this account is paying for gas. While it might not be a worry right now (as key accounts cannot self-destruct), in case of Account Abstraction it gets harder.

Stebalien: Yeah. In that world, I'll _require_ that the user pass specify that they want remaining funds burnt.

PR #1841, opened 2023-08-11
fix: use ptr::null instead of casting 0 to a pointer

Stebalien:
Clippy is getting angry about this.


fridrik01 (approved):
LGTM

~~Wondering why this was not caught in [clippy ci](https://github.com/filecoin-project/ref-fvm/actions/runs/5649014356/job/15302471843?pr=1807) for [event syscall PR](https://github.com/filecoin-project/ref-fvm/pull/1807)~~ Actually it was caught, I just missed it :(

PR #1845, opened 2023-08-16
chore: release fvm 2.6.0

arajasek:
As in the title.

### DIFF
--- a/fvm/CHANGELOG.md
+++ b/fvm/CHANGELOG.md
@@ -4,6 +4,16 @@ Changes to the reference FVM implementation.
 
 ## [Unreleased]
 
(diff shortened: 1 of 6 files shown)


fvm/CHANGELOG.md
```diff
@@ -4,6 +4,13 @@ Changes to the reference FVM implementation.
 
 ## [Unreleased]
 
+## 2.6.0 (2023-08-16)
+
+Breaking Changes:
```
Stebalien, line 9: Can we document what the client needs to do here?

Stebalien (approved):

Stebalien:
Let's merge and release this when we do the same in v3 to make it easier on clients.

PR #1846, opened 2023-08-17
feat: support nv21

arajasek:
For now, we "support" nv21 on master, and continue to use the Hygge pricelist for it. We will eventually move to a new pricelist, and NOT support earlier versions as we integrate breaking changes from `next` into `master`.


Stebalien:
I have some reservations about doing this on master before it's actually ready... although I guess that's what we usually do? I guess it won't really matter.

Maybe we can add an `nv21-alpha` feature and put it behind that flag?

Stebalien:
One funky part about this is that:

1. We're likely to release a new "point" release of the FVM for, e.g., the randomness changes (so we can bubble those now). However, FVM v3 likely _won't_ support nv21. Instead, we'll likely have an FVM v4 by then.

arajasek:
@Stebalien The way I think about it is that "nv21" is a fluid thing. Right now, I'd like to run a devnet in Lotus that upgrades to (something called) nv21, and tests out the migration we have written out, etc. In order to do that, I need some VM running nv21, which currently doesn't _really_ exist. I could patch to use `next` or some other branch that supports nv21, but I don't necessarily want to pull in all the changes in those branches.

Put another way, if we define supporting nv21 as the stuff we currently have in actors master, migration currently written, etc., then FVM3 _does_ support it. It will stop doing so eventually.

Stebalien:
Yeah... my concern is that someone could "think" they support nv21 because they're using this version of the FVM, even though it doesn't.

On the other hand, that's always been an issue. Can we... put it behind a feature flag and enable that feature flag in the FFI?

arajasek:
Now behind feature flag

Stebalien (approved):

Issue #1847, opened 2023-08-17
Stop hashing to "draw" randomness in the kernel


arajasek:
Instead, for the `get_beacon_randomness` and `get_chain_randomness` syscalls, simply return the digest to the caller, who can then request a second hash as they please.

PR #1848, opened 2023-08-17
feat: FVM no longer draws randomness for you

arajasek:
#1847 

I might need to iron out the FVM-actors-FVM dependencies for green CI, but should be reviewable.

### DIFF
--- a/fvm/src/kernel/default.rs
+++ b/fvm/src/kernel/default.rs
@@ -751,19 +751,18 @@ where
         &self,
         rand_epoch: ChainEpoch,
     ) -> Result<[u8; RANDOMNESS_LENGTH]> {
-        let current_epoch = self.call_manager.context().epoch;
-        if rand_epoch > current_epoch {
-            return Err(
-                syscall_error!(IllegalArgument; "randomness epoch {} is in the future", rand_epoch)
-                    .into(),
-            );
-        }
+        let lookback = self
+            .call_manager
+            .context()
+            .epoch
+            .checked_sub(rand_epoch)
+            .ok_or_else(
+                syscall_error!(IllegalArgument; "randomness epoch {} is in the future", rand_epoch),
+            )?;
 
-        let t = self.call_manager.charge_gas(
-            self.call_manager
-                .price_list()
-                .on_get_randomness(current_epoch - rand_epoch),
-        )?;
+        let t = self
+            .call_manager
+            .charge_gas(self.call_manager.price_list().on_get_randomness(lookback))?;
 
         t.record(
             self.call_manager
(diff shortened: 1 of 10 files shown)


fvm/src/kernel/default.rs
```diff
@@ -748,79 +745,51 @@ where
...
-        personalization: i64,
         rand_epoch: ChainEpoch,
-        entropy: &[u8],
     ) -> Result<[u8; RANDOMNESS_LENGTH]> {
         let t = self.call_manager.charge_gas(
             self.call_manager
                 .price_list()
-                .on_get_randomness(entropy.len()),
+                // TODO
+                .on_get_randomness(),
```
Stebalien, line 760: Let's pass in the lookback length and put this "todo" in the price list itself (that way, _all_ of our todos are in the price-list side of things and we can merge this before we necessarily have final gas numbers.

fvm/src/kernel/default.rs
```diff
@@ -748,79 +745,51 @@ where
...
         )?;
 
         // TODO(M2): Check error code
         // Specifically, lookback length?
```
Stebalien, line 764: Let's handle this TODO and remove it. At this point, all we have to do is verify that the epoch is before the current epoch.

Stebalien (approved):
Logic LGTM

fvm/src/kernel/default.rs
```diff
@@ -746,79 +743,60 @@ where
...
+                syscall_error!(IllegalArgument; "randomness epoch {} is in the future", rand_epoch)
+                    .into(),
+            );
+        }
+
         let t = self.call_manager.charge_gas(
             self.call_manager
                 .price_list()
-                .on_get_randomness(entropy.len()),
+                .on_get_randomness(self.call_manager.context().epoch - rand_epoch),
```
Stebalien, line 791:
```suggestion
                .on_get_randomness(current_epoch - rand_epoch),
```

Stebalien (approved):

Stebalien:
Nest step: disable conformance tests in `next` then merge this.

fvm/src/kernel/default.rs
```diff
@@ -746,79 +743,60 @@ where
...
     C: CallManager,
 {
     fn get_randomness_from_tickets(
         &self,
-        personalization: i64,
         rand_epoch: ChainEpoch,
-        entropy: &[u8],
     ) -> Result<[u8; RANDOMNESS_LENGTH]> {
+        let current_epoch = self.call_manager.context().epoch;
+        if rand_epoch > current_epoch {
```
Stebalien, line 755: FYI... it _is_ possible to write this as:

```suggestion
        let lookback = current_epoch.checked_sub(rand_epoch)
            .ok_or_else(syscall_error!(IllegalArgument; "randomness epoch {} is in the future", rand_epoch))?;
```

Stebalien: (lets you skip the subtraction below)

Stebalien: Not that you necessarily should do it this way. I just figured I'd let you know.

sdk/src/rand.rs
```diff
@@ -9,28 +9,16 @@ use crate::{sys, SyscallResult};
...
-    dst: i64,
-    round: ChainEpoch,
-    entropy: &[u8],
-) -> SyscallResult<[u8; RANDOMNESS_LENGTH]> {
-    let ret = unsafe {
-        sys::rand::get_beacon_randomness(dst, round, entropy.as_ptr(), entropy.len() as u32)?
-    };
+pub fn get_beacon_randomness(round: ChainEpoch) -> SyscallResult<[u8; RANDOMNESS_LENGTH]> {
+    let ret = unsafe { sys::rand::get_beacon_randomness(round)? };
     Ok(ret)
```
Stebalien, line 23: nit: no need for the temporary `ret` variable, the entire function can be written as:

```
unsafe { sys::rand::get_beacon_randomness(round) }
```

Stebalien (approved):
I have carekfully checked that we're using the right randomness functions.

PR #1849, opened 2023-08-18
chore: release fvm, fvm_shared, and HAMT crates

arajasek:
As in the title, I intend to use these releases to support nv21 development in Lotus.

### DIFF
--- a/ipld/hamt/CHANGELOG.md
+++ b/ipld/hamt/CHANGELOG.md
@@ -1,5 +1,17 @@
 # Changelog
 
+Changes to the reference FVM's HAMT implementation.
+
+## [Unreleased]
+
+## 0.8.0 [2023-08-18)
+
+Breaking Changes:
+
+- Deprecate default bitwidths in the HAMT
+  - Users must now always specify the bitwidth
+- TODO: Something about #1808?
+
 ## 0.7.0 [2023-06-28)
 
 Breaking Changes:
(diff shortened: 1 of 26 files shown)


shared/CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 ## [Unreleased]
 
+TODO: Can this be a patch release?
```
arajasek, line 5: TODO: I think patch is fine here? But I'm not sure.

ipld/hamt/CHANGELOG.md
```diff
@@ -1,5 +1,17 @@
...
+
+## [Unreleased]
+
+## 0.8.0 [2023-08-18)
+
+Breaking Changes:
+
+- Deprecate default bitwidths in the HAMT
+  - Users must now always specify the bitwidth
+- TODO: Something about #1808?
```
arajasek, line 13: I don't wanna go diving for context since the PR description doesn't say much here, but if someone wants to add something here, please do. If not will just drop.

Stebalien:
Ah, we need https://github.com/filecoin-project/ref-fvm/pull/1830 first. Well, at least we need to fix the _new_ bug I introduced.

shared/CHANGELOG.md
Stebalien: Eh, I'd just go with minor. It doesn't really hurt.

Stebalien (approved):

PR #1854, opened 2023-08-22
feat: charge explicit gas for batch verification

### LINKED ISSUE #1852: Fix gas accounting for seal proof verification
Upcoming changes to built-in actors to implement [direct data onboarding](https://github.com/filecoin-project/FIPs/discussions/730) will invoke the `batch_verify_seals` syscall directly from the miner actor. In the past, this syscall has been invoked only from the power actor's cron handler. It does not charge the correct amount of gas, though there's no apparent reason why it can't. This syscall should instead charge the correct amount of gas. Doing so won't have any effect on the existing use since the cron originator has unlimited gas budget.

Fixing this is necessary for the deployment of direct data onboarding. I don't think it needs a FIP since the end result is just that the syscall works as expected, with no change to existing on-chain observable behaviour. Any necessary things can be specified in the direct onboarding FIP.

- [ ] Fix the comment in [kernel/mod.rs](https://github.com/filecoin-project/ref-fvm/blob/bcfc903c1c4fdcae9dad8fdf3516b1bd6e92c900/fvm/src/kernel/mod.rs#L311-L316) by basically deleting it. The call is not privileged and should charge gas (reference to the pre-payment is both wrong and irrelevant here).
- [ ] Delete the comment in [kernel/default.rs](https://github.com/filecoin-project/ref-fvm/blob/master/fvm/src/kernel/default.rs#L589-L590)
- [ ] Fix the gas calculation in [kernel/default.rs](https://github.com/filecoin-project/ref-fvm/blob/master/fvm/src/kernel/default.rs#L595). A a minimum this might involve changing the [price list item](https://github.com/filecoin-project/ref-fvm/blob/bcfc903c1c4fdcae9dad8fdf3516b1bd6e92c900/fvm/src/gas/price_list.rs#L132) for `verify_seal_base` from 2000 to a real number for per-proof verification.

Changing the `verify_seal_base` price to `34721049` the amount [hardcoded in the power actor](https://github.com/filecoin-project/builtin-actors/blob/master/actors/power/src/lib.rs#L48), would then price this the same as the current cost of seal proof validation. However this price might (?) include an empirical discount that assumes some unknown number of additional proofs are submitted in the same epoch which can be verified in parallel. But a fixed price per item also obviously doesn't include a discount for the known number of proofs that actually are submitted when this is called directly from the miner.

FYI @ZenGround0 @Kubuxu

Stebalien:
This won't currently make a difference on mainnet because we only call this from the cron actor (in an implicit message). However, we _will_ need this when we have the miner actor call it directly.

fixes #1852

### DIFF
--- a/fvm/src/gas/price_list.rs
+++ b/fvm/src/gas/price_list.rs
@@ -27,6 +27,9 @@ use crate::kernel::SupportedHashes;
 // https://docs.rs/wasmtime/2.0.2/wasmtime/struct.InstanceLimits.html#structfield.table_elements
 const TABLE_ELEMENT_SIZE: u32 = 8;
 
+// Parallelism for batch seal verification.
+const BATCH_SEAL_PARALLELISM: usize = 8;
+
 /// Create a mapping from enum items to values in a way that guarantees at compile
 /// time that we did not miss any member, in any of the prices, even if the enum
 /// gets a new member later.
@@ -129,7 +132,10 @@ lazy_static! {
         tipset_cid_historical: Gas::new(215_000),
 
         compute_unsealed_sector_cid_base: Gas::new(98647),
-        verify_seal_base: Gas::new(2000), // TODO revisit potential removal of this
+        verify_seal_batch: ScalingCost {
+            flat: Gas::zero(), // TODO: Determine startup overhead.
+            scale: Gas::new(34721049), // TODO: Maybe re-benchmark?
+        },
 
         verify_aggregate_seal_per: [
             (
@@ -426,7 +432,7 @@ pub struct PriceList {
     pub(crate) tipset_cid_historical: Gas,
 
     pub(crate) compute_unsealed_sector_cid_base: Gas,
-    pub(crate) verify_seal_base: Gas,
+    pub(crate) verify_seal_batch: ScalingCost,
     pub(crate) verify_aggregate_seal_per: HashMap<RegisteredSealProof, Gas>,
     pub(crate) verify_aggregate_seal_steps: HashMap<RegisteredSealProof, StepCost>,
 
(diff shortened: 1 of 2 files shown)


fvm/src/gas/price_list.rs
```diff
@@ -27,6 +27,9 @@ use crate::kernel::SupportedHashes;
 // https://docs.rs/wasmtime/2.0.2/wasmtime/struct.InstanceLimits.html#structfield.table_elements
 const TABLE_ELEMENT_SIZE: u32 = 8;
 
+// Parallelism for batch seal verification.
+const BATCH_SEAL_PARALLELISM: usize = 8;
```
Stebalien, line 31: Is this right?

fvm/src/gas/price_list.rs
```diff
@@ -129,7 +132,10 @@ lazy_static! {
         tipset_cid_historical: Gas::new(215_000),
 
         compute_unsealed_sector_cid_base: Gas::new(98647),
-        verify_seal_base: Gas::new(2000), // TODO revisit potential removal of this
+        verify_seal_batch: ScalingCost {
+            flat: Gas::zero(), // TODO: Determine startup overhead.
+            scale: Gas::new(34721049), // TODO: Maybe re-benchmark?
```
Stebalien, line 137: Probably need to revisit/benchmark.

fvm/src/kernel/default.rs
```diff
@@ -575,53 +575,34 @@ where
...
-                                    err
-                                );
-                                false
-                            }
-                        }
+        let t = self
+            .call_manager
+            .charge_gas(self.call_manager.price_list().on_batch_verify_seal(vis))?;
+
+        // TODO: consider optimizing for one proof (i.e., don't charge a batch overhead?)
```
Stebalien, line 582: This may not be necessary, it depends on the 'startup' overhead.

fvm/src/kernel/default.rs
```diff
@@ -575,53 +575,34 @@ where
...
+        let out = vis
+            .par_iter()
+            .map(|seal| {
+                match catch_and_log_panic("verifying seals", || verify_seal(seal)) {
```
Stebalien, line 586: do we want to allow partial success when called directly from the miner actor?

Stebalien:
Closing in favor of #1882.

PR #1857, opened 2023-08-25
fix(fevm-bench): don't give contracts endowments

Stebalien:
1. We were giving endowments to contracts by default, which fails with no helpful error when the contracts constructor isn't "payable".
2. Include other misc fixes/improvements found while debugging the first issue.

### DIFF
--- a/tools/fvm-bench/Cargo.toml
+++ b/tools/fvm-bench/Cargo.toml
@@ -11,3 +11,4 @@ anyhow = "1.0.71"
 clap = { version = "4.3.9", features = ["derive"] }
 hex = "0.4.3"
 env_logger = "0.10.0"
+fvm = { path = "../../fvm", default-features = false }
--- a/tools/fvm-bench/Cargo.toml
+++ b/tools/fvm-bench/Cargo.toml
@@ -10,3 +10,4 @@ fvm_shared = { path = "../../shared" }
 anyhow = "1.0.71"
 clap = { version = "4.3.9", features = ["derive"] }
 hex = "0.4.3"
+env_logger = "0.10.0"
--- a/tools/fvm-bench/src/fevm.rs
+++ b/tools/fvm-bench/src/fevm.rs
@@ -18,11 +18,21 @@ pub fn run(
     let create_res = testkit::fevm::create_contract(tester, &mut account, contract)?;
 
     if create_res.msg_receipt.exit_code.value() != 0 {
-        return Err(anyhow!(
-            "actor creation failed: {} -- {:?}",
-            create_res.msg_receipt.exit_code,
-            create_res.failure_info,
-        ));
+        println!("Execution trace:");
+        for tr in create_res.exec_trace {
+            println!("{:?}", tr)
+        }
+        return Err(match create_res.failure_info {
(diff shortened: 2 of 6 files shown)


tools/fvm-bench/Cargo.toml
```diff
@@ -10,3 +10,4 @@ fvm_shared = { path = "../../shared" }
 anyhow = "1.0.71"
 clap = { version = "4.3.9", features = ["derive"] }
 hex = "0.4.3"
+env_logger = "0.10.0"
```
fridrik01, line 13: are we using this logger? I don't see

tools/fvm-bench/src/main.rs
```diff
@@ -48,6 +48,7 @@ struct Args {
 }
 
 fn run() -> anyhow::Result<()> {
+    env_logger::init();
```
fridrik01, line 51: are we using this logger somewhere? I don't see that we are using any of its log macros

tools/fvm-bench/src/fevm.rs
```diff
@@ -18,11 +18,21 @@ pub fn run(
...
+            Some(fi) => anyhow!(
+                "contract creation failed: {} -- {}",
+                create_res.msg_receipt.exit_code,
+                fi
+            ),
+            None => anyhow!(
+                "contract creation failed: {}",
+                create_res.msg_receipt.exit_code
+            ),
+        });
```
fridrik01, line 35: Why not keep this like it was?

tools/fvm-bench/src/fevm.rs
```diff
@@ -37,11 +47,19 @@ pub fn run(
     let invoke_res = testkit::fevm::invoke_contract(tester, &mut account, actor, &input_data, gas)?;
 
     if !invoke_res.msg_receipt.exit_code.is_success() {
-        return Err(anyhow!(
-            "contract invocation failed: {} -- {:?}",
-            invoke_res.msg_receipt.exit_code,
-            invoke_res.failure_info,
-        ));
+        return Err(match invoke_res.failure_info {
```
fridrik01, line 50: would printing traces not be useful here as well?

tools/fvm-bench/src/main.rs
Stebalien: We do in the FVM itself but we don't initialize it in the FVM as it's supposed to be initialized in the top-level project. I'm initializing it here to make debugging easier.

tools/fvm-bench/src/fevm.rs
Stebalien: "{:?}" versus "{}". It was _debug_ printing backtraces (hard to read) instead of formatting them as strings (we have a nice and readable format).

Unfortunately, `Option` doesn't implement `Display`, hence this logic.

tools/fvm-bench/src/fevm.rs
Stebalien: We do below. But yeah, I should probably move things around.

Stebalien:
Ok, I refactored it to print traces in both cases (an to dedup some code).

fridrik01 (approved):

PR #1864, opened 2023-08-30
feat: hamt: paranoid validation when reading nodes

Stebalien:
This duplicates some logic we already have in the go hamt implementation but failed to replicate here. The lack of validation here won't cause any issues on the current network as we only read HAMTs we've written, never untrusted HAMTs. However, we should fix this anyways.

### DIFF
--- a/ipld/hamt/src/node.rs
+++ b/ipld/hamt/src/node.rs
@@ -6,12 +6,13 @@ use std::borrow::Borrow;
 use std::fmt::Debug;
 use std::marker::PhantomData;
 
+use cid::Cid;
 use fvm_ipld_blockstore::Blockstore;
-use fvm_ipld_encoding::CborStore;
+use fvm_ipld_encoding::{CborStore, DAG_CBOR};
 use multihash::Code;
 use once_cell::unsync::OnceCell;
 use serde::de::DeserializeOwned;
-use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use serde::{Serialize, Serializer};
 
 use super::bitfield::Bitfield;
 use super::hash_bits::HashBits;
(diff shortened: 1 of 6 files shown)


Stebalien:
Thanks @creativcoder for noticing that something here was wrong in #1561. Could I ask for a review?

ipld/hamt/src/node.rs
```diff
@@ -48,17 +49,94 @@ where
...
+    K: PartialOrd + DeserializeOwned,
     V: DeserializeOwned,
     Ver: Version,
 {
-    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
-    where
-        D: Deserializer<'de>,
-    {
-        let (bitfield, pointers) = Deserialize::deserialize(deserializer)?;
+    pub fn load(
```
Stebalien, line 68: Ported from

https://github.com/filecoin-project/go-hamt-ipld/blob/d7766f8628ec68928eca53ffc9a9e27f80b67d4e/hamt.go#L354-L419

ipld/hamt/src/node.rs
```diff
@@ -48,17 +49,94 @@ where
...
+        store: &impl Blockstore,
+        k: &Cid,
+        depth: u32,
+    ) -> Result<Self, Error> {
+        let (bitfield, pointers): (Bitfield, Vec<Pointer<K, V, H, Ver>>) = store
+            .get_cbor(k)?
+            .ok_or_else(|| Error::CidNotFound(k.to_string()))?;
+
+        if pointers.len() > 1 << conf.bit_width {
+            return Err(Error::Dynamic(anyhow::anyhow!(
```
Stebalien, line 79: I'm planning on refactoring all the errors later anyways.

Stebalien:
Is there someone available on the @filecoin-project/forest team to review this?

lemmih (approved):

PR #1865, opened 2023-08-30
fix: use better types in HAMT hashing

Stebalien:
Using a u8 here means we can statically guarantee that no bitfield access will panic due to a bounds error.

Based on https://github.com/filecoin-project/ref-fvm/pull/1864


Stebalien:
cc @creativcoder

Issue #717, opened 2021-10-05
Idea: Upgradable actors


raulk:

Stebalien:
No, assuming we can reach consensus on the current actor addressing proposals.

maciejwitowski:
> No, assuming we can reach consensus on the current actor addressing proposals.
@Stebalien does it mean we don't need this now? Or we'll need it later?

PR #1866, opened 2023-08-30
Add support for actor upgrades

fridrik01:
See: https://github.com/filecoin-project/ref-fvm/issues/717

This PR adds support for actor upgrades through a new `sdk::actor::upgrade_actor` syscall. This allows actors to upgrade to a new version while still keeping the same address, balance, etc.

Instead of introducing a separate syscall to fetch `get_old_code_cid` this PR implements the alternative where we pass a upgrade metadata to the upgrade endpoint as proposed [here](https://github.com/filecoin-project/FIPs/discussions/396#discussioncomment-3096870).

Still remaining:
- Reject calls to `upgrade_actor` on self and update tests to use different actors.
- ~~Detect reentrancy attack~~
- ~~Add tests for calling upgrade recursively~~
- ~~Improved error check~~
- ~~Exec traces~~


fvm/src/call_manager/default.rs
```diff
@@ -412,6 +416,198 @@ where
...
+                    state.sequence,
+                    None,
+                ),
+            );
+        }
+
+        Ok(result)
+    }
+
+    fn upgrade_actor_inner<K: Kernel<CallManager = Self>>(
```
Stebalien, line 465: Please deduplicate this with the "send" logic, it'll make it much easier to maintain and review.

Stebalien: I.e., ideally we just have a single (internal) `call_actor` method on the call manager that takes an `Entrypoint`. We can either:

1. Use something like a visitor pattern (kind of complex but pluggable).
2. Make `Entrypoint` a "dumb" enum (like in https://github.com/filecoin-project/builtin-actors/pull/1339). This'll require a _little_ more logic in the `call_actor` method but shouldn't be too complex.

Externally, I think we'd still have to have both an upgrade and a send entrypoint.

fvm/src/call_manager/default.rs
```diff
@@ -412,6 +416,198 @@ where
...
+            new_code_cid
+        );
+        self.map_mut(|cm| {
+            let engine = cm.engine.clone(); // reference the RC.
+
+            // Make the kernel.
+            let kernel = K::new(
+                cm,
+                block_registry,
+                actor_id,
```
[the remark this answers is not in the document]
Stebalien: I.e., the new code _knows_ it's being called by itself but it will still likely want to know who the ultimate caller was.

fvm/src/call_manager/default.rs
```diff
@@ -412,6 +416,198 @@ where
...
+                1919,
+                TokenAmount::from_atto(0),
+                false,
+            );
+
+            // Make a store.
+            let mut store = engine.new_store(kernel);
+
+            // TODO: a hack until I find a better/simpler way to do this
```
Stebalien, line 518: So, honestly, we can probably treat these errors as "fatal" and check earlier. I.e., the _final_ version of upgrade (not for this PR, but later) would:

1. Lookup the target code CID in some "deployed actors" table in the init actor.
2. Check some metadata generated on deploy to see if the actor supports the upgrade entrypoint.

By the time we get to _this_ code, everything will have been checked and any errors would be considered "fatal".

---

For _now_, what we'd likely do is check if:

1. The caller is one of builtin actor types X/Y/Z (in this case, it would be restricted to an eth account).
2. The target code CID is an EVM actor (again, using the builtin actors manifest).

Because that's the immediate use-case (upgrading eth accounts into EVM actors).

fvm/src/call_manager/default.rs
```diff
@@ -412,6 +416,198 @@ where
...
+                .get_actor(actor_id)?
+                .ok_or_else(|| syscall_error!(NotFound; "actor not found: {}", actor_id))?
+                .code;
+            if code_id_before != code_id_after {
+                return Err(syscall_error!(Forbidden; "re-entrant upgrade detected").into());
+            }
+
+            self.state_tree_mut().set_actor(
+                origin,
+                ActorState::new(
```
Stebalien, line 452: Let's set the code cid _first_, then call upgrade. If upgrade fails, we'll roll back the transaction.

fvm/src/call_manager/default.rs
Stebalien: Also, see my comment on setting the code CID before upgrading. That should make this simpler.

fvm/src/call_manager/default.rs
```diff
@@ -412,6 +416,198 @@ where
         Ok(())
     }
 
+    fn upgrade_actor<K: Kernel<CallManager = Self>>(
```
Stebalien, line 419: We'll have to make sure to wrap this in a transaction. I.e., share the logic in `send`.

Which is annoying because I just removed the separate "transaction" abstraction... but oh well...

Stebalien (commented):
initial feedback

fvm/src/call_manager/default.rs
```diff
@@ -412,6 +405,62 @@ where
...
+        &mut self,
+        actor_id: ActorID,
```
Stebalien, line 410: We can pass in the "caller" here to preserve it.  It's kind of strange, but I don't think we have any better options.

fvm/src/call_manager/default.rs
```diff
@@ -412,6 +405,62 @@ where
...
+                state.sequence,
+                None,
+            ),
+        );
+
+        // run the upgrade entrypoint
+        let result = self.send::<K>(
+            actor_id,
+            Address::new_id(actor_id),
+            Entrypoint::Upgrade(UpgradeInfo { old_code_cid: code }),
```
Stebalien, line 442: Exposing _both_ an `upgrade_actor` method _and_ a send method that can take an `EntryPoint::Upgrade` is a bit strange. I'd rename `send` to `call_actor` and either:

1. Make `call_actor` public, moving the "upgrade" logic into the kernel (if reasonable, I haven't tried it?).
2. Make `call_actor` private and export a separate `send` method. In this case, I'd make `with_transaction` private again as we'd be handling the transaction logic internally.

I'm not entirely sure which way to go. IMO, the second option is "cleaner" from an API perspective (hides internals). However, the first one is more flexible and moves more logic into the kernel (which, IMO, is a good thing). I.e., it means that the call manager doesn't actually need to know about "upgrades".

Stebalien: If we go with option 1, I'd make `Entrypoint` a _trait_ so we can define new entrypoints without touching the call manager.

fvm/src/call_manager/default.rs
```diff
@@ -955,3 +1040,53 @@ impl EventsAccumulator {
...
+    fn new(entrypoint: Entrypoint) -> Self {
+        Self {
+            entrypoint,
+            params: Vec::new(),
+        }
+    }
+
+    fn maybe_put_registry(&mut self, br: &mut BlockRegistry) -> Result<()> {
+        match self.entrypoint {
+            Entrypoint::Invoke(_) => Ok(()),
```
Stebalien, line 1077: IMO, we should treat invoke params the same way (if possible).

fvm/src/call_manager/default.rs
```diff
@@ -955,3 +1040,53 @@ impl EventsAccumulator {
...
+
+impl EntrypointParams {
+    fn new(entrypoint: Entrypoint) -> Self {
+        Self {
+            entrypoint,
+            params: Vec::new(),
+        }
+    }
+
+    fn maybe_put_registry(&mut self, br: &mut BlockRegistry) -> Result<()> {
```
Stebalien, line 1075: So, if we go the trait route, I'd define it as:

```rust
trait Entrypoint {
    fn into_params(self, br: &mut BlockRegistry) -> Result<EntryPointParams>;
}
```

This means there's no concerns about `maybe_put_registry` being called multiple times (e.g.).

Stebalien (commented):
Some more thoughts. Let's talk tomorrow.

fvm/src/call_manager/mod.rs
```diff
@@ -191,3 +199,47 @@ pub struct FinishRet {
...
+            Entrypoint::Upgrade(_) => "upgrade",
+        }
+    }
+
+    fn into_params(self, br: &mut BlockRegistry) -> Result<Vec<wasmtime::Val>> {
+        match self {
+            Entrypoint::Invoke(_) => Ok(Vec::new()),
+            Entrypoint::Upgrade(ui) => {
+                let ui_params = to_vec(&ui).map_err(
+                    |e| crate::syscall_error!(IllegalArgument; "failed to serialize upgrade params: {}", e),
```
Stebalien, line 238: IMO, this is a fatal error.

fvm/src/call_manager/mod.rs
```diff
@@ -191,3 +199,47 @@ pub struct FinishRet {
...
+                )?;
+                let block_id = br.put_reachable(kernel::Block::new(CBOR, ui_params, Vec::new()))?;
```
Stebalien, line 240: Can we add a comment on why not `DAG_CBOR`? Specifically, the linked code isn't (currently) considered reachable.

fvm/src/kernel/default.rs
```diff
@@ -864,6 +874,95 @@ where
...
+                    Some(block) => self.blocks.put_reachable(block).unwrap_or_else(|_| {
+                        log::error!("failed to write to kernel block registry");
+                        NO_DATA_BLOCK_ID
+                    }),
+                };
+                Err(ExecutionError::Abort(Abort::Exit(
+                    ExitCode::OK,
+                    "actor upgraded".to_string(),
+                    block_id,
+                )))
```
Stebalien, line 956: Don't return an error here, return success. _Then_, in the syscall binding, we convert to an "abort" to unwind through the calling actor.

Otherwise, we need to infect everything with a new error type that's rarely relevant.

fridrik01: I am not totally sure I follow, so I can change this to `Ok(ExitCode::OK)` but then how do I include the block_id as that should be returned to the caller on success? Also, are you suggesting that we handle the conversion in the fvm/src or sdk/src?, also what exactly do you mean by "unwind through the calling actor"?

Stebalien: Hm.... I'll play with this a bit.

Stebalien: > also what exactly do you mean by "unwind through the calling actor"?

Long jump. I.e., instead of returning to the calling actor, we unwind the calling actor's stack.

---

Motivation:

1. My main motivation is that `Abort` ends up everywhere, and it only makes sense in this one case.
2. The syscall succeeded, so returning an error here is kind of strange. We do use errors to unwind/abort actors, but I prefer to keep the kernel a bit isolated from details like that where possible.

Stebalien: Ok, so, my idea of using `Abort` was a dud because abort always means "abort this actor on error" and we _don't_ want to abort on syscall errors. I.e., we need to be able to return syscall errors to the actor.

We _could_ add an `Exit` variant to `ExecutionError`. Kind of like your `Abort` solution, but it doesn't have duplicate the `OutOfGas` and/or `Fatal` conditions.

However, this case isn't an error so I'd prefer not to. We're using `Abort` as an from syscall bindings, but that's kind of a special case.

Stebalien: Hm. No, `::Exit` is even worse...

Stebalien: https://github.com/filecoin-project/ref-fvm/pull/1910

fvm/src/kernel/default.rs
```diff
@@ -864,6 +873,94 @@ where
...
+                self.caller,
+                Address::new_id(self.actor_id),
+                Entrypoint::Upgrade(UpgradeInfo { old_code_cid: code }),
+                params,
+                &TokenAmount::from_whole(0),
+                None,
+                false,
+            )?;
+
+            if result.exit_code == ExitCode::OK {
```
Stebalien, line 934: IIRC, we usually use `result.exit_code.is_success()` but.... it doesn't really matter.

fvm/src/kernel/default.rs
```diff
@@ -864,6 +873,94 @@ where
...
+                // after running the upgrade, our code cid must not have changed
+                let code_after_upgrade = cm
+                    .get_actor(self.actor_id)?
+                    .ok_or_else(|| syscall_error!(NotFound; "actor not found"))?
+                    .code;
+                if code != code_after_upgrade {
+                    return Err(syscall_error!(Forbidden; "re-entrant upgrade detected").into());
+                }
```
Stebalien, line 942: I _think_ this check should actually be in `send`, not `upgrade`. I.e.:

1. If an actor calls `A (upgrade -> upgrade -> upgrade)`, that's fine. Because, on success, we'll "unwind" past _all_ the upgrade calls all at once.
2. If an actor has a call stack like `A -> B -> A (upgrade)`, that's _not_ fine and we need to abort when we return to the top-level call into `A`. Otherwise, we'd end up running old code.

It's that last point that's problematic: We never want to re-enter some actor code after we've upgraded away from it.

Stebalien: And, in that second case, we should treat it as a "revert" of `B` (I think?).

fvm/src/kernel/error.rs
```diff
@@ -60,8 +60,8 @@ impl ExecutionError {
     pub fn is_recoverable(&self) -> bool {
         use ExecutionError::*;
         match self {
-            OutOfGas | Fatal(_) => false,
             Syscall(_) => true,
+            OutOfGas | Fatal(_) => false,
```
Stebalien, line 64: revert (probably my fault)

Stebalien (commented):
Only major thing is recursion handling. Otherwise, quibbles but LGTM. I'll take one more look tomorrow.

fvm/src/kernel/default.rs
Stebalien: Actually, after talking this through with @maciejwitowski,  there's an alternative here that I kind of like. Instead of catching this when we return to the top-level invocation of `A`, we can catch this on the upgrade syscall itself. To do this, we'd need to:

1. Keep a map of `ActorID -> reentrancy count`, incrementing the count each time we re-enter an actor already on the call-stack.
2. In the upgrade syscall, if this count is non-zero (for the current actor), return a `Forbidden` syscall error (or something like that).

One reason I like this approach is that, independent of _this_ particular change, I'd like to find a way to _expose_ whether or not a call is reentrant to actors. I.e., some kind of flag in their environment where they can detect that they're within a reentrant call. I expect most actors will, at that point, simply abort with an error.

fvm/src/kernel/error.rs
fridrik01: I think we can just delete this method as its not used anymore

fvm/src/kernel/default.rs
fridrik01: @Stebalien I played around with this approach in [109c5e9](https://github.com/filecoin-project/ref-fvm/pull/1866/commits/109c5e9fd7f8369e9ef3d3f940a432c8e276ee10)

Stebalien: Ah, I mean that we need to keep track of _calls_ in progress, not upgrades. I.e., if some actor `A` is already on the call stack, no "deeper" instance of `A` should be able to call the `upgrade` syscall.

fridrik01: Oh lets talk sync today

fvm/src/kernel/default.rs
```diff
@@ -136,10 +137,24 @@ where
             return Err(syscall_error!(LimitExceeded; "cannot store return block").into());
         }
 
+        self.call_manager
+            .get_actor_call_stack_mut()
+            .entry(self.actor_id)
+            .and_modify(|count| *count += 1)
+            .or_insert(1);
```
Stebalien, line 144: This needs to happen within the call manager to handle the call into the top-level actor. It'll also help encapsulate things a bit.

I.e., I'd expose some form of `get_actor_recursion_count(actor_id)`, or something like that, but otherwise keep this datastructure internal (if possible).

fvm/src/call_manager/mod.rs
```diff
@@ -111,6 +119,15 @@ pub trait CallManager: 'static {
...
     ) -> Result<()>;
 
+    // returns the actor call stack
+    fn get_actor_call_stack(&self) -> &Vec<(ActorID, &'static str)>;
+
+    /// add an actor to the calling stack.
+    fn actor_call_stack_push(&mut self, actor_id: ActorID, entrypoint: &Entrypoint);
+
+    /// pop an actor from the calling stack.
+    fn actor_call_stack_pop(&mut self) -> Option<(ActorID, &'static str)>;
```
Stebalien, line 129: Can we avoid adding these to the interface?

fvm/src/call_manager/mod.rs
```diff
@@ -111,6 +119,15 @@ pub trait CallManager: 'static {
         delegated_address: Option<Address>,
     ) -> Result<()>;
 
+    // returns the actor call stack
+    fn get_actor_call_stack(&self) -> &Vec<(ActorID, &'static str)>;
```
Stebalien, line 123:
```suggestion
    fn get_actor_call_stack(&self) -> &[(ActorID, &'static str)];
```

(slightly more idiomatic)

fvm/src/call_manager/mod.rs
fridrik01: definetly!

fvm/src/call_manager/default.rs
```diff
@@ -638,15 +634,19 @@ where
...
-        self.send_resolved::<K>(from, to, method, params, value, read_only)
+        self.actor_call_stack.push((to, entrypoint.func_name()));
+        let res = self.send_resolved::<K>(from, to, entrypoint, params, value, read_only);
+        self.actor_call_stack.pop();
+
+        res
     }
 
     /// Send with resolved addresses.
     fn send_resolved<K>(
```
Stebalien, line 645: now that we're mostly down to nits, I'd rename all `send_*` methods on the call manager to `call_actor_*` to avoid confusion.

fvm/src/kernel/mod.rs
```diff
@@ -209,6 +209,14 @@ pub trait ActorOps {
         delegated_address: Option<Address>,
     ) -> Result<()>;
 
+    fn upgrade_actor<K: Kernel>(
+        &mut self,
+        new_code_cid: Cid,
+        params_id: BlockId,
+    ) -> Result<SendResult>;
+
+    fn can_actor_upgrade(&self) -> bool;
```
Stebalien, line 218: IMO, this should be an implementation detail of the specific kernel, not something exposed on the interface. I.e., don't call it in the syscall binding, unconditionally call the `upgrade_actor` method on the kernel and check this precondition internally.

fvm/src/syscalls/bind.rs
```diff
@@ -49,38 +49,69 @@ pub(super) trait BindSyscall<Args, Ret, Func> {
     ) -> anyhow::Result<&mut Self>;
 }
 
+pub enum ControlFlow<T> {
```
Stebalien, line 52: nit: docs (I know, I wrote this code....)

sdk/src/actor.rs
```diff
@@ -107,6 +107,21 @@ pub fn create_actor(
     }
 }
 
+/// Upgrades an actor using the given block which includes the old code cid and the upgrade params
+pub fn upgrade_actor(new_code_cid: Cid, params: Option<IpldBlock>) -> SyscallResult<Response> {
```
Stebalien, line 111: We can leave it like this for now, but we should have a better API.

- We generally take `Cid` by reference (it's large so we avoid copying till we need to).
- There is no "success" case here.

It should probably look something like:

```suggestion
pub enum UpgradeError {
    CodeNotInstalled, // code not available, could be worded better.
    InvalidUpgradeTarget, // code doesn't implement the upgrade entrypoint.
    ReentrentUpgrade, // actor already on the call stack.
    UpgradeFailed(Response),
}
pub fn upgrade_actor(new_code_cid: Cid, params: Option<IpldBlock>) -> Result<std::convert::Infallible, UpgradeError> {
```

I wish we could use `!` (the correct never type) but that's not stable yet.

Stebalien: The reason to do this is that we want to treat _every_ possible return here as an error. If the user blindly writes `upgrade_actor(...)?` or `upgrade_actor(...).expect("upgrade failed")`, that should work.

testing/test_actors/actors/fil-upgrade-actor/src/actor.rs
```diff
@@ -0,0 +1,128 @@
...
+                None,
+                Default::default(),
+            )
+            .unwrap();
+        }
+        // test that calling an upgrade with actor already on the call stack fails
+        5 => {
+            let new_code_cid = sdk::actor::get_actor_code_cid(&Address::new_id(10000)).unwrap();
+            let res = sdk::actor::upgrade_actor(new_code_cid, None);
+            assert_eq!(res, Err(ErrorNumber::Forbidden));
```
Stebalien, line 116: Now that I think about this, I wonder if this should be `IllegalOperation`? That would let us distinguish between two possible failure cases.

fvm/src/kernel/default.rs
```diff
@@ -864,6 +876,114 @@ where
...
+        if self.blocks.is_full() {
+            return Err(syscall_error!(LimitExceeded; "cannot store return block").into());
+        }
+
+        let result = self.call_manager.with_transaction(|cm| {
+            let origin = cm.origin();
+
+            let state = cm
+                .get_actor(self.actor_id)?
+                .ok_or_else(|| syscall_error!(NotFound; "actor not found"))?;
```
Stebalien, line 927: I think the correct error here is `IllegalOperation`. This can happen if we, e.g., try to upgrade after selfdestructing (which _could_ happen, in theory).

sdk/src/sys/actor.rs
```diff
@@ -128,6 +131,39 @@ super::fvm_syscalls! {
...
+    ///
+    /// # Errors
+    ///
+    /// | Error                 | Reason                                               |
+    /// |-----------------------|------------------------------------------------------|
+    /// | [`NotFound`]          | there's no actor deployed with the target code cid.  |
+    /// | [`InvalidHandle`]     | parameters block not found.                          |
+    /// | [`LimitExceeded`]     | recursion limit reached.                             |
+    /// | [`IllegalArgument`]   | invalid code cid buffer.                             |
+    /// | [`Forbidden`]         | target actor doesn't have an upgrade endpoint.       |
```
Stebalien, line 161: Once we're done with everything else, let's make sure to go back and revisit this.

sdk/src/actor.rs
Stebalien: Hm. `ReentrentUpgrade` may not _quite_ work (see my other comments on the conflict with selfdestruct). Honestly, I'd just change the variant in this enum to `IllegalOperation` and document it as an "either or" case (where, 99% of the time, it'll be because the actor doesn't exist).

testing/integration/tests/main.rs
```diff
@@ -1002,6 +1003,224 @@ fn test_oom4() {
...
+        assert!(
+            res.msg_receipt.exit_code.is_success(),
+            "{:?}",
+            res.failure_info
+        );
+        let val: i64 = res.msg_receipt.return_data.deserialize().unwrap();
+        assert_eq!(val, 666);
+    }
+
+    {
```
Stebalien, line 1154: Let's document what each of these cases is doing.

Stebalien (commented):
Testing: Let's make sure we cover cases like:

- Selfdestruct -> upgrade
- Re-entrent call -> upgrade
- upgrade -> upgrade
- failure in a recursive upgrade where the top-level upgrade sticks (e.g., the code CID should be changed).
- upgrade to some code CID that doesn't exist (maybe?)
- upgrade to some code CID that doesn't implement the `upgrade` entrypoint.

You're probably already covering some of these cases.

fvm/src/call_manager/default.rs
```diff
@@ -764,35 +777,29 @@ where
...
+                }
             })();
 
             let invocation_data = store.into_data();
             let last_error = invocation_data.last_error;
             let (mut cm, block_registry) = invocation_data.kernel.into_inner();
 
-            // Resolve the return block's ID into an actual block, converting to an abort if it
-            // doesn't exist.
-            let result = result.and_then(|ret_id| {
```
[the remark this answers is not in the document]
fridrik01: yes, will undo this refactoring, it also makes the error handling a bit simpler

testing/integration/tests/main.rs
fridrik01: Note that this was already documented in the upgrade-actor

testing/test_actors/actors/fil-upgrade-actor/src/actor.rs
fridrik01: Not sure about changing it to `IllegalOperation` as that implies the actor is not in the correct state, which its not necessarily correct here

fvm/src/kernel/default.rs
fridrik01: Cool yes `IllegalOperation` definetly makes sense here

fridrik01:
> Testing: Let's make sure we cover cases like:
> 
> * Selfdestruct -> upgrade
> * Re-entrent call -> upgrade
> * upgrade -> upgrade
> * failure in a recursive upgrade where the top-level upgrade sticks (e.g., the code CID should be changed).
> * upgrade to some code CID that doesn't exist (maybe?)
> * upgrade to some code CID that doesn't implement the `upgrade` entrypoint.
> 
> You're probably already covering some of these cases.

From this list, we are at least covering:
- Selfdestruct -> upgrade: covered with method 5 in fil-upgrade-actor
- Re-entrent call -> upgrade: covered with method 4 in fil-upgrade-actor
- upgrade -> upgrade:  covered with method 3 in fil-upgrade-actor
- upgrade to some code CID that doesn't implement the upgrade entrypoint: covered in fil-syscall-actor).

From the remaining cases: 
- upgrade to some code CID that doesn't exist (maybe?):  Right now we blindly accept code_cids in the kernel (as long as they could be parsed) and add them to the actor state. What would be the way to go do check if that code actually exists? (maybe `store.get_cbor`)?
-  failure in a recursive upgrade where the top-level upgrade sticks (e.g., the code CID should be changed): Not sure I understand this, so its a case of "upgrade -> upgrade" , how would the 2nd upgrade CID stick around as long as the syscall succeded.

sdk/src/actor.rs
fridrik01: I changed to  take CID by reference, will take a look at infallable/UpgradeError next

testing/integration/tests/readonly_test.rs
```diff
@@ -1,65 +0,0 @@
...
-use fvm_shared::message::Message;
-use fvm_shared::state::StateTreeVersion;
-use fvm_shared::version::NetworkVersion;
-use fvm_test_actors::wasm_bin::READONLY_ACTOR_BINARY;
-use num_traits::Zero;
-
-#[test]
-fn readonly_actor_tests() {
-    // Instantiate tester
-    let mut tester = new_tester(
```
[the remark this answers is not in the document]
Stebalien: 😡

fvm/src/kernel/default.rs
Stebalien: We can remove this now, right?

testing/test_actors/actors/fil-upgrade-actor/src/actor.rs
Stebalien: Yeah, you're probably right. Anyways, now that I went back and looked at the code... we use a `SYS_INVALID_RECEIVER` exit code instead of a `Forbidden` error number to indicate that the target code doesn't implement the "upgrade" entrypoint. So there's no ambiguity anyways.

sdk/src/actor.rs
Stebalien: Per our conversation below, `InvalidUpgradeTarget` likely _isn't_ something we need to include in this enum (that's covered by an exit code, not an error number, and we should probably leave the exit codes alone).

Stebalien:
> upgrade to some code CID that doesn't exist (maybe?): Right now we blindly accept code_cids in the kernel (as long as they could be parsed) and add them to the actor state. What would be the way to go do check if that code actually exists? (maybe store.get_cbor)?

It's probably fine to leave that out, in that case.

> failure in a recursive upgrade where the top-level upgrade sticks (e.g., the code CID should be changed): Not sure I understand this, so its a case of "upgrade -> upgrade" , how would the 2nd upgrade CID stick around as long as the syscall succeded.

I mean, I want to make sure that the following works:

1. Upgrade from code A to code B.
2. Inside code B's upgrade function, try to recursively upgrade to code C. Fail.
3. Return success from code B's upgrade function.

At this point, the actor's code CID should be `B`, not `A` or `C`.

Stebalien:
> upgrade -> upgrade: covered with method 3 in fil-upgrade-actor

Are you sure it's testing the right thing? I.e., `upgrade (to code A) -> upgrade (to code B)` should result in code B.

See https://github.com/filecoin-project/ref-fvm/pull/1866#discussion_r1360115148

Stebalien (approved):
🎉

PR #1871, opened 2023-08-31
chore: update deps

Stebalien:
Preparing for the nv21 code freeze. Notably, this updates wasmtime to v12.


Stebalien:
The integer encoding library is causing some problems with some of our async code, which doesn't need to be async so I might just kill it.

PR #1874, opened 2023-09-02
contract-errors-specified

lordshashank:
Defined contract errors in fvm workbench so as to get more specific error while transaction gets reverted. Included the revert and Panic errors. Currently have included the tests along with changes in simplecoin.sol for testing the new functions made. All the tests were passing. Would remove the tests if not needed.
Any improvement in code style is invited.


tools/fvm-bench/src/fevm.rs
```diff
@@ -79,9 +97,147 @@ pub fn run(
 
     let invoke_res = testkit::fevm::invoke_contract(tester, &mut account, actor, &input_data, gas)?;
     let BytesDe(returnval) = invoke_res.msg_receipt.return_data.deserialize().unwrap();
+    println!("Revert Reason: {}", parse_eth_revert(&returnval).unwrap());
```
Stebalien, line 100: We need to check the `exit_code` here. If it's 33, it's a revert. Otherwise, attempting to parse this as a revert will fail.

Stebalien: I'd also call this to handle reverts from the constructor above (line 87).

Stebalien (commented):
Two small changes but otherwise LGTM. Thanks! And sorry about the slow review.

tools/fvm-bench/src/fevm.rs
lordshashank: > We need to check the `exit_code` here. If it's 33, it's a revert. Otherwise, attempting to parse this as a revert will fail.

Sorry, I don't understand. The panic and revert errors also revert with exit code 33 only ig.

lordshashank: Both done @Stebalien 🙂🚀.
I have included the checking in handleResult now, it would cover all cases imo.

Stebalien (approved):
Thanks!

Stebalien:
Ah, it looks like there's a clippy failure.

Stebalien:
And it looks like the test isn't quite correct.

lordshashank:
Ah! really sorry, when I shifted the code to handleResult(), I made few letters capital and forgot to change it in test. Have resolved the both errors now.

PR #1876, opened 2023-09-06
feat(car): replace integer-encoding with unsigned-varint

Stebalien:
This is what we use in the FVM and the builtin actors. I'm switching to it now because we can't upgrade to the latest version of the integer-encoding library due to a breaking change (the futures it produces are no longer send).

### DIFF
--- a/ipld/car/src/util.rs
+++ b/ipld/car/src/util.rs
@@ -4,7 +4,7 @@
 
 use cid::Cid;
 use futures::{AsyncRead, AsyncReadExt, AsyncWrite, AsyncWriteExt};
-use integer_encoding::{VarIntAsyncReader, VarIntAsyncWriter};
+use unsigned_varint::io::ReadError;
 
 use super::error::Error;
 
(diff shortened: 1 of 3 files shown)


Stebalien:
Broken out of #1871 which I'd like to merge before I cut a release.

ipld/car/src/util.rs
```diff
@@ -44,7 +47,9 @@ pub(crate) async fn ld_write<'a, W>(writer: &mut W, bytes: &[u8]) -> Result<(),
 where
     W: AsyncWrite + Send + Unpin,
 {
-    writer.write_varint_async(bytes.len()).await?;
+    let mut len = unsigned_varint::encode::usize_buffer();
+    let len = unsigned_varint::encode::usize(bytes.len(), &mut len);
+    writer.write_all(len).await?;
```
fridrik01, line 52: Nit, but would find this easier to read than calling it `len` each time

```suggestion
    let mut buff = unsigned_varint::encode::usize_buffer();
    let len = unsigned_varint::encode::usize(bytes.len(), &mut buff);
    writer.write_all(len).await?;
```

fridrik01 (approved):
LGTM

PR #1883, opened 2023-09-12
Add the right dir to run make command

jochasinga:


Stebalien:
It looks like there are several problems here, thanks for noticing. Here's a more complete fix: https://github.com/filecoin-project/ref-fvm/pull/1884.

PR #1884, opened 2023-09-12
chore: fix build instructions

Stebalien:
- Don't set rust flags manually.
- Cargo fmt check, don't actually format.
- Don't allow warnings in the makefile.
- Let rustup install the correct targets automatically.

replaces #1883


fridrik01 (approved):
LGTM

PR #1885, opened 2023-09-13
feat: charge event gas

Stebalien:
This charges the correct event gas and adds a few benchmarks. It also:

1. Tries to remove some variability when recording events by pre-allocating space.`
2. Validates utf8 up-front, then slices the validated string for better performance (and easier benchmarking).

### DIFF
--- a/fvm/src/kernel/default.rs
+++ b/fvm/src/kernel/default.rs
@@ -1061,6 +1061,11 @@ where
             return Err(syscall_error!(IllegalArgument; "total event value lengths exceeded the max size: {} > {MAX_TOTAL_VALUES_LEN}", event_values.len()).into());
         }
 
+        // We validate utf8 all at once for better performance.
+        let event_keys = std::str::from_utf8(event_keys)
+            .context("invalid event key")
+            .or_illegal_argument()?;
+
         let mut key_offset: usize = 0;
         let mut val_offset: usize = 0;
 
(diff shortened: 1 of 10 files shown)


testing/test_actors/actors/fil-gas-calibration-actor/src/actor.rs
```diff
@@ -160,10 +160,26 @@ fn on_send(p: OnSendParams) -> Result<()> {
...
+
+    let iterations = p.iterations as u64;
+    for i in 0..iterations {
+        random_mutations(&mut value, p.seed + i, MUTATION_COUNT);
+
+        let entries: Vec<_> = random_chunk(&value, p.entries, p.seed + i)
+            .into_iter()
+            .map(|d| Entry {
+                flags: p.flags,
+                key: char::MAX.to_string(),
```
fridrik01, line 173: don't we want to vary the key here, and also test for utf8 chars so we can make sure its parsed correctly in the kernel (see my comment above)

Stebalien: I'll add a comment.

I'm testing varying the number of entries and the size of the values, but _not_ the keys as that adds too many variables. Instead:

1. I'm independently testing the cost of validating utf8.
2. I'm assuming that all the _other_ costs related to keys will be the same as the costs for values (both byte buffers under the covers).

fvm/src/kernel/default.rs
```diff
@@ -1098,10 +1103,6 @@ where
                 .context("event entry key out of range")
                 .or_illegal_argument()?;
 
-            let key = std::str::from_utf8(key)
-                .context("invalid event key")
-                .or_illegal_argument()?;
```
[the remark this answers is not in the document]
Stebalien: Ok, I've added several tests.

fridrik01 (approved):
LGTM

PR #1886, opened 2023-09-14
feat: cleanup event error cases

Stebalien:
1. Fail with LimitExceeded where it makes sense.
2. Validate that we comsume all keys/values.

### DIFF
--- a/fvm/src/kernel/default.rs
+++ b/fvm/src/kernel/default.rs
@@ -1053,12 +1053,11 @@ where
             ))?;
 
         if event_headers.len() > MAX_NR_ENTRIES {
-            return Err(syscall_error!(IllegalArgument; "event exceeded max entries: {} > {MAX_NR_ENTRIES}", event_headers.len()).into());
+            return Err(syscall_error!(LimitExceeded; "event exceeded max entries: {} > {MAX_NR_ENTRIES}", event_headers.len()).into());
         }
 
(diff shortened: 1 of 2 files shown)


Stebalien:
Implements https://github.com/filecoin-project/FIPs/pull/831

Stebalien:
Builds on #1885

fvm/src/kernel/default.rs
```diff
@@ -1079,23 +1078,19 @@ where
                         .into(),
                 );
             }
+
             if header.key_len > MAX_KEY_LEN as u32 {
                 let tmp = header.key_len;
-                return Err(syscall_error!(IllegalArgument; "event key exceeded max size: {} > {MAX_KEY_LEN}", tmp).into());
+                return Err(syscall_error!(LimitExceeded; "event key exceeded max size: {} > {MAX_KEY_LEN}", tmp).into());
```
fridrik01, line 1084: nit, but in https://github.com/filecoin-project/FIPs/pull/831 it lists that we check each entry key individually for valid UTF8  while in our impl we do it all upfront for all keys.

Stebalien: Gah, my patch reverts my previous fixes.

Stebalien: Fixed.

PR #1892, opened 2023-09-21
chore: release v4.0.0-alpha.2

Stebalien:
We're skipping alpha.1 as we used that release to mark the switch over and I like to keep the version bump at the _end_ of the release cycle (to make it clear where the release happened).


Stebalien:
So, this isn't going to pass tests until we cut at least a single v4 release.

PR #1894, opened 2023-09-21
fix: auto-create the burnt funds account

Stebalien:
The FVM relies on this so it can burn funds on self-destruct.


Stebalien:
I should have done more testing before merging everything but.... maintaining the branch dance was getting rather painful. But this should fix the last issue.

Stebalien:
This works and unborks CI, so I'm just going to merge it for now.

PR #1896, opened 2023-09-26
Remove support of v1 proofs

elmattic:
This is the counterpart of [PR 1391](https://github.com/filecoin-project/builtin-actors/pull/1391) from builtin actors repository.

### DIFF
--- a/shared/src/sector/registered_proof.rs
+++ b/shared/src/sector/registered_proof.rs
@@ -115,16 +115,6 @@ impl Default for RegisteredSealProof {
 #[derive(PartialEq, Eq, Copy, Clone, Debug, Hash)]
 #[cfg_attr(feature = "arb", derive(arbitrary::Arbitrary))]
 pub enum RegisteredPoStProof {
-    StackedDRGWinning2KiBV1,
-    StackedDRGWinning8MiBV1,
-    StackedDRGWinning512MiBV1,
-    StackedDRGWinning32GiBV1,
-    StackedDRGWinning64GiBV1,
-    StackedDRGWindow2KiBV1,
-    StackedDRGWindow8MiBV1,
-    StackedDRGWindow512MiBV1,
-    StackedDRGWindow32GiBV1,
-    StackedDRGWindow64GiBV1,
     StackedDRGWindow2KiBV1P1,
     StackedDRGWindow8MiBV1P1,
     StackedDRGWindow512MiBV1P1,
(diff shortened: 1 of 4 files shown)


arajasek (approved):
LGTM

Stebalien:
@LesnyRumcajs before I cut a release, I realize this might cause issues in Forest. Is this change fine?

shared/src/sector/registered_proof.rs
```diff
@@ -115,16 +115,6 @@ impl Default for RegisteredSealProof {
 #[derive(PartialEq, Eq, Copy, Clone, Debug, Hash)]
 #[cfg_attr(feature = "arb", derive(arbitrary::Arbitrary))]
 pub enum RegisteredPoStProof {
-    StackedDRGWinning2KiBV1,
-    StackedDRGWinning8MiBV1,
-    StackedDRGWinning512MiBV1,
-    StackedDRGWinning32GiBV1,
-    StackedDRGWinning64GiBV1,
```
Stebalien, line 122: Did we mean to remove winning post here?

shared/src/sector/registered_proof.rs
```diff
@@ -253,19 +217,19 @@ impl RegisteredSealProof {
         match self {
             Self::StackedDRG64GiBV1
             | Self::StackedDRG64GiBV1P1
-            | Self::StackedDRG64GiBV1P1_Feat_SyntheticPoRep => Ok(StackedDRGWinning64GiBV1),
+            | Self::StackedDRG64GiBV1P1_Feat_SyntheticPoRep => Ok(StackedDRGWindow64GiBV1P1),
```
Stebalien, line 220: Winning v. Window post?

shared/src/sector/registered_proof.rs
```diff
@@ -369,21 +333,11 @@ macro_rules! i64_conversion {
...
-    StackedDRGWindow2KiBV1P1 => 10,
-    StackedDRGWindow8MiBV1P1 => 11,
-    StackedDRGWindow512MiBV1P1 => 12,
-    StackedDRGWindow32GiBV1P1 => 13,
-    StackedDRGWindow64GiBV1P1 => 14,
+    StackedDRGWindow2KiBV1P1 => 0,
+    StackedDRGWindow8MiBV1P1 => 1,
+    StackedDRGWindow512MiBV1P1 => 2,
+    StackedDRGWindow32GiBV1P1 => 3,
+    StackedDRGWindow64GiBV1P1 => 4,
```
Stebalien, line 340: We should not be re-assigning numbers.

shared/src/sector/registered_proof.rs
elmattic: I don't think so.

shared/src/sector/registered_proof.rs
elmattic: Agree. This was more or less done so we can drop array size in builtin-actors repository.(`REGISTERED_POST_PROOF_VARIANTS`).

shared/src/sector/registered_proof.rs
elmattic: Should be Winning according to the doc comment.

elmattic:
@Stebalien Let me open a new PR to address your comments, ok?

shared/src/sector/registered_proof.rs
arajasek: Yeah, unfortunately we can't do that (see [here](https://github.com/filecoin-project/builtin-actors/pull/1391#discussion_r1340149993)).

shared/src/sector/registered_proof.rs
arajasek: Can we consider deleting this method entirely? I'm guessing it's used in Forest, but can we make it a helper method there? The FVM really doesn't care about winning posts.

arajasek (commented):
Sorry, dropped the ball on review here. Thanks for fixing up @elmattic !

shared/src/sector/registered_proof.rs
elmattic: I have nothing against. Let's remove it.

PR #1898, opened 2023-09-27
feat: make it possible to construct a GasDuration

Stebalien:
This lets us construct "gas charge" objects with constant times (for tracing). We primarily need this to convert between different gas duration type versions.

- Use an explicit enum with 3 options to avoid allocating/using an Arc+OnceCell when we don't need it.
- Implement Debug manually to hide implementation details.
- Change the assertion inside `GasTimer::new` to "debug only". If someone passes a "filled" duration, treat the gas timer as a no-op.

### DIFF
--- a/fvm/src/gas/timer.rs
+++ b/fvm/src/gas/timer.rs
@@ -1,3 +1,4 @@
+use std::fmt;
 // Copyright 2021-2023 Protocol Labs
 // SPDX-License-Identifier: Apache-2.0, MIT
 use std::sync::Arc;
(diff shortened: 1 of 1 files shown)


fvm/src/gas/timer.rs
```diff
@@ -53,16 +80,19 @@ impl GasTimer {
     /// Create a new timer that will update the elapsed time of a charge when it's finished.
     ///
     /// As a side effect it will establish the cell in the `GasDuration`, if it has been empty so far.
+    ///
+    /// When compiled in debug mode, passing a "filled" duration will panic.
     pub fn new(duration: &mut GasDuration) -> Self {
-        assert!(duration.get().is_none(), "GasCharge::elapsed already set!");
-
+        debug_assert!(duration.get().is_none(), "GasCharge::elapsed already set!");
```
LesnyRumcajs, line 86: Would adding some set of unit tests be worth it here?

LesnyRumcajs (approved):

PR #1901, opened 2023-09-28
amt: support `start_at` that is out of range in the first place

zhiqiangxu:
This issue is found [here](https://github.com/filecoin-project/builtin-actors/pull/1326#discussion_r1339431461), instead of panic, we can just return. 

This is useful when `start_at` is estimated which may be out of range.


Stebalien (approved):
Nice catch! Could you add a quick test case to exercise this?

Stebalien:
(I'll merge this now so I can cut a bug-fix release for you)

Stebalien:
Nvm. I added a test in #1902.

PR #1906, opened 2023-10-04
Add feature to disable code upgrades

fridrik01:
Adding feature flag to have the code upgrade syscall disabled until we officially release it


Stebalien:
I've implemented a minimal version of this in the main PR that just disables the syscall, but still compiles the code.

PR #1910, opened 2023-10-04
actor upgrade error refactor

Stebalien:
Add a general-purpose ControlFlow syscall return value and use it. This way we don't need to add an `::Abort` case to the `ExecutionError` itself.

### DIFF
--- a/fvm/src/kernel/default.rs
+++ b/fvm/src/kernel/default.rs
@@ -5,7 +5,6 @@ use std::convert::{TryFrom, TryInto};
 use std::panic::{self, UnwindSafe};
 use std::path::PathBuf;
 
-use crate::syscalls::error::Abort;
 use anyhow::{anyhow, Context as _};
 use cid::Cid;
 use filecoin_proofs_api::{self as proofs, ProverId, PublicReplicaInfo, SectorId};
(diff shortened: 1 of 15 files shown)


fvm/src/syscalls/bind.rs
```diff
@@ -49,39 +49,69 @@ pub(super) trait BindSyscall<Args, Ret, Func> {
...
-    fn into(self) -> Result<Result<Self::Value, SyscallError>, Abort> {
+    fn into_control_flow(self) -> ControlFlow<Self::Value> {
         match self {
-            Ok(value) => Ok(Ok(value)),
+            Ok(value) => ControlFlow::Return(value),
             Err(e) => match e {
-                ExecutionError::Syscall(err) => Ok(Err(err)),
-                ExecutionError::OutOfGas => Err(Abort::OutOfGas),
-                ExecutionError::Fatal(err) => Err(Abort::Fatal(err)),
-                ExecutionError::Abort(e) => Err(e),
```
fridrik01, line 84: +1!

fridrik01 (approved):
Looks great!

fvm/src/kernel/default.rs
```diff
@@ -938,27 +941,18 @@ where
...
-                    None => NO_DATA_BLOCK_ID,
-                    Some(block) => self.blocks.put_reachable(block).unwrap_or_else(|_| {
-                        log::error!("failed to write to kernel block registry");
-                        NO_DATA_BLOCK_ID
-                    }),
+            Ok(InvocationResult { exit_code, value }) => {
+                let (block_stat, block_id) = match value {
+                    None => (BlockStat { codec: 0, size: 0 }, NO_DATA_BLOCK_ID),
+                    // TODO: Check error cases. At a minimum, we could run out of gas here!
+                    Some(block) => (block.stat(), self.blocks.put_reachable(block)?),
```
fridrik01, line 948: hmm, I don't see `put_reachable` ever checking if we run out of gas

Issue #1861, opened 2023-08-30
Implement external iteration for the AMT


Stebalien:
I.e., implement a rust `Iterator` for the AMT and re-implement the `for_each` implementation in terms of this new iterator.

See #286.

Stebalien:
The remaining part here is to implement `iter_from`, iterating from an offset.

Stebalien:
Ok, we also need `iter_mut`.

PR #1912, opened 2023-10-18
[WIP] Implement external iteration for the AMT

jdjaustin:
Implement a Rust `Iterator` for the AMT and re-implement the `for_each` implementation in terms of this new iterator. Also improve documentation and rename `bitwidth` to `branching_factor`. Closes issue #1861

### DIFF
--- a/ipld/amt/src/amt.rs
+++ b/ipld/amt/src/amt.rs
@@ -1,6 +1,99 @@
 // Copyright 2021-2023 Protocol Labs
-// Copyright 2019-2022 ChainSafe Systems
+// Copyright 2019-2023 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
+//! In the context of this code, an Array Mapped Trie (AMT) is a data structure 
+//! utilizing an IPLD blockstore that solves the problem of referencing shared 
+//! data without copying an entire array. This implementation is similar to an 
+//! [IPLD vector](https://github.com/ipld/specs/blob/master/data-structures/vector.md) 
+//! but supports internal node compression and therefore sparse arrays. This is 
+//! the Rust implementation of the Go implementation documented [here](https://pkg.go.dev/github.com/filecoin-project/go-amt-ipld/v4#section-readme):
+//! "The AMT algorithm produces a tree-like graph, with a single root node addressing 
+//! a collection of child nodes which connect downward toward leaf nodes which store 
+//! the actual entries. No terminal entries are stored in intermediate elements 
+//! of the tree... We can divide up the AMT tree structure into "levels" or "heights", 
+//! where a height of zero contains the terminal elements, and the maximum height 
+//! of the tree contains the single root node. Intermediate nodes are used to span 
+//! across the range of indexes."
+//!
+//! 
+//! The maximum width for any node in the AMT structure is determined by 
+//! `2 ^ branching_factor`, meaning a node with the default branching factor of 
+//! `3` has a maximum index range of `8` and can therefore be indexed from `0` 
+//! to `(2 ^ 3) - 1 = 7`. The maximum index range for the overall structure is 
+//! determined by both the branching factor and the height of the structure; the 
+//! width of this range is `branching_factor ^ (height + 1)`. The height is specified 
+//! using a bottom-up numbering scheme, with the terminal leaves at a height of 
+//! `0` and the root node at the maximum height. Nodes can be either a `Link` or 
+//! a `Leaf` variant, which are actually a vector of links or a vector of values, 
+//! respectively. Each entry in the `Link` variant's vector may contain a CID or 
+//! a cache which holds a pointer to another `Node`; the pointer's value can only 
+//! be written once. Clearing the value of the cache and updating the CID requires 
+//! flushing, which is discussed in more detail below.
+//!
+//! An example with a single root node that is also a leaf node. This AMT has the 
+//! default branching factor, a height of `0`, and contains a single element at 
+//! index `2`.
+//! ```text
+//!                    ____________
+//!                   | root node  |  <--height 0
+//!                   |____________|
+//!                         |
+//!         | 0  | 1  | 2  | 3  | 4  | 5  | 6  | 7  |  <-- index
+//!         |None|None|Some|None|None|None|None|None|  <-- value
+//! ```
+//!
+//! A less trivial example is an AMT with a branching factor of `2` and a height 
+//! of `1`. The children nodes are leaf nodes in this example, and the leaves contain 
+//! values at indices `2`, `5`, `8`, and `15`.
+//! ```text
+//!                           ____________
+//!                          | root node  |  <-- height 1
+//!                          |____________|
+//!                                 |
+//!           __________________________________________________________________
+//!          |                     |                     |                      |
+//!    ____________           ____________          ____________           ____________
+//!   | child node |         | child node |        | child node |         | child node |  <-- height 0
+//!   |____________|         |____________|        |____________|         |____________|
+//! | 0  | 1  | 2  | 3  |  | 4  | 5  | 6  | 7  |  | 8  | 9  | 10 | 11 |  | 12 | 13 | 14 | 15 |  <-- index
+//! |None|None|Some|None|  |None|Some|None|None|  |Some|None|None|None|  |None|None|None|Some|  <-- value       
+//! ```
+//!
+//! Extending this example a bit further, let's say we create an empty AMT with 
+//! a branching factor of two using `Amt::new_with_branching_factor` and then 
+//! push a value to index `16` using `.set`. This will cause the AMT to expand 
+//! to a height of `2` with a structure as follows:
+//! ```text
+//!                           ____________
+//!                          | root node  |  <-- height 2
+//!                          |____________|
+//!                                 |
+//!           ___________________________________________________________________________________
+//!          |                     |                                      |                      |
+//!    ____________           ____________                          ____________           ____________
+//!   | child node |         | child node |                        | child node |         | child node |  <-- height 1
+//!   |____________|         |____________|                        |____________|         |____________|
+//!          |                      |                                     |                      |
+//!     ___________             _______________________              ___________            ___________
+//!    |   |   |   |           |               |   |   |            |   |   |   |          |   |   |   |
+//!    _   _   _   _      ____________         _   _   _            _   _   _   _          _   _   _   _
+//!   |_| |_| |_| |_|    | child node |       |_| |_| |_|          |_| |_| |_| |_|        |_| |_| |_| |_| <-- height 0
+//!  (indices 0 to 15)   |____________|       (indices 20 to 31)  (indices 32 to 47)      (indices 48 to 63)
+//!                     | 16 | 17 | 18 | 19 |
+//!                     |Some|None|None|None|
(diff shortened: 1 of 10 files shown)


Stebalien:
@jdjaustin let me know when you are ready for a full review.

jdjaustin:
> @jdjaustin let me know when you are ready for a full review.

@Stebalien I think this is ready for a full review.

ipld/amt/src/iter.rs
```diff
@@ -0,0 +1,491 @@
...
+    /// let results: Vec<_> = results
+    ///     .into_iter()
+    ///     .map(|(k, v)| (k.clone(), v.clone()))
+    ///     .collect();
+    ///
+    /// assert_eq!(kvs, results);
+    ///
+    /// # anyhow::Ok(())
+    /// ```
+    pub fn iter_from(&self, key: usize) -> Vec<Result<(usize, &V), crate::Error>> {
```
Stebalien, line 153: This should return an iterator.

ipld/amt/src/iter.rs
```diff
@@ -0,0 +1,491 @@
...
+        let mut results = Vec::new();
+        for res in self.iter() {
+            let (k, v) = res.expect("Failed to generate iterator from AMT");
+            if k >= key {
```
Stebalien, line 157: This needs to seek to the correct position without iterating (see `for_each_ranged`).

ipld/amt/src/iter.rs
```diff
@@ -0,0 +1,491 @@
...
+    ///     .collect();
+    ///
+    /// assert_eq!(kvs, results);
+    ///
+    /// # anyhow::Ok(())
+    /// ```
+    pub fn iter_from(&self, key: usize) -> Vec<Result<(usize, &V), crate::Error>> {
+        let mut results = Vec::new();
+        for res in self.iter() {
+            let (k, v) = res.expect("Failed to generate iterator from AMT");
```
Stebalien, line 156: ?

Stebalien (commented):
I'll take a look at the rest later, but this section doesn't look quite done.

ipld/amt/src/iter.rs
```diff
@@ -0,0 +1,559 @@
...
+                                        .map(Box::new)
+                                }) {
+                                    Ok(node) => {
+                                        stack.idx += 1;
+                                        self.stack.push(IterStack {
+                                            node: Some(node.as_ref()),
+                                            idx: 0,
+                                        });
+                                    }
+                                    Err(e) => return Some(Err(e)),
```
Stebalien, line 302: Ideally we'd increment the index, even on error (e.g., so we can skip over missing sections).

ipld/amt/src/iter.rs
```diff
@@ -0,0 +1,559 @@
...
+                            Some(Link::Dirty(node)) => {
+                                stack.idx += 1;
+                                self.stack.push(IterStack {
+                                    node: Some(node.as_ref()),
+                                    idx: 0,
+                                });
+                            }
+                            None => {
+                                stack.idx += 1;
+                                self.key += 2_usize.pow(self.bit_width);
```
Stebalien, line 314: This depends on the height. You need to call `nodes_for_height` with the `bit_width` and `root.height - self.stack.len()` (plus or minus 1-2, you'll need to test this carefully).

ipld/amt/src/iter.rs
```diff
@@ -0,0 +1,559 @@
...
+        self.iter()
+    }
+}
+
+pub struct Iter<'a, V, BS, Ver> {
+    stack: Vec<IterStack<'a, V>>,
+    blockstore: BS,
+    bit_width: u32,
+    ver: PhantomData<Ver>,
+    key: usize,
```
Stebalien, line 249: ditto, u64

ipld/amt/src/iter.rs
```diff
@@ -0,0 +1,559 @@
...
+}
+
+pub struct IterStack<'a, V> {
+    pub(crate) node: Option<&'a Node<V>>,
```
Stebalien, line 253: Does this need to be an `Option`?

ipld/amt/src/iter.rs
```diff
@@ -0,0 +1,559 @@
...
+    ///     .collect();
+    ///
+    /// assert_eq!(kvs, results);
+    ///
+    /// # anyhow::Ok(())
+    /// ```
+    pub fn iter_from(&self, key: usize) -> Result<Iter<'_, V, &BS, Ver>, crate::Error> {
+        let mut iter = self.iter();
+        while key > iter.key {
+            let stack = iter.stack.last_mut().expect("Stack is empty");
```
Stebalien, line 158: what if the AMT is empty?

ipld/amt/src/iter.rs
```diff
@@ -0,0 +1,559 @@
...
+    /// let results: Vec<_> = results
+    ///     .into_iter()
+    ///     .map(|(k, v)| (k.clone(), v.clone()))
+    ///     .collect();
+    ///
+    /// assert_eq!(kvs, results);
+    ///
+    /// # anyhow::Ok(())
+    /// ```
+    pub fn iter_from(&self, key: usize) -> Result<Iter<'_, V, &BS, Ver>, crate::Error> {
```
Stebalien, line 155: We should be able to seek to the exact offset we want in logn time. We need to implement similar logic to `get` to skip entire sub-trees. I can also take a crack at implementing this if you don't feel comfortable with it.

Stebalien (changes requested):
❤️

ipld/amt/src/amt.rs
```diff
@@ -50,6 +143,7 @@ pub struct AmtImpl<V, BS, Ver> {
 /// // Generate cid by calling flush to remove cache
 /// let cid = amt.flush().unwrap();
 /// ```
+// TODO(jdjaustin): in another PR: remove or refactor version typestate, and refactor `AmtImpl`
```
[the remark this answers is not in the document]
jdjaustin: IIRC this comment was left by @aatifsyed so although I don't recall the reasoning here, maybe he can elaborate.

ipld/amt/src/lib.rs
```diff
@@ -10,14 +10,17 @@
 mod amt;
 mod diff;
 mod error;
+mod iter;
 mod node;
 mod root;
 mod value_mut;
 
-pub use self::amt::{Amt, Amtv0};
+pub use self::amt::{Amt, AmtImpl, Amtv0};
```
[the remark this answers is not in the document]
jdjaustin: No. Fixed.

ipld/amt/src/iter.rs
Stebalien: Ah, that's not quite sparse enough. Try changing 64 to 500.

Basically, the current logic works for 2 levels, but not more.

ipld/amt/src/iter.rs
```diff
@@ -0,0 +1,553 @@
...
+    /// let mut amt = Amt::new(store);
+    /// amt.set(2, "two".to_owned());
+    /// amt.set(1, "one".to_owned());
+    ///
+    /// let vals = amt.values().collect::<Vec<_>>();
+    /// assert_eq!(vals, ["one", "two"]);
+    ///
+    /// # anyhow::Ok(())
+    /// ```
+    pub fn values(&self) -> impl Iterator<Item = &V> {
```
Stebalien, line 76: I'd drop both the keys/values methods. They're not all _that_ useful and they'll panic on error.

ipld/amt/src/iter.rs
```diff
@@ -0,0 +1,553 @@
...
+    stack: Vec<IterStack<'a, V>>,
+    blockstore: BS,
+    bit_width: u32,
+    ver: PhantomData<Ver>,
+    key: u64,
+}
+
+pub struct IterStack<'a, V> {
+    pub(crate) node: Option<&'a Node<V>>,
+    pub(crate) idx: u64,
```
Stebalien, line 254: note: _this_ can be a usize.

ipld/amt/src/iter.rs
jdjaustin: Good catch. Fixed.

ipld/amt/src/iter.rs
jdjaustin: I can make another pass at it, but to simplify things, how about we address it in another PR since this one is quite large already?

Stebalien: In that case, I'd remove it from this PR entirely. The implementation will likely be significantly different anyways.

jdjaustin: Sounds good, I'll open a tracking issue.

jdjaustin: #1923

ipld/amt/src/iter.rs
jdjaustin: #1923

ipld/amt/src/iter.rs
```diff
@@ -0,0 +1,559 @@
...
+    fn next(&mut self) -> Option<Self::Item> {
+        loop {
+            let stack = self.stack.last_mut()?;
+            match stack.node {
+                Some(Node::Leaf { vals }) => {
+                    while stack.idx < vals.len() {
+                        match vals[stack.idx] {
+                            Some(ref v) => {
+                                stack.idx += 1;
+                                self.key += 1;
```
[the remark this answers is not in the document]
jdjaustin: You are correct. Added a test for this and the fix.

ipld/amt/src/iter.rs
jdjaustin: At some (pun intended) point I think it did, but with changes since then, it doesn't need to be any more. Fixed.

ipld/amt/src/amt.rs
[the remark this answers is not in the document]
jdjaustin: I'll create another issue for discussing the TODOs so as to not block this PR.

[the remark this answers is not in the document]
jdjaustin: #1925

ipld/amt/src/iter.rs
```diff
@@ -0,0 +1,364 @@
...
+}
+
+impl<'a, V, BS, Ver> Iterator for Iter<'a, V, BS, Ver>
+where
+    BS: Blockstore,
+    V: Serialize + DeserializeOwned,
+{
+    type Item = Result<(u64, &'a V), crate::Error>;
+    fn next(&mut self) -> Option<Self::Item> {
+        let root_height = self.stack[0].height;
```
Stebalien, line 92: Won't this panic if we call `next` after exhausting the iterator? Technically that's allowed by spec, but it's not very nice.

I'd either:

1. Store the height in `Iter` _only_, always calculating the current height as `self.height - self.stack.len()`.
2. Compute the next height as `prev_stack.height - 1`.

ipld/amt/src/iter.rs
```diff
@@ -0,0 +1,364 @@
...
+        loop {
+            if self.key > MAX_INDEX {
+                return None;
+            }
```
Stebalien, line 96:
```rust
        while self.key <= MAX_INDEX {
            ...
        }
        None
```

(not necessary, just slightly easier to read).

Stebalien (commented):
I think everything else is correct but I'll make one final pass before merging.

PR #1914, opened 2023-10-20
Add syscall for BLS aggregate signatures

DrPeterVanNostrand:
This PR reopens the previously closed PR #1836.


fvm/src/kernel/default.rs
```diff
@@ -511,14 +511,26 @@ where
...
             .iter()
-            .map(|len| {
-                let (plaintext, rem) = plaintexts_concat.split_at(*len as usize);
-                plaintexts_concat = rem;
-                plaintext
+            .map(|&len| {
+                let start = offset;
+                offset = start
+                    .checked_add(len as usize)
+                    .context("invalid bls signature length")
```
DrPeterVanNostrand, line 521: Should this be `invalid bls plaintext length`?

fvm/src/gas/price_list.rs
```diff
@@ -102,19 +101,14 @@ lazy_static! {
...
-                    scale: Gas::new(26),
-                },
-            }
-        },
         secp256k1_recover_cost: Gas::new(1637292),
+        bls_pairing_cost: Gas::new(8299302),
+        bls_hashing_cost: ScalingCost {
+            flat: Gas::zero(),
+            // TODO: Can we just use the "sha256" number below?
+            scale: Gas::new(7),
```
DrPeterVanNostrand, line 109: Just a note: this PR did not change the BLS signature hash function (that maps a plaintext's bytes to a curve point). I'm guessing that this gas price change (to that of SHA256's) was intentional as SHA256 is used internally by the BLS hash function.

Stebalien: Yeah, that makes sense. I'll leave it as its own separate price just to be safe.

fvm/src/kernel/default.rs
Stebalien: Yep!

Stebalien:
Merged in #2003

PR #1919, opened 2023-11-01
feat: pretty-print addresses when debug-formatting

Stebalien:
We currently just print the protocol and bytes, but that's not very human friendly.

Now:

```rust
println!("{:?}", Address::new_id(1));
```

Prints:

> `Address("f01")`


raulk (approved):
A unit test would be awesome to prevent regressions!

Stebalien:
> A unit test would be awesome to prevent regressions!

I thought I could race you on the review... apparently not.

Issue #1787, opened 2023-06-12
Pluggable System Calls & Kernel


Stebalien:
Right now, the FVM provides a single system call API that's difficult to extend. Use-cases include:

1. Other blockchains (e.g., for IPC). Said chains may have their own syscalls and definitely don't want some of our special syscalls (proof validation, etc.).
2. Development tools. E.g., this would allow users to easily add special debugging syscalls, compile the FVM without proof validation support, etc.

Proposal:

1. Make the `syscalls::bind_syscalls` function pluggable. E.g., consider making it a possibly static method on the kernel?
2. Make the `Kernel` extensible. I.e., define a _base_ kernel that must be implemented (along with a basic `bind_syscalls` function) and let the user define custom "mixins".

yourarj:
@Stebalien Would like to discuss the approach and impl.

Stebalien:
impl Kernel for ... where Kernel
The "naive" approach is to:

1. Add some form of `bind_syscalls` method to the kernel.
2. Let the user "wrap" the default kernel in a custom kernel that extends `bind_syscalls`.

However, this has a few problems:

1. We need the price list to be extensible.
2. The kernel doesn't expose the call manager. I guess we could allow this?
3. Really, the kernel only exposes methods needed by syscalls, by design.

We'll probably need to:

1. Expose all the kernel internals. IMO, this is probably fine? Although we may want to mark the `DefaultKernel` struct as non-exhaustive.
2. Make the price list pluggable. I think the right answer here is to make the `PriceList` an associated type of the `Machine`.
3. For wrapping, we could use something like [ambassador](https://crates.io/crates/ambassador), but that's still not great. I think it may be better to use newtypes. I.e.:

```rust
pub trait IpldBlockOps {
    fn block_open(&mut self, cid: &Cid) -> Result<(BlockId, BlockStat)>;
    // ...
}
pub trait BaseKernel {
    type IpldBlockOps: IpldBlockOps;
    type CryptoOps: CryptoOps;

    pub fn ipld(&mut self) -> &mut Self::IpldBlockOps;
    pub fn crypto(&mut self) -> &mut Self::CryptoOps;
    // ...
}

pub trait MyKernelExtension: BaseKernel {
    type CustomOps: CustomOps;

    pub fn custom(&mut self) -> &mut Self::CustomOps;
}
```

This way, the user can override specific features/modules without having to wrap the entire kernel trait.

Stebalien:
[...]

PR #1922, opened 2023-11-15
Add initial support for pluggable syscalls

fridrik01:
See: https://github.com/filecoin-project/ref-fvm/issues/1787

This PR adds initial support for pluggable syscall by allowing us to extend the default kernel and add new custom syscalls through a new kernel.

The implementation does this by moving the global `bind_syscall` function into kernel via a `SyscallHandler` trait where each kernel needs to add their new syscalls. 

The implementation uses [Ambassador](https://crates.io/crates/ambassador) crate which automatically create stubs for delegating kernel implementations we are not overriding in new kernels, minimizing the code we need to write.

### DIFF
--- a/fvm/src/kernel/default.rs
+++ b/fvm/src/kernel/default.rs
@@ -6,23 +6,17 @@ use std::path::PathBuf;
 
 use anyhow::{anyhow, Context as _};
 use cid::Cid;
-use filecoin_proofs_api::{self as proofs, ProverId, SectorId};
 use fvm_ipld_blockstore::Blockstore;
-use fvm_ipld_encoding::{bytes_32, IPLD_RAW};
+use fvm_ipld_encoding::IPLD_RAW;
 use fvm_shared::address::Payload;
-use fvm_shared::consensus::ConsensusFault;
 use fvm_shared::crypto::signature;
 use fvm_shared::econ::TokenAmount;
 use fvm_shared::error::ErrorNumber;
 use fvm_shared::event::{ActorEvent, Entry, Flags};
-use fvm_shared::piece::{zero_piece_commitment, PaddedPieceSize};
 use fvm_shared::sys::out::vm::ContextFlags;
 use fvm_shared::upgrade::UpgradeInfo;
-use fvm_shared::{commcid, ActorID};
-use lazy_static::lazy_static;
+use fvm_shared::ActorID;
 use multihash::MultihashDigest;
-use rayon::iter::{IndexedParallelIterator, IntoParallelRefIterator, ParallelIterator};
-use rayon::prelude::ParallelDrainRange;
 
 use super::blocks::{Block, BlockRegistry};
 use super::error::Result;
@@ -32,17 +26,12 @@ use crate::call_manager::{
     CallManager, Entrypoint, InvocationResult, INVOKE_FUNC_NAME, NO_DATA_BLOCK_ID,
     UPGRADE_FUNC_NAME,
 };
-use crate::externs::{Chain, Consensus, Rand};
+use crate::externs::{Chain, Rand};
 use crate::gas::GasTimer;
 use crate::init_actor::INIT_ACTOR_ID;
 use crate::machine::{MachineContext, NetworkConfig, BURNT_FUNDS_ACTOR_ID};
 use crate::state_tree::ActorState;
 use crate::{ipld, syscall_error};
-
-lazy_static! {
-    static ref NUM_CPUS: usize = num_cpus::get();
-    static ref INITIAL_RESERVE_BALANCE: TokenAmount = TokenAmount::from_whole(300_000_000);
-}
 
 const BLAKE2B_256: u64 = 0xb220;
 const ENV_ARTIFACT_DIR: &str = "FVM_STORE_ARTIFACT_DIR";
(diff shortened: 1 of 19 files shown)


fvm/src/kernel/mod.rs
```diff
@@ -115,6 +123,13 @@ pub trait Kernel:
     ) -> Result<CallResult>;
 }
 
+pub trait SyscallHandler {
+    fn bind_syscalls(
+        &self,
+        linker: &mut Linker<InvocationData<impl Kernel + 'static>>,
```
Stebalien, line 129: Ok, so, I _think_ the solution here may be:

```rust
pub trait SyscallHandler<K: Kernel> {
    fn bind_syscalls(
        &self,
        linker: &mut Linker<InvocationData<K>>,
    ) -> anyhow::Result<()>;
}
```

You'd implement:

```rust
impl<K: Kernel> SyscallHandler<K> for BaseKernel {
    fn bind_syscalls(
        &self,
        linker: &mut Linker<InvocationData<K>>,
    ) -> anyhow::Result<()> { .. }
}

impl SyscallHandler<FilecoinKernel> for FilecoinKernel {
    fn bind_syscalls(
        &self,
        linker: &mut Linker<InvocationData<FilecoinKernel>>,
    ) -> anyhow::Result<()> { .. }
}
```

(maybe...)

fridrik01: I think in order to get this to work, I must first get the ambassador stuff to work. Unfortunately, it does not seem to support traits that extend other traits (like Kernel) so I need to consolidate everything into a single Kernel trait; will see if that works...

Stebalien: You'd need to use ambassador on each trait.

fridrik01: yebb that was the issue, still working on getting it to work but getting closer I hope

fvm/src/syscalls/mod.rs
```diff
@@ -233,107 +236,204 @@ fn min_table_elements(module: &Module) -> Option<u32> {
...
-        // accidentally break it.
-        linker.bind("actor", "upgrade_actor", actor::upgrade_actor)?;
+impl<C> SyscallHandler<DefaultKernel<C>> for DefaultKernel<C>
+where
+    C: CallManager,
+{
+    fn bind_syscalls(
+        &self,
+        //linker: &mut Linker<InvocationData<impl Kernel + 'static>>,
+        linker: &mut wasmtime::Linker<InvocationData<DefaultKernel<C>>>,
```
Stebalien, line 246:
```suggestion
impl<K> SyscallHandler<K> for DefaultKernel<K::CallManager>
where
    K: Kernel,
{
    fn bind_syscalls(
        &self,
        linker: &mut wasmtime::Linker<InvocationData<K>>,
```

fridrik01: Awesome, this worked!!

fvm/src/kernel/default.rs
```diff
@@ -57,37 +44,26 @@ const TEST_ACTOR_ALLOWED_TO_CALL_CREATE_ACTOR: ActorID = 98;
 pub struct DefaultKernel<C> {
```
Stebalien, line 44: I think we should consider renaming this to the "base kernel". But that can be in a subsequent PR to avoid bloating this one.

fvm/src/kernel/default.rs
```diff
@@ -57,37 +44,26 @@ const TEST_ACTOR_ALLOWED_TO_CALL_CREATE_ACTOR: ActorID = 98;
...
-    caller: ActorID,
-    actor_id: ActorID,
-    method: MethodNum,
-    value_received: TokenAmount,
-    read_only: bool,
+    pub caller: ActorID,
+    pub actor_id: ActorID,
+    pub method: MethodNum,
+    pub value_received: TokenAmount,
+    pub read_only: bool,
```
Stebalien, line 51: I'm starting to wonder how this will interact with testing. E.g., I want to support `FilecoinKernel<TestKernel<DefaultKernel>>`. Options include:

1. Using accessor methods here. IIRC, we never actually _change_ any of these fields (or the ones below), we just read them.
2. Require that all kernels "deref" to some `BaseKernel`.
3. Alternatively, require that all kernels expose some form of "invocation state"?

I think we can leave it alone for now, but I figured I'd comment on this in case you had any ideas.

fvm/src/kernel/mod.rs
```diff
@@ -68,32 +98,13 @@ pub trait Kernel:
 {
     /// The [`Kernel`]'s [`CallManager`] is
     type CallManager: CallManager;
+    type Kernel: Kernel<CallManager = Self::CallManager>;
```
Stebalien, line 101: The kernel shouldn't have a kernel...

fvm/src/kernel/mod.rs
```diff
@@ -104,7 +115,7 @@ pub trait Kernel:
     /// kernel specifying its Self.
     /// This method is part of the Kernel trait so it can refer to the Self::CallManager
     /// associated type necessary to constrain K.
-    fn send<K: Kernel<CallManager = Self::CallManager>>(
+    fn send<K: Kernel<CallManager = <Self as Kernel>::CallManager>>(
```
Stebalien, line 118: Do we _need_ this?

fvm/src/kernel/mod.rs
```diff
@@ -229,7 +241,15 @@ pub trait ActorOps {
     fn balance_of(&self, actor_id: ActorID) -> Result<TokenAmount>;
 }
 
+// we are not delegating this trait as ambassador seems to not support feature flags
```
Stebalien, line 244: You know.... can we just get rid of the feature flag? The same way we did for "upgrades"? I.e.:

1. Use the feature flag to decide whether or not to _bind_ the syscall.
2. Don't put this method behind a feature flag.

fvm/src/kernel/default.rs
```diff
@@ -876,11 +743,7 @@ where
             .create_actor(code_id, actor_id, delegated_address)
     }
 
-    fn upgrade_actor<K: Kernel>(
-        &mut self,
-        new_code_cid: Cid,
-        params_id: BlockId,
-    ) -> Result<CallResult> {
+    fn upgrade_actor(&mut self, new_code_cid: Cid, params_id: BlockId) -> Result<CallResult> {
```
Stebalien, line 746: Ah, I'd keep this like this and pass `K` to `call_actor`. It makes testing easier.

fvm/src/kernel/mod.rs
```diff
@@ -21,38 +21,68 @@ use fvm_shared::sys::out::vm::MessageContext;
...
 ///
 /// - Manages the Actor's state.
 /// - Tracks and charges for IPLD & syscall-specific gas.
 ///
 /// Actors may call into the kernel via the syscalls defined in the [`syscalls`][crate::syscalls]
 /// module.
+#[delegatable_trait]
 pub trait Kernel:
-    ActorOps
+    SyscallHandler<<Self as Kernel>::Kernel>
```
Stebalien, line 82: Do we need this? IMO:

1. This really should be the other way around (kernels are syscall handlers).
2. I'm not sure if we even need to care about that either. We can likely treat them as different things.


So, I think the `SyscallHandler` implementation changes to:

```rust
impl<K, T> SyscallHandler<K> for DefaultFilecoinKernel<T>
where
    K: Kernel,
    T: SyscallHandler<Self>
```

fvm/src/kernel/mod.rs
```diff
@@ -21,38 +21,68 @@ use fvm_shared::sys::out::vm::MessageContext;
...
 use crate::machine::Machine;
+use crate::syscalls::InvocationData;
 
 pub struct CallResult {
     pub block_id: BlockId,
     pub block_stat: BlockStat,
     pub exit_code: ExitCode,
 }
 
+pub trait ConstructKernel<C> {
```
Stebalien, line 51: So... can we:

1. Leave these on `Kernel`.
2. Not implement the delegate logic on the top-level kernel.

Or does that not make sense?

fvm/src/kernel/default.rs
fridrik01: Done, but moved to the Kernel trait due to ambassador errors which I could not get working

fvm/src/kernel/mod.rs
fridrik01: makes sense!

fvm/src/kernel/mod.rs
fridrik01: Are you suggesting removing SyscallHandler from the Kernel? How would we call the `bind_syscalls` on it when initializing the engine then, or am I misunderstanding?

Stebalien: I was... but yeah, that might be even more annoying (we'd have to change all `: Kernel` constraints to `: SyscallHandler` or `: Kernel + SyscallHandler`. So, maybe later but not now.

Stebalien (approved):
We're going to want to re-think the traits here a bit, but I think this is at a nice stopping point where we can merge and iterate on master.

fvm/src/kernel/default.rs
```diff
@@ -185,6 +172,100 @@ where
             },
         })
     }
+
+    fn upgrade_actor<K: Kernel<CallManager = C>>(
```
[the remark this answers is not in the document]
Stebalien: (or we can just leave it, we're going to have to refactor all this anyways)

fvm/src/kernel/default.rs
fridrik01: See https://github.com/filecoin-project/ref-fvm/pull/1930

PR #1927, opened 2023-11-16
test: exclude test logic from coverage reports

Stebalien:
That way, we actually get useful information.


Stebalien:
I kind of doubt this'll work... but maybe?

PR #1928, opened 2023-11-17
fix test failure when running cargo test

fridrik01:
Running `cargo test` seems to non-deterministically select different networks causing the Address test_debug() test to fail


Stebalien:
See https://github.com/filecoin-project/ref-fvm/pull/1926. We can also fix it this way, but I'd kind of prefer to just isolate the problematic test.

fridrik01:
> See #1926. We can also fix it this way, but I'd kind of prefer to just isolate the problematic test.

That seems reasonable, but if someone calls `set_current_network` from any other future tests then we will have a flaky test on our hands.

Stebalien:
Yeah, fair enough.

Stebalien:
Ah, well. This one is still a bit racy (rust runs tests in parallel). We'll need to assert that the output is either `f01` or `t01`.

Stebalien (approved):

PR #1930, opened 2023-11-22
Rename DefaultKernel to BaseKernel

fridrik01:
After landing https://github.com/filecoin-project/ref-fvm/pull/1922  which enabled initial support for pluggable syscalls we need a better name for `DefaultKernel`, renaming here to `BaseKernel` as was suggested in https://github.com/filecoin-project/ref-fvm/pull/1922#discussion_r1399626766


Stebalien:
Ok, so, I'm not sure if we want to do this. Let's punt for now.

PR #1931, opened 2023-11-29
fix: set the default thread-pool size by the available concurrency

Stebalien:
And default to 8 if we can't get this information. This isn't a perfect metric (e.g., it doesn't take IO into account) but it's better than nothing).

### DIFF
--- a/fvm/src/executor/threaded.rs
+++ b/fvm/src/executor/threaded.rs
@@ -9,10 +9,10 @@ use super::{ApplyKind, ApplyRet, Executor};
 
 lazy_static! {
     static ref EXEC_POOL: yastl::Pool = yastl::Pool::with_config(
-        8,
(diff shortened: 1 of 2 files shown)


fvm/src/executor/threaded.rs
```diff
@@ -9,10 +9,10 @@ use super::{ApplyKind, ApplyRet, Executor};
 
 lazy_static! {
     static ref EXEC_POOL: yastl::Pool = yastl::Pool::with_config(
-        8,
+        std::thread::available_parallelism().map(|n|n.get()).unwrap_or(8),
```
fridrik01, line 12: we should probably mention this change in `fvm/CHANGELOG.md`

fridrik01 (approved):
lgtm,

PR #1933, opened 2023-12-01
[WIP] Kernel Refactor

Stebalien:
This is an attempt at refactoring the kernel, implementing a bunch of things I've been thinking about for a while. I'll probably need to break it into multiple PRs and/or continue to work on it, but it's good enough right now for some basic feedback.

This PR:

1. Reworks the layering to look more like https://github.com/filecoin-project/FIPs/discussions/845.
2. Improves the ability to "delegate" the implementation.
3. Moves the circulating supply to the Filecoin kernel (could be a separate PR).
4. Makes `SyscallHandler::bind_syscalls` a _static_ method. IMO, this is "more correct" as we memoize the result of calling this method so the kernel on which we called `bind_syscalls` may not (usually, _will_ not) be the kernel on which we actually _use_ those syscalls.
5. Makes syscalls explicitly declare the "kernel ops" they need. E.g., `K: DebugOps` for the `debug` syscalls.
6. Makes the `SyscallHandler` an associated type on the `Kernel`. This makes it a bit easier to wrap/alter the kernel without touching the syscall bindings.

### DIFF
--- a/fvm/src/kernel/default.rs
+++ b/fvm/src/kernel/default.rs
@@ -31,6 +31,7 @@ use crate::gas::GasTimer;
 use crate::init_actor::INIT_ACTOR_ID;
 use crate::machine::{MachineContext, NetworkConfig, BURNT_FUNDS_ACTOR_ID};
 use crate::state_tree::ActorState;
+use crate::syscalls::DefaultSyscallHandler;
 use crate::{ipld, syscall_error};
 
 const BLAKE2B_256: u64 = 0xb220;
@@ -67,13 +68,8 @@ where
     C: CallManager,
 {
     type CallManager = C;
-
-    fn into_inner(self) -> (Self::CallManager, BlockRegistry)
-    where
-        Self: Sized,
-    {
-        (self.call_manager, self.blocks)
-    }
+    type Limiter = <<C as CallManager>::Machine as Machine>::Limiter;
+    type SyscallHandler = DefaultSyscallHandler;
 
     fn new(
         mgr: C,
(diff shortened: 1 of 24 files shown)


fvm/src/kernel/mod.rs
```diff
@@ -48,36 +48,12 @@ pub struct CallResult {
...
-    + MessageOps
-    + NetworkOps
-    + RandomnessOps
-    + SelfOps
-    + LimiterOps
-    + 'static
-{
-    /// The [`Kernel`]'s [`CallManager`] is
+/// The base [`Kernel`] trait implemented by all kernels.
+pub trait Kernel: SyscallHandler + 'static {
```
Stebalien, line 52: I've also tried:

1. Inverting this to `SyscallHandler: Kernel`. I like that in theory, but it's strictly uglier in practice.
2. `SyscallHandler + Kernel` everywhere because they're _kind_ of orthogonal and only happen to be implemented on the same type. This actually helps a bit with some circular type issues so I may have to go back to it.

I haven't yet tried merging them into a single trait but.... then I'd need to define `Kernel` as `Kernel<K: Kernel>`, and I'd rather not do that.

I also haven't tried _completely_ separating them and treating them as entirely separate types. I might try doing that next to see if it helps with some of the type issues.

fvm/src/kernel/default.rs
```diff
@@ -95,176 +89,35 @@ where
...
-                exit_code,
-                value: None,
-            } => CallResult {
-                block_id: NO_DATA_BLOCK_ID,
-                block_stat: BlockStat { codec: 0, size: 0 },
-                exit_code,
-            },
-        })
+    fn price_list(&self) -> &PriceList {
+        self.call_manager.price_list()
```
Stebalien, line 97: I really want to remove some of these methods from the base "kernel".

Stebalien:
@fridrik01 I'm asking for "light feedback", not an actual review.

fvm/src/kernel/mod.rs
Stebalien: Ok, I tried separating them and that's worse. I've moved it to an associated type and that seems like the best way forwards.

testing/conformance/src/vm.rs
```diff
@@ -234,137 +229,85 @@ where
...
     }
 }
 
-impl<M, C, K> IpldBlockOps for TestKernel<K>
+impl<M, C, K> CallOps<Self> for TestKernel<K>
 where
     M: Machine,
     C: CallManager<Machine = TestMachine<M>>,
-    K: Kernel<CallManager = C>,
+    K: Kernel<CallManager = C> + CallOps<Self>,
```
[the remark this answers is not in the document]
Stebalien: So, nothing worked except to make this code non-generic. Honestly, there's no reason for the test kernel to be generic over the kernel, so that's probably fine?

Stebalien:
Review first:

- https://github.com/filecoin-project/ref-fvm/pull/1935
- https://github.com/filecoin-project/ref-fvm/pull/1936
- https://github.com/filecoin-project/ref-fvm/pull/1938
- https://github.com/filecoin-project/ref-fvm/pull/1939

Stebalien:
Everything here that's actually useful has been merged.

PR #1936, opened 2023-12-06
feat: move total_fil_circ_supply to the FilecoinKernel

Stebalien:
It's filecoin specific.


fvm/src/syscalls/filecoin.rs
```diff
@@ -178,3 +178,15 @@ pub fn batch_verify_seals(
     }
     Ok(())
 }
+
+/// Returns the network circ supply split as two u64 ordered in little endian.
```
arajasek, line 182: You could drop the mention of "splitting into two u64"s here, that's obvious from the `sys::TokenAmount` type.

fvm/src/kernel/filecoin.rs
```diff
@@ -226,6 +235,13 @@ where
             verify_replica_update(replica)
         }))
     }
+
+    fn total_fil_circ_supply(&self) -> Result<TokenAmount> {
+        // From v15 and onwards, Filecoin mainnet was fixed to use a static circ supply per epoch.
```
arajasek, line 240: Drop this comment? Do we really care about the fix in nv15?

arajasek (approved):
Approved with comment cleanup nits.

PR #1939, opened 2023-12-06
fix: make delegation simpler by introducing a prelude module

Stebalien:
Users can now glob-import the kernel's "prelude" module instead of, e.g., glob-importing the entire `kernel` and the top-level `fvm` crate. Ideally ambassador would handle importing all of its internal types, but it doesn't.


maciejwitowski (approved):
LGTM

PR #1941, opened 2023-12-07
refactor: make partial kernels possible

Stebalien:
This change has two parts:

1. Remove most (all except gas) "ops" from the base kernel trait, allowing kernels to implement only some operations.
2. Make syscalls declare the ops they actually need instead of depending on the entire kernel.

This should make it possible to implement custom kernels that expose fewer operations.

It also removes the `LimiterOps` trait as that's not really an "op".

### DIFF
--- a/fvm/src/kernel/default.rs
+++ b/fvm/src/kernel/default.rs
@@ -65,6 +65,7 @@ where
     C: CallManager,
 {
     type CallManager = C;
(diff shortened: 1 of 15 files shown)


maciejwitowski (approved):

maciejwitowski (approved):
These changes are all compiler-checked so I'm happy to approve them.

PR #1943, opened 2023-12-07
refactor: simplify the DefaultFilecoinKernel generics

Stebalien:
It's not _actually_ generic over the inner kernel type (only works for
the `DefaultKernel`), so we might as well simplify that.

I.e., was:

    DefaultFilecoinKernel<DefaultKernel<impl CallManager>>

Now:

    DefaultFilecoinKernel<impl CallManager>


Stebalien:
I'm just going to merge this one.

PR #1944, opened 2023-12-07
fix: bind syscalls in conformance tests

Stebalien:
Unfortunately, I had to drop the generics to prevent some circular type constraint evaluation issues. This is likely because:

1. `SyscallHandler` is generic over "self".
2. `SyscallHandler<Self>` is only implemented if `Self: FilecoinKernel`.
3. `FilecoinKernel` is only implemented if `Self: SyscallHandler<Self>`.

Rust _can_ figure all of this out if we specify concrete types, but it can't yet prove this generically.

NOTE: Ideally tests would catch this, but we don't yet have conformance tests for v4.

### DIFF
--- a/fvm/src/syscalls/mod.rs
+++ b/fvm/src/syscalls/mod.rs
@@ -4,9 +4,9 @@ use anyhow::{anyhow, Context as _};
 use num_traits::Zero;
 use wasmtime::{AsContextMut, ExternType, Global, Linker, Memory, Module, Val};
 
-use crate::call_manager::{backtrace, CallManager};
+use crate::call_manager::backtrace;
 use crate::gas::{Gas, GasInstant, GasTimer};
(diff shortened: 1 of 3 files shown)


fvm/src/syscalls/mod.rs
```diff
@@ -329,14 +329,22 @@ where
...
+        + ActorOps
+        + IpldBlockOps
+        + CryptoOps
+        + DebugOps
+        + EventOps
+        + GasOps
+        + MessageOps
+        + NetworkOps
+        + RandomnessOps
+        + SelfOps,
```
fridrik01, line 344: This made no sense until I noticed  https://github.com/filecoin-project/ref-fvm/pull/1941 :smile:

fridrik01 (approved):
LGTM

PR #1947, opened 2023-12-12
feat: cleanup & document the context API

Stebalien:
We're now exposing this to users so we need to clean this up a bit:

1. Document it.
2. Remove `Memory::read_cbor` from the public API, it's not safe (time-wise) to call on untrusted memory.

### DIFF
--- a/fvm/src/syscalls/context.rs
+++ b/fvm/src/syscalls/context.rs
@@ -2,23 +2,29 @@
 // SPDX-License-Identifier: Apache-2.0, MIT
 use std::io::Cursor;
 use std::ops::{Deref, DerefMut};
-use std::panic;
 
 use cid::Cid;
-use fvm_ipld_encoding::from_slice;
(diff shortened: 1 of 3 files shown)


fvm/src/syscalls/context.rs
```diff
@@ -87,6 +109,11 @@ impl Memory {
         .context("failed to parse cid")
     }
 
+    /// Write a CID to actor memory at the given offset.
+    ///
+    /// If the CID's length exceeds the specified length, this function will with
+    /// [`ErrorNumber::BufferTooSmall`]. For all other failures (e.g., memory out of bounds errors),
```
fridrik01, line 115:
```suggestion
    /// If the CID's length exceeds the specified length, this method returns an
    /// [`ErrorNumber::BufferTooSmall`] error. For all other failures (e.g., memory out of bounds errors),
```

fridrik01 (approved):
LGTM

PR #1951, opened 2023-12-14
refactor: remove GasOps from the Kernel

Stebalien:
It's not really a proper "ops" trait, it's more "basic functionality".

Additionally, the concept of a global price list is going to go away as we make the kernel more pluggable.

1. Move `GasOps::gas_available`/`GasOps::charge_gas` to the main `Kernel` trait as all kernels require them.
2. Remove `GasOps::gas_used`, it's only used for testing.
3. Remove `GasOps::price_list`. We're still exposing it on the CallManager, but one step at a time).

### DIFF
--- a/fvm/src/engine/mod.rs
+++ b/fvm/src/engine/mod.rs
@@ -572,6 +572,7 @@ impl Engine {
             last_memory_bytes: memory_bytes,
             last_charge_time: GasTimer::start(),
             memory: self.inner.dummy_memory,
+            wasm_prices: self.inner.config.wasm_prices,
         };
 
         let mut store = wasmtime::Store::new(&self.inner.engine, id);
--- a/fvm/src/gas/price_list.rs
+++ b/fvm/src/gas/price_list.rs
@@ -259,8 +259,6 @@ lazy_static! {
 
         block_persist_compute: Gas::new(172000),
 
-        syscall_cost: Gas::new(14000),
-
         // TODO(#1347)
         builtin_actor_manifest_lookup: Zero::zero(),
         // TODO(#1347)
@@ -286,6 +284,8 @@ lazy_static! {
             // Charge 0.4gas/byte for copying/fill.
             memory_copy_per_byte_cost: Gas::from_milligas(400),
             memory_fill_per_byte_cost: Gas::from_milligas(400),
+
+            host_call_cost: Gas::new(14000),
         },
 
         event_per_entry: ScalingCost {
@@ -441,9 +441,6 @@ pub struct PriceList {
     /// Gas cost to cover the cost of flushing a block.
     pub(crate) block_persist_compute: Gas,
 
-    /// General gas cost for performing a syscall, accounting for the overhead thereof.
-    pub(crate) syscall_cost: Gas,
-
     /// Rules for execution gas.
     pub(crate) wasm_rules: WasmGasPrices,
 
@@ -504,6 +501,27 @@ pub struct WasmGasPrices {
     pub(crate) memory_access_cost: Gas,
     /// Gas cost for every byte copied in Wasm memory.
     pub(crate) memory_copy_per_byte_cost: Gas,
+
+    /// Gas cost for a call from wasm to the system.
+    pub(crate) host_call_cost: Gas,
+}
+
+impl WasmGasPrices {
+    /// Returns the gas required for initializing memory.
+    pub(crate) fn init_memory_gas(&self, min_memory_bytes: usize) -> Gas {
+        self.memory_fill_base_cost + self.memory_fill_per_byte_cost * min_memory_bytes
+    }
+
+    /// Returns the gas required for growing memory.
+    pub(crate) fn grow_memory_gas(&self, grow_memory_bytes: usize) -> Gas {
+        self.memory_fill_base_cost + self.memory_fill_per_byte_cost * grow_memory_bytes
+    }
+
+    /// Returns the gas required for initializing tables.
+    pub(crate) fn init_table_gas(&self, min_table_elements: u32) -> Gas {
+        self.memory_fill_base_cost
+            + self.memory_fill_per_byte_cost * min_table_elements * TABLE_ELEMENT_SIZE
+    }
 }
 
 impl PriceList {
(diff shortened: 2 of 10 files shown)


fvm/src/engine/mod.rs
```diff
@@ -572,6 +572,7 @@ impl Engine {
             last_memory_bytes: memory_bytes,
             last_charge_time: GasTimer::start(),
             memory: self.inner.dummy_memory,
+            wasm_prices: self.inner.config.wasm_prices,
```
Stebalien, line 575: In the future, we may want to store a copy of the entire engine here. But that's not necessary yet.

fvm/src/gas/price_list.rs
```diff
@@ -286,6 +284,8 @@ lazy_static! {
             // Charge 0.4gas/byte for copying/fill.
             memory_copy_per_byte_cost: Gas::from_milligas(400),
             memory_fill_per_byte_cost: Gas::from_milligas(400),
+
+            host_call_cost: Gas::new(14000),
```
Stebalien, line 288: So, the "wasm" prices are _generally_ just the prices used for wasm instrumentation. But the host call cost is so tightly bound to the wasm runtime that it kind of makes sense to just put it in this structure (and means that the syscall binding logic doesn't need to know about the global price list.

fvm/src/gas/price_list.rs
```diff
@@ -504,6 +501,27 @@ pub struct WasmGasPrices {
     pub(crate) memory_access_cost: Gas,
     /// Gas cost for every byte copied in Wasm memory.
     pub(crate) memory_copy_per_byte_cost: Gas,
+
+    /// Gas cost for a call from wasm to the system.
+    pub(crate) host_call_cost: Gas,
+}
+
+impl WasmGasPrices {
```
Stebalien, line 509: Moved from the global `PriceList` struct.

fvm/src/gas/price_list.rs
```diff
@@ -504,6 +501,27 @@ pub struct WasmGasPrices {
     pub(crate) memory_access_cost: Gas,
     /// Gas cost for every byte copied in Wasm memory.
     pub(crate) memory_copy_per_byte_cost: Gas,
+
+    /// Gas cost for a call from wasm to the system.
+    pub(crate) host_call_cost: Gas,
```
Stebalien, line 506: Renamed form `syscall_cost` as wasm calls these "host calls".

fvm/src/syscalls/gas.rs
```diff
@@ -4,10 +4,11 @@ use std::str;
 
 use super::Context;
 use crate::gas::Gas;
-use crate::kernel::{ClassifyResult, GasOps, Result};
+use crate::kernel::{ClassifyResult, Result};
+use crate::Kernel;
 
 pub fn charge_gas(
```
Stebalien, line 10: This syscall will go away once we get rid of the original prove-commit method: https://github.com/filecoin-project/FIPs/pull/820.

fvm/src/syscalls/gas.rs
```diff
@@ -21,6 +22,6 @@ pub fn charge_gas(
         .map(|_| ())
 }
 
-pub fn available(context: Context<'_, impl GasOps>) -> Result<u64> {
+pub fn available(context: Context<'_, impl Kernel>) -> Result<u64> {
```
Stebalien, line 25: This syscall will _hopefully_ go away and turn into a read of a global variable. See https://github.com/filecoin-project/FIPs/discussions/845

fvm/src/syscalls/linker.rs
```diff
@@ -141,15 +141,6 @@ fn memory_and_data<'a, K: Kernel>(
     (Memory::new(mem), data)
 }
 
-macro_rules! charge_syscall_gas {
```
Stebalien, line 144: Changed to a function there was no need to make it a macro.

fvm/src/syscalls/linker.rs
```diff
@@ -171,9 +162,9 @@ macro_rules! impl_syscall {
                     // If we're returning a zero-sized "value", we return no value therefore and expect no out pointer.
                     linker.0.func_wrap(module, name, move |mut caller: Caller<'_, InvocationData<K>> $(, $t: $t)*| {
                         charge_for_exec(&mut caller)?;
+                        charge_syscall_gas(&mut caller)?;
```
Stebalien, line 165: Ordering didn't matter.

testing/conformance/src/vm.rs
```diff
@@ -268,24 +281,23 @@ impl FilecoinKernel for TestKernel {
         extra: &[u8],
     ) -> Result<Option<ConsensusFault>> {
         let charge = self
-            .0
             .price_list()
             .on_verify_consensus_fault(h1.len(), h2.len(), extra.len());
```
Stebalien, line 285: Yes, we still need to access the price list on the kernel. However, the prices we're accessing here are _not_ actually "global" prices but instead filecoin-specific costs.

fridrik01: hmm, not sure I understand the difference, isn't it still resolving to the same global price_list (via filecoinKernel.call_manager.price_list())? 

See  definition in Callmanager:

```rust
/// Returns the current price list.
fn price_list(&self) -> &PriceList {
    self.machine().context().price_list
}
```

where `price_list` is expanded to NetworkConfig using `derive_more` macro :

```rust
pub price_list: &'static PriceList,
```

How is this not actually the global price_list?

fvm/src/engine/mod.rs
Stebalien: Honestly, no reason. I'm pulling out a small piece of the engine config here so I considered just stashing the engine. But yeah... this comment makes no sense without knowing my local history of this patch.

fvm/src/gas/price_list.rs
Stebalien: You mean the VM (wasmtime)? We should, but we haven't bothered re-measuring them for a while.

But I should expand on this a bit more: The _rest_ of the prices here are actually baked into the compiled wasm modules when we instrument the code with gas accounting. That means they get cached in the engine (along with the compiled wasm modules) between epochs (across multiple FVM instances).

The host-call cost doesn't quite belong here because we charge for that on the outside so it doesn't become a part of the compiled wasm modules. Although, TBH, we _should_ charge for it inside wasm as we generally avoid "after the fact" charges like this.

testing/conformance/src/vm.rs
Stebalien: Yes. My point is that, while these prices are in the "global" price list, they really shouldn't be. So this _is_ the global price list, but the "global" price list contains Filecoin-specific prices.

I'm calling this out to explain why exposing this price list on the `kernel` doesn't make that much sense, even if we're accessing it here.

fridrik01 (approved):
LGTM

PR #1952, opened 2023-12-14
refactor: move send/upgrade out of the kernel

Stebalien:
Make them higher-level ops like everything else. I considered combining them into a single "call op" however, not all users may want to implement actor upgrading.

I've also moved the generic from the methods to the trait itself. This resolves anorth's comment on the `TestKernel` by allowing additional bounds to be specified by the _implementer_, instead of having all the bounds encoded on the trait itself:

> Note that KK, the type of the kernel to crate for the receiving actor, is ignored, and Self is passed as the type parameter for the nested call. If we could find the correct bound to specify KK for the call, we would. This restricts the ability for the TestKernel to itself be wrapped by another kernel type.

### DIFF
--- a/fvm/src/kernel/filecoin.rs
+++ b/fvm/src/kernel/filecoin.rs
@@ -97,8 +97,8 @@ pub trait FilecoinKernel: Kernel {
 #[delegate(NetworkOps, where = "C: CallManager")]
 #[delegate(RandomnessOps, where = "C: CallManager")]
 #[delegate(SelfOps, where = "C: CallManager")]
-#[delegate(SendOps<K>, generics = "K", where = "C: CallManager, K: FilecoinKernel")]
-#[delegate(UpgradeOps<K>, generics = "K", where = "C: CallManager, K: FilecoinKernel")]
+#[delegate(SendOps<K>, generics = "K", where = "K: FilecoinKernel")]
+#[delegate(UpgradeOps<K>, generics = "K", where = "K: FilecoinKernel")]
 pub struct DefaultFilecoinKernel<C>(pub DefaultKernel<C>);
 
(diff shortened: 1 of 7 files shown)


Stebalien:
After this, the only parts remaining from https://github.com/filecoin-project/ref-fvm/pull/1933 are:

1. Moving `SyscallHandler` to an associated type. I'm not actually sure if we want to do this (doesn't seem to matter in practice).
2. Refactor the syscall layering per https://github.com/filecoin-project/FIPs/discussions/845. But I'm not sure if I've gotten that layering right, so I'd rather take it one step at a time.

fvm/src/kernel/filecoin.rs
```diff
@@ -97,6 +97,8 @@ pub trait FilecoinKernel: Kernel {
 #[delegate(NetworkOps, where = "C: CallManager")]
 #[delegate(RandomnessOps, where = "C: CallManager")]
 #[delegate(SelfOps, where = "C: CallManager")]
+#[delegate(SendOps<K>, generics = "K", where = "C: CallManager, K: FilecoinKernel")]
+#[delegate(UpgradeOps<K>, generics = "K", where = "C: CallManager, K: FilecoinKernel")]
```
fridrik01, line 101: Would this not suffice?

```suggestion
#[delegate(SendOps<K>, generics = "K", where = "K: FilecoinKernel")]
#[delegate(UpgradeOps<K>, generics = "K", where = "K: FilecoinKernel")]
```

fridrik01 (approved):
Only one minor comment, but otherwise lgtm.

Great to move the send/upgrade into these ops to make things more flexible (and remove boilerplate code)

PR #1953, opened 2023-12-15
fix: don't record the call error twice if we hit the max depth

Stebalien:
We'll record the trace error in `call_actor` itself.


fridrik01 (approved):
LGTM

PR #1954, opened 2023-12-15
feat: record the entire actor state on invoke

Stebalien:
This change has two parts:

1. Record the _entire_ actor state on invoke. This will let us accurately determine the state, address, etc. of an actor at the time of invocation even if the transaction is later reverted.
2. Stop tracing non-"send" events. Specifically:
    - I've stopped tracing upgrades. They're not traditional "calls" and we'll want to handle them differently anyways.
    - Previously, we were tracing InvokeActor events for implicit constructor calls and upgrades. I've removed those as well as I want all InvokeActor events to be paired with some other event.

This should finally give us enough information in Lotus to accurately translate an FVM trace into an EVM trace. Furthermore, the extra state we're recording in the InvokeActor event will be useful for debugging.

### DIFF
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -183,18 +183,20 @@ where
         K: Kernel<CallManager = Self>,
     {
         if self.machine.context().tracing {
-            self.trace(ExecutionEvent::Call {
-                from,
-                to,
-                entrypoint,
-                params: params.as_ref().map(Into::into),
-                value: value.clone(),
-                gas_limit: std::cmp::min(
-                    gas_limit.unwrap_or(Gas::from_milligas(u64::MAX)).round_up(),
-                    self.gas_tracker.gas_available().round_up(),
-                ),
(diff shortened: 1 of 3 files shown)


Stebalien:
Basically, I tried really hard to produce a reasonable trace in lotus without this information, it's just not possible. For example:

1. Actor A calls some actor B.
  1. Actor B creates some smart contract C, assigning a new ID to that smart contract.
  2. Actor B calls smart contract C using the _id_ address.
  3. The call reverts.
2. Actor A creates a new smart contract D (re-using the same ID 😱 ).

When converting this trace to an Ethereum JSON-RPC trace, we'd (a) try to resolve the ID in step 1.2 and end up with the 0x address of actor **D**, not actor **C**.

We're resolving this by putting enough information in the trace itself such that we _never_ need to look at the state tree, ever, when translating the trace. This removes a significant amount of guess work.

fridrik01 (approved):
lgtm

PR #1955, opened 2023-12-15
Move dependencies into workspace

fridrik01:
This PR moves all dependencies and metadata (version, license, etc) to the root workspace so that we can manage them in a central place.


fvm/Cargo.toml
```diff
@@ -1,63 +1,54 @@
...
-default-features = false
-features = ["cranelift", "pooling-allocator", "parallel-compilation"]
-
-[dependencies.wasmtime-environ]
-version = "12.0.1"
-
-[dependencies.wasmtime-runtime]
-version = "12.0.1"
-default-features = false
+fvm = { workspace = true, features = ["testing"], default-features = false }
```
Stebalien, line 51: IIRC, `default-features = false` doesn't matter here (if you specify `workspace = true`). Honestly, I would just stick with the path dependency (I think we want that anyways).

Stebalien: Maybe. I think cargo will warn if it's actually an issue?

ipld/hamt/fuzz/Cargo.toml
```diff
@@ -9,20 +9,15 @@ edition = "2021"
...
+arbitrary = { workspace = true, features = ["derive"] }
+ahash = { workspace = true }
 
-fvm_ipld_hamt = { path = ".." }
-fvm_ipld_blockstore = { path = "../../blockstore" }
-
-
-# Prevent this from interfering with workspaces
-[workspace]
-members = ["."]
```
Stebalien, line 22: Ah.... I _think_ this may have been why we didn't include this in the workspace. We need to check the `cargo tree` (and enabled features) in the workspace itself to make sure we're not "unifying" them. I'll go ahead and do that.

Stebalien: Basically, look up "feature unification in rust" if you want to have bad dreams.

fvm/Cargo.toml
Stebalien: Hm. I guess this one works?

ipld/hamt/fuzz/Cargo.toml
Stebalien: Hm. Actually... this seems fine? I _thought_ I saw the arbitrary feature/crates getting included in the main dependency tree... but now I'm not seeing that.

Stebalien:
Ok, so, I think this actually works. But is there any reason not to migrate the rest of the deps (e.g., `wat` in the integration tests, the rest of the `fvm` dependencies, etc.

Are w ejust moving "shared" deps?

fridrik01:
> Ok, so, I think this actually works. But is there any reason not to migrate the rest of the deps (e.g., `wat` in the integration tests, the rest of the `fvm` dependencies, etc.
> 
> Are we just moving "shared" deps?

Yeah, I skipped moving dependencies that only were used in one crate into the root workspace

Stebalien (approved):

PR #1956, opened 2023-12-15
fix: revert backtrace frame from entrypoint to method

Stebalien:
This is a pretty public API (information gets displayed to the user) and I'd rather not go about changing it at the moment. We'll probably bring this back eventually, but I'm trying to cut a release and integrate with the FFI with minimal outward-facing changes.


Stebalien:
> We'll probably bring this back eventually

Also, when we do that, I expect we'll change the `Frame` to make more sense in this case in this case, maybe? Maybe even include the code CID?

Stebalien:
Merging now so I can start cutting releases.

Issue #1862, opened 2023-08-30
Implement external iteration for the KAMT


Stebalien:
Port the changes in #1833 to the KAMT. This _should_ be fairly straight-forward.

lordshashank:
Hey @Stebalien I am working on this issue, I am implementing writing iter.rs for kamt types referring to that implemented for hamt types, I could not find the load function for node in kamt type so as to implement the Link case for next node as done here for hamt type. Should I have to write the load function for kamt node or am I doing something wrong, would be great if you could  help.

Stebalien:
I am currently on vacation and won't be back for a few weeks. However, I think the answer is no.

See the original implementation in https://github.com/filecoin-project/ref-fvm/pull/1833/files. The `load` function came afterwards and exists to perform more paranoid validation. I'd like to port that as well, but that's a separate task.

lordshashank:
Hey @Stebalien raised a pr resolving this https://github.com/filecoin-project/ref-fvm/pull/1958 . Feel free to review it whenever you get time.

PR #1958, opened 2023-12-27
Resolving Kamt issue#1862

lordshashank:
- Resolving https://github.com/filecoin-project/ref-fvm/issues/1862
- implemented iter and iter_from for kamt referring to hamt implementation.
- wrote simple test for iter_from [here](https://github.com/lordshashank/ref-fvm/blob/088dcbf015756ac7bd887cfa273d2de522b572a4/ipld/kamt/src/kamt.rs#L409)
- Also implemented function to declare kamt with custom bit width [here](https://github.com/lordshashank/ref-fvm/blob/088dcbf015756ac7bd887cfa273d2de522b572a4/ipld/kamt/src/kamt.rs#L86) and wrote its test [here](https://github.com/lordshashank/ref-fvm/blob/088dcbf015756ac7bd887cfa273d2de522b572a4/ipld/kamt/src/kamt.rs#L437)

### DIFF
--- a/ipld/kamt/src/iter.rs
+++ b/ipld/kamt/src/iter.rs
@@ -18,7 +18,6 @@ pub struct IterImpl<'a, BS, V, K, H, const N: usize = 32> {
     current: std::slice::Iter<'a, KeyValuePair<K, V>>,
 }
 
-pub type Iter<'a, BS, V, K, H, const N: usize = 32> = IterImpl<'a, BS, V, K, H, N>;
 impl<'a, K, V, BS, H, const N: usize> IterImpl<'a, BS, V, K, H, N>
 where
     K: DeserializeOwned,
@@ -44,15 +43,14 @@ where
         Q: PartialEq,
         H: AsHashedKey<Q, N>,
     {
-        println!("IterImpl::new_from");
         let hashed_key = H::as_hashed_key(key);
         let mut hash = HashBits::new(&hashed_key);
         let mut node = root;
-        let mut ext;
         let mut stack = Vec::new();
 
         loop {
             let idx = hash.next(conf.bit_width)?;
+            let ext;
             stack.push(node.pointers[node.index_for_bit_pos(idx)..].iter());
             (node, ext) = match stack.last_mut().unwrap().next() {
                 Some(p) => match p {
--- a/ipld/kamt/src/iter.rs
+++ b/ipld/kamt/src/iter.rs
@@ -12,13 +12,13 @@ use crate::node::{match_extension, ExtensionMatch, Node};
 use crate::pointer::Pointer;
 use crate::{AsHashedKey, Config, Error, KeyValuePair};
 #[doc(hidden)]
-pub struct IterImpl<'a, BS, V, K, H, const N: usize = 32> {
+pub struct Iter<'a, BS, V, K, H, const N: usize = 32> {
     store: &'a BS,
     stack: Vec<std::slice::Iter<'a, Pointer<K, V, H, N>>>,
     current: std::slice::Iter<'a, KeyValuePair<K, V>>,
 }
 
-impl<'a, K, V, BS, H, const N: usize> IterImpl<'a, BS, V, K, H, N>
+impl<'a, K, V, BS, H, const N: usize> Iter<'a, BS, V, K, H, N>
 where
     K: DeserializeOwned,
     V: DeserializeOwned,
@@ -97,7 +97,7 @@ where
         }
     }
 }
-impl<'a, K, V, BS, H, const N: usize> Iterator for IterImpl<'a, BS, V, K, H, N>
+impl<'a, K, V, BS, H, const N: usize> Iterator for Iter<'a, BS, V, K, H, N>
 where
     BS: Blockstore,
     K: DeserializeOwned + PartialOrd,
@@ -145,7 +145,7 @@ where
     }
 }
 
-impl<'a, K, V, BS, H, const N: usize> FusedIterator for IterImpl<'a, BS, V, K, H, N>
+impl<'a, K, V, BS, H, const N: usize> FusedIterator for Iter<'a, BS, V, K, H, N>
 where
     K: DeserializeOwned + PartialOrd,
     V: DeserializeOwned,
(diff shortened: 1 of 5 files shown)


Stebalien:
This is almost correct except for `iter_from`. Unlike `HAMT`, the `KAMT` supports "extensions" to collapse keys with spans of identical bits.

You can see how we use them here:

https://github.com/filecoin-project/ref-fvm/blob/088dcbf015756ac7bd887cfa273d2de522b572a4/ipld/kamt/src/node.rs#L241-L244

I'm pretty sure all you need to do is save the extension instead of discarding it, and call:

```rust
        match match_extension(conf, hashed_key, ext)? {
            ExtensionMatch::Full { .. } => {},
            ExtensionMatch::Partial { .. } => return Err(Error::StartKeyNotFound),
        }
```

At the end of every loop. But you'll also need to write a test or two. I.e., make a kamt with the keys 0x00000000000001, 0x00000000000002, 0x10000000000001, 0x10000000000002, 0x20000000000001, 0x20000000000002, etc... then try to start at `0x10000000000001`.

Stebalien:
And sorry this took so long for me to get to!

ipld/kamt/src/iter.rs
```diff
@@ -145,3 +155,40 @@ where
     BS: Blockstore,
 {
 }
+
+/// Helper method to check if a key matches an extension (if there is one)
+/// and return the number of levels skipped. If the key doesn't match,
+/// this will be the number of levels where the extension has to be split.
+fn match_extension<'a>(
```
Stebalien, line 162: We probably need to export this from the `node` module with `pub(crate)`.

ipld/kamt/src/kamt.rs
```diff
@@ -81,6 +82,17 @@ where
         }
     }
 
+    /// Construct kamt with a bit width
+    pub fn new_with_bit_width(store: BS, bit_width: u32) -> Self {
```
Stebalien, line 86: Let's drop this for now. We have it on the HAMT for historical reasons, but `new_with_config` is more general.

ipld/kamt/src/kamt.rs
```diff
@@ -352,6 +364,133 @@ where
...
+    ///     x = x+1;
+    /// }
+    /// assert_eq!(x,2)
+    /// ```
+    pub fn iter(&self) -> IterImpl<BS, V, K, H, N> {
+        IterImpl::new(&self.store, &self.root)
+    }
+
+    /// Iterate over the KAMT starting at the given key.
+    /// iteration:
```
Stebalien, line 410: stray line?

ipld/kamt/src/kamt.rs
```diff
@@ -352,6 +364,133 @@ where
...
+    /// Iterate over the KAMT starting at the given key. This can be used to implement "ranged" iteration:
+    ///
+    /// ```rust
+    /// use fvm_ipld_kamt::Kamt;
+    /// use fvm_ipld_blockstore::MemoryBlockstore;
+    /// use fvm_ipld_kamt::id::Identity;
+    /// let store = MemoryBlockstore::default();
+    ///
+    /// // Create a Kamt with 5 keys, a-e.
+    /// let mut kamt: Kamt<_, u32, String, Identity> = Kamt::new_with_bit_width(store, 5);
```
Stebalien, line 446: I'd just change this to `Kamt::new_with_config` (or drop the bit-width, I don't think that it's actually necessary to specify it?

ipld/kamt/src/iter.rs
```diff
@@ -0,0 +1,156 @@
...
+use crate::pointer::Pointer;
+use crate::{AsHashedKey, Config, Error, KeyValuePair};
+#[doc(hidden)]
+pub struct IterImpl<'a, BS, V, K, H, const N: usize = 32> {
+    store: &'a BS,
+    stack: Vec<std::slice::Iter<'a, Pointer<K, V, H, N>>>,
+    current: std::slice::Iter<'a, KeyValuePair<K, V>>,
+}
+
+pub type Iter<'a, BS, V, K, H, const N: usize = 32> = IterImpl<'a, BS, V, K, H, N>;
```
Stebalien, line 21: We don't need a separate `Iter` and `IterImpl`. We only have that for the `Hamt` to handle the multiple versions.

ipld/kamt/src/iter.rs
```diff
@@ -0,0 +1,156 @@
...
+        root: &'a Node<K, V, H, N>,
+        key: &Q,
+        conf: &'a Config,
+    ) -> Result<Self, Error>
+    where
+        K: Borrow<Q>,
+        Q: PartialEq,
+        H: AsHashedKey<Q, N>,
+    {
+        println!("IterImpl::new_from");
```
Stebalien, line 47: remove println.

ipld/kamt/src/iter.rs
```diff
@@ -0,0 +1,156 @@
...
+        let hashed_key = H::as_hashed_key(key);
+        let mut hash = HashBits::new(&hashed_key);
+        let mut node = root;
+        let mut ext;
```
Stebalien, line 51: This should be inside the loop (doesn't cost anything extra and keeps us from keeping it around between iterations).

Stebalien: And, inside the loop, it doesn't need to be `mut`.

Stebalien (commented):
Getting down to the final nits. The only real remaining thing is a good test-case.

ipld/kamt/src/iter.rs
```diff
@@ -0,0 +1,154 @@
...
+use fvm_ipld_blockstore::Blockstore;
+use fvm_ipld_encoding::de::DeserializeOwned;
+use fvm_ipld_encoding::CborStore;
+
+use crate::hash_bits::HashBits;
+use crate::node::{match_extension, ExtensionMatch, Node};
+use crate::pointer::Pointer;
+use crate::{AsHashedKey, Config, Error, KeyValuePair};
+#[doc(hidden)]
+pub struct IterImpl<'a, BS, V, K, H, const N: usize = 32> {
```
Stebalien, line 15: Let's rename this to `Iter` and remove the `#[doc(hidden)]`. We only did it that way because we had multiple versions.

lordshashank: done

PR #1965, opened 2024-01-22
Implement external iteration for AMT range iteration

Stebalien:
1. Implement `iter_from` to replace `for_each_ranged`.
2. Make the diffing logic use the iterator so we can remove the old code.

Part of #1861.


ipld/amt/src/iter.rs
```diff
@@ -41,17 +42,110 @@ where
...
+    /// use fvm_ipld_amt::Amt;
+    /// use fvm_ipld_blockstore::MemoryBlockstore;
+    ///
+    /// let store = MemoryBlockstore::default();
+    ///
+    /// let mut amt = Amt::new(store);
+    /// let kvs: Vec<u64> = (0..=5).collect();
+    /// kvs
+    ///     .iter()
+    ///     .map(|k| amt.set(u64::try_from(*k).unwrap(), k.to_string()))
```
rvagg, line 66:
```suggestion
    ///     .map(|k| amt.set(*k, k.to_string()))
```
?

ipld/amt/src/iter.rs
```diff
@@ -41,17 +42,110 @@ where
...
+    pub fn iter_from(&self, start: u64) -> Result<Iter<'_, V, &BS, Ver>, Error> {
+        // Short-circuit when we're starting at 0.
+        if start == 0 {
+            return Ok(self.iter());
+        }
+
+        let height = self.height();
+        let bit_width = self.bit_width();
+
+        if start >= nodes_for_height(bit_width, height + 1) {
```
rvagg, line 85:
```suggestion
        // Fast-path for case where start is beyond what we know this amt could currently contain.
        if start >= nodes_for_height(bit_width, height + 1) {
```

ipld/amt/src/amt.rs
```diff
@@ -517,25 +518,16 @@ where
...
     {
-        if let Some(start_at) = start_at {
-            if start_at >= nodes_for_height(self.bit_width(), self.height() + 1) {
-                return Ok((0, None));
+        let mut num_traversed = 0;
+        for kv in self.iter_from(start_at.unwrap_or(0))? {
+            let (k, v) = kv?;
+            if limit.map(|l| num_traversed >= l).unwrap_or(false) {
+                return Ok((num_traversed, Some(k)));
             }
```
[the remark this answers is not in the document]
Stebalien: Because...

1. I may or may not have a limit and I didn't really want to have a sentinel max value.
2. `take(limit)` takes a usize, this function takes a u64, and I didn't want to break anything.
3. I needed to take the limit, plus 1 (to get the next index). And I needed a way to detect when I'm on the last one. So I would have needed something like:

```rust
let mut iter = self.iter_from(...);
for kv in iter.by_ref().take(limit) {
}
if let Some(next_idx) = iter.next() {
}
```

(basically, I wrote it that way, and then realized this way was simpler)

ipld/amt/src/iter.rs
Stebalien: Hm. well, I'd need to change the line above to `0u64..=5` (I'm copying this document from the existing method, but you're right that that way is better).

Maybe file an issue and I can fix both `iter` and `iter_from` in another PR?

rvagg (approved):
lgtm, I'm not so familiar with this version of the amt but it matches my expectations from knowing the algorithm

ipld/amt/src/iter.rs
rvagg: https://github.com/filecoin-project/ref-fvm/pull/1971 there you go, I don't think you need `0u64..=5` cause the type is specified on the receiver

PR #1971, opened 2024-01-25
doc(amt): simplify example doc

rvagg:


Stebalien (approved):
Ah, I missed the `Vec<u64>`.

PR #1976, opened 2024-01-25
feat: port hamt validation logic to kamt

Stebalien:
This ensures that:

1. The KAMT is well-formed (mostly).
2. We won't panic when reading "bad" KAMTs (we don't currently use this library to read KAMTs we didn't write, but that may change eventually).

This will have a bit of a performance impact, but it should be minimal.


Stebalien:
Note: "mostly" comes from the fact that we _don't_ validate that keys seen when iterating are in the correct positions, aren't duplicates, etc. That would have a more significant performance impact.

rvagg (approved):
nice (and I recognise most of these cause I wrote their originals! https://github.com/filecoin-project/go-hamt-ipld/commit/570d0599136603300093ba6eeee3fe155dfec6e7)

adding failure case tests would be good of course, but I know that's tedious work

PR #1977, opened 2024-01-31
Backport FVMv4 tracing changes to v3

Stebalien:
FVMv3 exposed actor code-CIDs in traces, but v4 now exposes the entire invoked actor. This patch backports the FVMv4 logic to v3.

However, note: this change is slightly simplified. FVMv3 has a bug where we might include _multiple_ invoke events for the same actor if, e.g., we had to implicitly create it. This was fixed in FVMv4 but that fix has _not_ been backported. In practice, this shouldn't cause any issues in FVMv3 as long as the _last_ invoke event is used (which is how the filecoin-ffi library handles this).

### DIFF
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -660,7 +660,10 @@ where
             .ok_or_else(|| syscall_error!(NotFound; "actor does not exist: {}", to))?;
 
         if self.machine.context().tracing {
-            self.trace(ExecutionEvent::InvokeActor(state.code));
+            self.trace(ExecutionEvent::InvokeActor {
+                id: to,
+                state: state.clone(),
+            });
         }
 
         // Transfer, if necessary.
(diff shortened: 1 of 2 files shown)


Stebalien:
For some more context, I'm backporting this because it makes it easier to translate FVMv3 traces into FVMv4 traces in the FFI (i.e., easier to abstract over versions) without losing information.

rvagg (approved):
This is, technically, a breaking change though for consumers of the tracing API though, right? We're not very strict on those kinds of breakages in terms of versioning in this I guess?

Stebalien:
> This is, technically, a breaking change though for consumers of the tracing API though, right? We're not very strict on those kinds of breakages in terms of versioning in this I guess?

We use minor versions for breaking changes, and major versions when we drop support for old network versions. We'd like to switch to something better (e.g., separate fvm1, fvm2, etc. crates). But this is what we do for now...

PR #1984, opened 2024-02-09
feat: shared: check bls zero address without lazy_static

Stebalien:
Trying to clean up this module a bit. I'll remove the other uses of lazy_static in a future (significantly more breaking) change.


rvagg (approved):
slightly amusing that it's that simple; I had to check it myself to see

Stebalien:
Yeah, I was kind of surprised.

PR #1986, opened 2024-02-09
feat: slim down shared

Stebalien:
- `BLOCK_GAS_LIMIT`: Neither the FVM nor the builtin actors care about the block gas limit. And if they did, it would have to be a runtime parameter passed from the client.
- `EPOCH_DURATION_SECONDS`: Different networks may have different block times, this constant is useless.
- `math`, `reward`, `smooth`: part of the reward actor calculations, not relevant to the FVM itself (moved to the builtin actors).
- `TOTAL_FILECOIN`, `TOTAL_FILECOIN_BASE`: Network parameter, not a concern of the FVM.
- `BLOCKS_PER_EPOCH`: protocol parameter, moved to the builtin actors.
- `MAX_SECTOR_NUMBER`: moved to the builtin actors.
- `Spacetime`, `SectorQuality`: moved to the builtin actors.
- `ZERO_ADDRESS`, `*_LOOKBACK`, `ALLOWABLE_CLOCK_DRIFT`, `NetworkParams`, `DefaultNetworkParams`: unused.


Stebalien:
IIRC, this was in draft pending merging stuff in the built-in actors, but we might as well merge it now.

shared/src/lib.rs
```diff
@@ -62,51 +45,11 @@ pub const IDENTITY_HASH: u64 = 0x0;
 /// The maximum supported CID size.
 pub const MAX_CID_LEN: usize = 100;
 
-/// Identifier for Actors, includes builtin and initialized actors
-pub type ActorID = u64;
-
 /// Default bit width for the hamt in the filecoin protocol.
 pub const HAMT_BIT_WIDTH: u32 = 5;
-/// Total gas limit allowed per block. This is shared across networks.
-pub const BLOCK_GAS_LIMIT: u64 = 10_000_000_000;
```
rvagg, line 71: Is this only enforced in the miner at message selection time and in nodes validating messages in blocks packed by miners? Otherwise this isn't interesting anywhere else?

rvagg (approved):
seems fine to me, I can't find these referenced in builtin-actors, my only question is about the blockgaslimit

shared/src/lib.rs
Stebalien: Yes. We only enforce per-message gas limits in the FVM. Otherwise cron would be an issue.

Stebalien: Really, the FVM doesn't even see "blocks", it just sees a bunch of messages to be applied at a specific height on top of a previous state-tree.

PR #1987, opened 2024-02-09
chore: fvm: remove another lazy_static

Stebalien:
We can pre-compute the "empty" block and avoid a lazy static. It matters less here because we're not inside an actor (so we'd only compute this once on start anyways, but I'm trying to reduce our usage of lazy_static to the minimum.

### DIFF
--- a/ipld/blockstore/src/block.rs
+++ b/ipld/blockstore/src/block.rs
@@ -1,10 +1,12 @@
+use std::fmt;
+
 // Copyright 2021-2023 Protocol Labs
 // SPDX-License-Identifier: Apache-2.0, MIT
 use cid::multihash::{self, MultihashDigest};
 use cid::Cid;
 
 /// Block represents a typed (i.e., with codec) IPLD block.
-#[derive(Copy, Clone, Debug, Eq, PartialEq)]
+#[derive(Copy, Clone)]
 pub struct Block<D>
 where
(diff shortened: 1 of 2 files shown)


ipld/blockstore/src/block.rs
```diff
@@ -35,6 +37,31 @@ where
     }
 }
 
+// Manually implement PartialEq/Eq so we can compare across blocks with different backing buffers.
+impl<D1, D2> PartialEq<Block<D2>> for Block<D1>
```
Stebalien, line 41: I needed to be able to compare `Block<Vec<u8>>` to `Block<&'static [u8]>`.

ipld/blockstore/src/block.rs
```diff
@@ -35,6 +37,31 @@ where
...
+    D1: AsRef<[u8]>,
+    D2: AsRef<[u8]>,
+{
+    fn eq(&self, other: &Block<D2>) -> bool {
+        self.codec == other.codec && self.data.as_ref() == other.data.as_ref()
+    }
+}
+impl<D> Eq for Block<D> where D: AsRef<[u8]> {}
+
+// Manually implement debug so we get the same result regardless of the backing buff for data.
```
Stebalien, line 52: Drive-by fix.

Stebalien:
Eh, this is pretty safe.

PR #1992, opened 2024-02-09
chore: update rust and fix clippy lints

Stebalien:
Needed to upgrade wasmtime, and now's a reasonable time to do this.

### DIFF
--- a/shared/src/address/mod.rs
+++ b/shared/src/address/mod.rs
@@ -243,7 +243,7 @@ impl quickcheck::Arbitrary for Address {
     }
 }
 
-pub(self) fn parse_address(addr: &str) -> Result<(Address, Network), Error> {
+fn parse_address(addr: &str) -> Result<(Address, Network), Error> {
     if addr.len() > MAX_ADDRRESS_TEXT_LEN || addr.len() < 3 {
         return Err(Error::InvalidLength);
     }
@@ -385,6 +385,19 @@ pub(crate) fn from_leb_bytes(bz: &[u8]) -> Result<u64, Error> {
     Ok(id)
 }
 
+/// Returns an address hash for given data
+fn address_hash(ingest: &[u8]) -> [u8; 20] {
+    let digest = blake2b_simd::Params::new()
+        .hash_length(PAYLOAD_HASH_LEN)
+        .to_state()
+        .update(ingest)
+        .finalize();
+
+    let mut hash = [0u8; 20];
+    hash.copy_from_slice(digest.as_bytes());
+    hash
+}
+
 #[cfg(test)]
 mod tests {
     // Test cases for FOR-02: https://github.com/ChainSafe/forest/issues/1134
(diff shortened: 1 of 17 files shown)


shared/src/address/mod.rs
```diff
@@ -385,6 +385,19 @@ pub(crate) fn from_leb_bytes(bz: &[u8]) -> Result<u64, Error> {
     Ok(id)
 }
 
+/// Returns an address hash for given data
```
Stebalien, line 388: Clippy complained about this function being placed after the test module (which, IMO, is a reasonable complaint as tests usually come last).

shared/src/econ/mod.rs
```diff
@@ -81,17 +81,17 @@ impl Zero for TokenAmount {
...
+        self.atto.cmp(&other.atto)
     }
 }
 
-impl Ord for TokenAmount {
+impl PartialOrd for TokenAmount {
     #[inline]
-    fn cmp(&self, other: &Self) -> Ordering {
-        self.atto.cmp(&other.atto)
+    fn partial_cmp(&self, other: &Self) -> Option<Ordering> {
```
Stebalien, line 93: Clippy was unhappy that `partial_cmp` could, in theory, return a different result than `cmp`. It wouldn't have, but this is the "correct" way to do it.

testing/test_actors/actors/fil-oom-actor/src/lib.rs
```diff
@@ -1,5 +1,6 @@
 // Copyright 2021-2023 Protocol Labs
 // SPDX-License-Identifier: Apache-2.0, MIT
+#![allow(clippy::slow_vector_initialization)]
```
Stebalien, line 3: Slowly allocating a bunch of memory is the point of this test.

shared/src/econ/mod.rs
rvagg: heh, _"Not doing this may silently introduce an error upon refactoring."_, fair enough

testing/integration/tests/gas_calibration_test.rs
```diff
@@ -550,7 +550,9 @@ fn on_scan_cbor_links() {
         let mut iter_obs: HashMap<String, Vec<Obs>> = Default::default();
 
         for event in ret.exec_trace {
-            let ExecutionEvent::GasCharge(charge) = event else { continue };
+            let ExecutionEvent::GasCharge(charge) = event else {
+                continue;
+            };
```
rvagg, line 555: all these `let .. else {}` line breaks, are those coming from new formatting rules? they're not a clippy thing as far as I can tell

Stebalien: Yes. Rustfmt changed it's rules.

Stebalien: Which is unfortunate. I have some nostalgia for `foo() or die "oops";` from my (very brief) perl days.

PR #1993, opened 2024-02-09
chore: update wasmtime

Stebalien:


fvm/src/engine/mod.rs
```diff
@@ -209,12 +223,17 @@ fn wasmtime_config(ec: &EngineConfig) -> anyhow::Result<wasmtime::Config> {
     c.cranelift_debug_verifier(false);
     c.native_unwind_info(false);
     c.wasm_backtrace(false);
+    c.wasm_backtrace_details(WasmBacktraceDetails::Disable);
```
Stebalien, line 226: likely unnecessary, but I might as well.

Stebalien:
This seems to sync mainnet.

rvagg:
I'm currently looking at this, but dong some due diligence on how we currently configure and run it and what the delta is between wasmtime 12 and 17.

fvm/src/engine/mod.rs
```diff
@@ -158,16 +161,23 @@ fn wasmtime_config(ec: &EngineConfig) -> anyhow::Result<wasmtime::Config> {
...
+    // wasmtime default: true
+    // We don't want threads, there is no way to ensure determinism
     c.wasm_threads(false);
 
     // wasmtime default: true
-    // simd isn't supported in wasm-instrument, but if we add support there, we can probably enable this.
+    // simd isn't supported in wasm-instrument, but if we add support there, we can probably enable
+    // this.
     // Note: stack limits may need adjusting after this is enabled
     c.wasm_simd(false);
```
rvagg, line 172: This is a bit sad; I haven't looked at wasm-instrument yet but this shouldn't be a big deal to enable should it? just lots of instructions and everything at 128 bits.

rvagg:
We seem to be doing explicit config in here, reiterating all of the available and relevant options and setting custom defaults. If that's the aim, there's some that are not set here that we might want to consider?

* `[static_memory_forced](https://docs.rs/wasmtime/latest/wasmtime/struct.Config.html#method.static_memory_forced)` since we set `static_memory_maximum_size` to max memory I guess we implicitly want this to be `true` so maybe be explicit about it?
* [static_memory_guard_size](https://docs.rs/wasmtime/latest/wasmtime/struct.Config.html#method.static_memory_guard_size) assuming we're not supporting 32-bit runtimes, we're getting 2GB of this as a default (so a total of 8GB virtual address space per instance cause we have the before guard set already).
* [dynamic_memory_guard_size](https://docs.rs/wasmtime/latest/wasmtime/struct.Config.html#method.dynamic_memory_guard_size) is 64K by default, but maybe it only needs to be `0` since we're doing static=max?
* [dynamic_memory_reserved_for_growth](https://docs.rs/wasmtime/latest/wasmtime/struct.Config.html#method.dynamic_memory_reserved_for_growth) defaults to 2GB on 64-bit, but maybe could be `0`?

rvagg (approved):
👍 seems like the big ticket items in the update are memory related and the default enablement of proposed extensions (that all complicate things for our use case!) that have been already explicitly disabled here.

Stebalien:
> We seem to be doing explicit config in here, reiterating all of the available and relevant options and setting custom defaults.

It's mostly to ensure that features we don't want remain off.

> `static_memory_forced`

Good idea, implemented.

> `static_memory_guard_size`

This shouldn't affect consensus, so I'm going to leave it at the default for now. Although we should consider 4GiB for performance.

> `dynamic_memory_*`

I think we can ignore these if we force static memory.

Stebalien:
Tested on mainnet and it still works.

PR #1997, opened 2024-04-04
chore: update to wasmtime 19.0.1

Stebalien:
This looks like a pretty uneventful upgrade.

### DIFF
--- a/fvm/src/engine/mod.rs
+++ b/fvm/src/engine/mod.rs
@@ -196,7 +196,13 @@ fn wasmtime_config(ec: &EngineConfig) -> anyhow::Result<wasmtime::Config> {
 
     // wasmtime default: false
     // Cool proposal to allow function references, but we don't support it yet.
+    #[cfg(feature = "wasmtime/gc")]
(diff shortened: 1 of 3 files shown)


fvm/src/engine/mod.rs
```diff
@@ -225,7 +231,6 @@ fn wasmtime_config(ec: &EngineConfig) -> anyhow::Result<wasmtime::Config> {
     c.native_unwind_info(false);
     c.wasm_backtrace(false);
     c.wasm_backtrace_details(WasmBacktraceDetails::Disable);
-    c.wasm_reference_types(false);
```
Stebalien, line 228: now behind a feature flag

rvagg (approved):
lgtm but any idea what's up with the fuzzing CI job? `error: toolchain 'nightly-x86_64-unknown-linux-gnu' is not installed`

Stebalien:
The fuzzing job probably broke when we switched away from Rust Nightly, when we were no longer automatically installing Rust Nightly, and unfortunately the fuzzer requires Rust Nightly.

PR #1999, opened 2024-04-13
Non Interactive Proof of Replication

NemanjaLu92:


shared/src/sector/seal.rs
```diff
@@ -33,6 +33,18 @@ pub struct SealVerifyInfo {
...
+#[derive(Clone, Debug, PartialEq, Eq, Serialize_tuple, Deserialize_tuple)]
+pub struct NISealVerifyInfo {
+    pub registered_proof: RegisteredSealProof,
+    pub sector_id: SectorID,
+    pub randomness: SealRandomness,
+    #[serde(with = "strict_bytes")]
+    pub proof: Vec<u8>,
+    pub sealed_cid: Cid,   // Commr
+    pub unsealed_cid: Cid, // Commd
+}
```
anorth, line 46: Can we do this without introducing a new NISealVerifyInfo type, and thus without a new syscall at all? The proof type carries the information that it is NI, and then determines that certain fields will be ignored.
- deal_ids: already ignored and always empty
- interactive_randomness: let's just specify it shall be zero (and ignored) for NI proofs

NemanjaLu92: Yeah, that was my initial thinking too. Will change that.

anorth:
This is probably ready for review. We'll need it to land so the actors change can depend on it.

@Stebalien the associated FIP is accepted. Can we land this direct to master?

Stebalien:
It looks like this doesn't build with the "proofs" feature enabled.

rvagg:
just needs the additional `TryFrom` matches added I believe

NemanjaLu92:
> just needs the additional `TryFrom` matches added I believe

True, but I need to change dependancy to the `filecoin-proofs-api` which doesn't include NI PoRep Registered proof in latest version (17.0). How should we proceed? Create a new release in `filecoin-proofs-api` and update dependency in ref-fvm?

rjan90:
> True, but I need to change dependancy to the filecoin-proofs-api which doesn't include NI PoRep Registered proof in latest version (17.0). How should we proceed? Create a new release in filecoin-proofs-api and update dependency in ref-fvm?

A new Filecoin-proofs-api release is expected to be out today (Thursday May 16th), which includes the NI-PoRep registered proofs.

Stebalien:
It looks like `storage-proofs-update` is either missing a minimum version specification for something, or it isn't enabling some feature flag... @cryptonemo?

rjan90:
Can we close this, since its superseded by https://github.com/filecoin-project/ref-fvm/pull/2010?

Stebalien:
Yes.

PR #2010, opened 2024-05-28
NI-PoRep support

rvagg:
#1999 rebased on master with some minor changes and local Cargo.toml linking

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,11 +4,11 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.21.0"
+version = "0.22.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "8a30b2e23b9e17a9f90641c7ab1549cd9b44f296d3ccbf309d2863cfe398a0cb"
+checksum = "6e4503c46a5c0c7844e948c9a4d6acd9f50cccb4de1c48eb9e291ea17470c678"
 dependencies = [
- "gimli",
+ "gimli 0.29.0",
 ]
 
 [[package]]
@@ -19,9 +19,9 @@ checksum = "f26201604c87b1e01bd3d98f8d5d9a8fcbb815e8cedb41ffccbeb4bf593a35fe"
 
 [[package]]
 name = "aes"
-version = "0.8.3"
+version = "0.8.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ac1f845298e95f983ff1944b728ae08b8cebab80d684f0a832ed0fc74dfa27e2"
+checksum = "b169f7a6d4742236a0a00c541b845991d0ac43e546831af1249753ab4c3aa3a0"
 dependencies = [
  "cfg-if",
  "cipher",
@@ -30,9 +30,9 @@ dependencies = [
 
 [[package]]
 name = "ahash"
-version = "0.8.7"
+version = "0.8.11"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "77c3a9648d43b9cd48db467b3f87fdd6e146bcc88ab0180006cef2179fe11d01"
+checksum = "e89da841a80418a9b391ebaea17f5c112ffaaa96f621d2c285b5174da76b9011"
 dependencies = [
  "cfg-if",
  "once_cell",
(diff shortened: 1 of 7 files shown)


rvagg:
@anorth @Stebalien @NemanjaLu92 I'd like to propose this replace https://github.com/filecoin-project/ref-fvm/pull/1999 and be merged as it is. It's essentially the same as #1999 but rebased on master and with a correct proofs dependency version. I've been using this for local testing but I'm keen to move us closer to a tag because cargo linking to the git repo for ref-fvm in filecoin-ffi is a bit of a nightmare thanks to the multiple versions.

shared/src/sector/registered_proof.rs
```diff
@@ -92,14 +98,19 @@ impl RegisteredSealProof {
...
 
             StackedDRG32GiBV1
             | StackedDRG64GiBV1
             | StackedDRG32GiBV1P1
             | StackedDRG64GiBV1P1
             | StackedDRG32GiBV1P1_Feat_SyntheticPoRep
-            | StackedDRG64GiBV1P1_Feat_SyntheticPoRep => Ok(1920),
+            | StackedDRG64GiBV1P1_Feat_SyntheticPoRep
+            | StackedDRG32GiBV1P1_Feat_NiPoRep
+            | StackedDRG64GiBV1P1_Feat_NiPoRep => Ok(1920),
```
cryptonemo, line 113: I'm not familiar with this size here, but is this value correct for NI-PoRep?

anorth: Should it be 10x larger? @irene 

(We're never going to use this if we only support aggregation, but let's get it right).

shared/src/sector/registered_proof.rs
```diff
@@ -410,6 +451,21 @@ impl TryFrom<RegisteredSealProof> for filecoin_proofs_api::RegisteredSealProof {
                 Ok(Self::StackedDrg512MiBV1_1_Feat_SyntheticPoRep)
             }
             Invalid(i) => Err(format!("unsupported proof type: {}", i)),
```
anorth, line 453: Move this case to the bottom

shared/src/sector/registered_proof.rs
lucaniz: it should be 12.8x higher

Stebalien: Can someone produce a proof and measure it? The last time I yoloed one of these values I halted proving on the network (testnet, luckily).

lucaniz: tbh...not really yoloing here, but I agree producing a proof and measuring it would be the best way to know it.

lucaniz: 126 is the number  of circuits per NI-PoRep (thanks @vmx )
> https://github.com/filecoin-project/rust-fil-proofs/blob/0dfef1d8416c8240400fc188e90f0e55fa6b90ae/filecoin-proofs/src/constants.rs#L152

which means the expected size for NI PoRep proof is 192 * 10 * 12.6

As I said, having someone running a proof would be the ultimate confirmation

Stebalien: We need the _exact_ size. I say yoloing because we're attempting to estimate the size by multiplying numbers. I'm assuming there's some rounding somewhere.

`ceil(192 * 10) * 12.6 != ceil(192 * 12.6) * 10`  (24192 versus 24200)

lucaniz: I understand, but my take is we have 126 circuits, each circuit is 192 bytes. So you should not have issues with rounding no?
the 3 terms were to explain where we are coming from and the comparison with current porep.

ANyway, ofc let's run the proof if anyone can to it easily

Stebalien: Got it, that makes sense.

lucaniz: So given we have aggregation within the sector (i thought we did not want to consider aggregation at all here, but i was wrong), the correct value is 23092 bytes (thanks @rvagg for measuring it!).
[ Note that this matches what the Snarkpack paper says about aggregating Vs not aggregating 128 proofs , where they observe 24kb for the non aggregated case and 23kb for the aggregated case, ref here, section 7  https://eprint.iacr.org/2021/529.pdf ]

rvagg: going with `23_092` because it's what I measured from running it, 🤞

rvagg:
feedback addressed

rvagg:
Also renamed all the NI-PoRep proofs types to be `V1P2` rather than `V1P1`

Cargo.lock
```diff
@@ -667,41 +706,41 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.4.6"
+version = "4.5.4"
```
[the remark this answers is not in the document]
rvagg: @Stebalien I didn't do anything special here to make this bump; probably just a `cargo build`. They don't check in a Cargo.lock so maybe these deps float a bit more than they would otherwise? https://github.com/filecoin-project/rust-filecoin-proofs-api/blob/master/.gitignore#L2 should we fix that?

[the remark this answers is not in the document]
rvagg: It's also possible that I deleted cargo.lock and had it rebuild because there was some rebasing involved in this PR because the original branch was well behind master. I'll have a play with doing it differently.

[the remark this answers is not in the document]
rvagg: ahhh I fixed it, it was a rebase problem and was probably from me deleting the lock because of conflicts and rebuilding; I'll have to remember to be more careful about that

Stebalien (approved):

Cargo.lock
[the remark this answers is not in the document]
Stebalien: Yeah, I wish cargo handled that kind of thing better.

PR #2011, opened 2024-06-06
fix(ci): don't use nightly for fuzz build

rvagg:
testing a theory that it's `+nightly` that makes this frustratingly unstable in CI, e.g. currently broken because of `-lc++` linking https://github.com/filecoin-project/ref-fvm/actions/runs/9394298344/job/25872490528?pr=2010


rvagg:
nope, wac-a-mole I don't have patience for: `the option `Z` is only accepted on the nightly compiler`

PR #2015, opened 2024-06-12
chore: release fvm 4.3.0

rjan90:
Bump FVM-release to unblock the NI-PoRep and nv23-testing in Butterfly-network

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -725,7 +725,7 @@ dependencies = [
  "cid 0.10.1",
  "fvm_ipld_bitfield 0.6.0",
  "fvm_ipld_encoding 0.4.0",
- "fvm_shared 4.2.0",
+ "fvm_shared 4.3.0",
  "libfuzzer-sys",
  "multihash 0.18.1",
  "rand",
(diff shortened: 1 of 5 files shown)


Stebalien (approved):

Stebalien:
Released and tagged. But I _think_ we'll need to release v2/v3 with the latest proofs version to make all this work (we can test that theory by trying to update the FVM in the FFI and seeing if it works).

rjan90:
> But I think we'll need to release v2/v3 with the latest proofs version to make all this work (we can test that theory by trying to update the FVM in the FFI and seeing if it works).

Do we have some place where how to do that is written down, or just a link to a PR/code where I can see how its done? I´m a bit unfamiliar with how to release those v2/v3 pacakagase

Stebalien:
1. Checkout release/v3.
2. Update `filecoin-proofs-api` to v18. Also check to see if we updated any other proofs libraries in master and backport those updates to the release/v3 branch. But I don't think we did?
3. Cut a new v3.N release (see https://github.com/filecoin-project/ref-fvm/pull/1982). v3.10 in this case, I think.
4. Repeat for v2.

PR #2020, opened 2024-06-20
feat: simplify the verify-signatures feature

Stebalien:
A new ambassador crate was released so we can now use attributes inside
delegable traits (and some macros that were previously unexported are
now exported by default). Basically, I'm just deleting code.

### DIFF
--- a/fvm/src/kernel/mod.rs
+++ b/fvm/src/kernel/mod.rs
@@ -220,11 +220,11 @@ pub trait ActorOps {
     fn balance_of(&self, actor_id: ActorID) -> Result<TokenAmount>;
 }
 
-#[cfg(feature = "verify-signature")]
 /// Cryptographic primitives provided by the kernel.
 #[delegatable_trait]
 pub trait CryptoOps {
     /// Verifies that a signature is valid for an address and plaintext.
+    #[cfg(feature = "verify-signature")]
     fn verify_signature(
         &self,
         sig_type: SignatureType,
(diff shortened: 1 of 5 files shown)


testing/test_actors/actors/fil-syscall-actor/Cargo.toml
```diff
@@ -6,7 +6,7 @@ publish = false
 
 [target.'cfg(target_arch = "wasm32")'.dependencies]
 fvm_ipld_encoding = { workspace = true }
-fvm_sdk = { workspace = true, default-features = false }
```
Stebalien, line 9: doesn't actually do anything and cargo complains.

fvm/src/kernel/mod.rs
```diff
@@ -341,19 +307,15 @@ pub trait EventOps {
...
-    ambassador_impl_CryptoOps, ambassador_impl_DebugOps, ambassador_impl_EventOps,
-    ambassador_impl_IpldBlockOps, ambassador_impl_MessageOps, ambassador_impl_NetworkOps,
-    ambassador_impl_RandomnessOps, ambassador_impl_SelfOps, ambassador_impl_SendOps,
-    ambassador_impl_UpgradeOps,
-};
-
 /// Import this module (with a glob) if you're implementing a kernel, _especially_ if you want to
 /// use ambassador to delegate the implementation.
 pub mod prelude {
+    pub use super::{
```
Stebalien, line 313: auto-sorted.

fvm/src/kernel/mod.rs
```diff
@@ -341,19 +307,15 @@ pub trait EventOps {
     ) -> Result<()>;
 }
 
-// Unfortunately, I need to do this to make it possible to name these macros by path. I'm hiding
-// this because I really don't want users to glob import the `kernel` module.
-#[doc(hidden)]
-pub use {
```
Stebalien, line 347: automatically exported now.

rvagg: do we have a breaking change now though because these traits are no longer exported on the crate root? (as per [notes](https://docs.rs/ambassador/0.4.0/ambassador/#03x---04x)); or do we consider this unlikely to really impact anyone?

rvagg (approved):
sgtm, but potentially a minor break?

fvm/src/kernel/mod.rs
Stebalien: Ah, you're right. I'll make a note of it but I doubt it'll impact anyone.

PR #2027, opened 2024-07-23
 ci: automate creation of GitHub releases and tags

galargh:
This automates the creation of GitHub releases and associated GitHub tags based on the changes to the version in the selected Cargo.toml files.

The automation uses the same principles as the one you might know from Go projects where we control the current version via version.json file (https://github.com/ipdxco/unified-github-workflows?tab=readme-ov-file#versioning). This is the exact same workflow just updated to work with Cargo.toml.

~~Note, that the automation doesn't take care of cargo publishing currently. It doesn't create `crate_name@crate_version` tags either. To handle these, I'd suggest creating another workflow triggered on tag creation or github release publishing, which creates additional tags and performs cargo publishing.~~

The automation takes care of publishing `crate_name@crate_version` packages. It also creates GitHub releases for each tag.

### DIFF
--- a/.github/workflows/release-check.yml
+++ b/.github/workflows/release-check.yml
@@ -1,0 +1,21 @@
+name: Release Checker
+
+on:
+  pull_request_target:
+    paths: ["Cargo.toml"]
+    types: [ opened, synchronize, reopened, labeled, unlabeled ]
+  workflow_dispatch:
+
+permissions:
+  contents: write
+  pull-requests: write
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  release-check:
+    uses: ipdxco/unified-github-workflows/.github/workflows/release-check.yml@v1.0
+    with:
+      sources: '["Cargo.toml"]'
--- a/.github/workflows/release-check.yml
+++ b/.github/workflows/release-check.yml
@@ -3,18 +3,7 @@ name: Release Checker
 on:
   pull_request_target:
     paths:
-      - "Cargo.toml"
-      - "fvm/Cargo.toml"
-      - "testing/integration/Cargo.toml"
-      - "ipld/amt/Cargo.toml"
-      - "ipld/bitfield/Cargo.toml"
-      - "ipld/blockstore/Cargo.toml"
-      - "ipld/car/Cargo.toml"
-      - "ipld/encoding/Cargo.toml"
-      - "ipld/hamt/Cargo.toml"
-      - "ipld/kamt/Cargo.toml"
-      - "sdk/Cargo.toml"
-      - "shared/Cargo.toml"
+      - "**/Cargo.toml"
     types: [ opened, synchronize, reopened, labeled, unlabeled ]
   workflow_dispatch:
 
--- a/.github/workflows/release-check.yml
+++ b/.github/workflows/release-check.yml
@@ -45,9 +45,9 @@ jobs:
           "shared/Cargo.toml"
         ]
       separator: "@"
-  cargo-publish-dry-run:
+  cargo-publish:
     needs: [release-check]
-    if: toJSON(fromJSON(needs.release-check.outputs.json)) != '[]'
+    if: needs.release-check.outputs.json != '[]'
     runs-on: ubuntu-latest
     strategy:
       fail-fast: false
--- a/.github/workflows/release-check.yml
+++ b/.github/workflows/release-check.yml
@@ -2,7 +2,18 @@ name: Release Checker
 
 on:
   pull_request_target:
-    paths: ["Cargo.toml"]
+    paths:
+      - "fvm/Cargo.toml"
+      - "testing/integration/Cargo.toml"
+      - "ipld/amt/Cargo.toml"
+      - "ipld/bitfield/Cargo.toml"
+      - "ipld/blockstore/Cargo.toml"
+      - "ipld/car/Cargo.toml"
+      - "ipld/encoding/Cargo.toml"
+      - "ipld/hamt/Cargo.toml"
+      - "ipld/kamt/Cargo.toml"
+      - "sdk/Cargo.toml"
+      - "shared/Cargo.toml"
     types: [ opened, synchronize, reopened, labeled, unlabeled ]
   workflow_dispatch:
 
@@ -18,4 +29,34 @@ jobs:
   release-check:
     uses: ipdxco/unified-github-workflows/.github/workflows/release-check.yml@v1.0
     with:
-      sources: '["Cargo.toml"]'
+      sources: |
+        [
+          "fvm/Cargo.toml",
+          "testing/integration/Cargo.toml",
+          "ipld/amt/Cargo.toml",
+          "ipld/bitfield/Cargo.toml",
+          "ipld/blockstore/Cargo.toml",
+          "ipld/car/Cargo.toml",
+          "ipld/encoding/Cargo.toml",
+          "ipld/hamt/Cargo.toml",
+          "ipld/kamt/Cargo.toml",
+          "sdk/Cargo.toml",
+          "shared/Cargo.toml"
+        ]
+      separator: "@"
+  cargo-publish-dry-run:
+    needs: [release-check]
+    if: toJSON(fromJSON(needs.release-check.outputs.json)) != '[]'
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        source: ${{ fromJSON(needs.release-check.outputs.json).*.source }}
+    steps:
+      - uses: actions/checkout@v4
+      - env:
+          source: ${{ matrix.source }}
+        run: |
+          pushd "$(dirname $source)"
+          cargo publish --dry-run
+          popd
--- a/.github/workflows/release-check.yml
+++ b/.github/workflows/release-check.yml
@@ -3,6 +3,7 @@ name: Release Checker
 on:
   pull_request_target:
     paths:
+      - "Cargo.toml"
       - "fvm/Cargo.toml"
       - "testing/integration/Cargo.toml"
       - "ipld/amt/Cargo.toml"
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -78,7 +78,7 @@ Once the release is prepared, it'll go through a review:
 
 Finally, an [FVM "owner"](https://github.com/orgs/filecoin-project/teams/fvm-crate-owners/members) will:
 
-1. Merge the release PR to master.
+1. Merge the release PR to master (this will trigger GitHub release/tag creation if the workspace version in the root `Cargo.toml` was modified).
 2. For each released crate, create a git tag: `crate_name@crate_version`.
 3. Run `cargo publish` for each released crate (in dependency order).
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -115,18 +115,21 @@ This section describes the automated parts of the release process and the manual
 
 #### Current State
 
-1. On a release pull request creation, a [Release Checker](.github/workflows/release-check.yml) workflow will run. It will perform the following actions:
-    1. Extract the version from the modified `Cargo.toml` files. Process each crate in the workspace **independently**.
+1. Create a pull request which updates the [`version` in one or more `Cargo.toml` files](https://github.com/search?q=repo%3Afilecoin-project%2Fref-fvm+path%3ACargo.toml+%2F%5Eversion+%3D%2F&type=code).
+   - Title the PR `chore: release vX.Y.Z`
+2. On such a release PR's creation, a [Release Checker](.github/workflows/release-check.yml) workflow will run. It will perform the following actions:
+    1. Extract the version from the modified `Cargo.toml` files, and process each crate in the workspace **independently**.
     2. Check if a git tag for the version, using the `crate_name@version` as the pattern, already exists. Continue only if it does not.
     3. Create a draft GitHub release with the version as the tag.
     4. Comment on the pull request with a link to the draft release.
     5. Run `cargo publish --dry-run` for the crate for which the release is proposed.
 2. On pull request merge, a [Releaser](.github/workflows/release.yml) workflow will run. It will perform the following actions:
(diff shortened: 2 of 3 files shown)


Stebalien:
Does this do a dry-run publish to prevent cutting a release if the eventual publish will fail (e.g., because we failed to release some dependency in the workspace)?

rvagg:
> dry-run publish

I imagine this will just do a github release draft publish like version.json workflow? I think @Stebalien is asking about doing a `cargo publish --dry-run`, which I don't imagine to be part of this and that we probably need a second workflow to do with the same trigger as this? Or can we just add a new step in .github/workflows/releaser.yml to do this?

galargh:
> I imagine this will just do a github release draft publish like version.json workflow? I think @Stebalien is asking about doing a cargo publish --dry-run, which I don't imagine to be part of this and that we probably need a second workflow to do with the same trigger as this? Or can we just add a new step in .github/workflows/releaser.yml to do this?

Yes, we can add this as a new job in the `releaser.yml` and/or `release-check.yml`, `web3-bot` won't overwrite it. I'll propose that in a comment to show you, how this could look like.

.github/workflows/release-check.yml
```diff
@@ -0,0 +1,21 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  release-check:
+    uses: ipdxco/unified-github-workflows/.github/workflows/release-check.yml@v1.0
+    with:
+      sources: '["Cargo.toml"]'
```
galargh, line 21: That way, we would run the cargo publish dry run, only if the release-check created a draft GitHub Release. Please note, that the output that I reference doesn't exist yet, but I think it's a great idea to expose it.

```suggestion
  release-check:
    uses: ipdxco/unified-github-workflows/.github/workflows/release-check.yml@v1.0
    with:
      sources: '["Cargo.toml"]'
  cargo-publish-dry-run:
    needs: [release-check]
    if: steps.release-check.outputs.release != ''
    runs-on: ubuntu-latest
    steps:
      - uses: actions/checkout@v4
      - run: cargo publish --dry-run
         shell: bash
```

rvagg: 👌 nice, this should be good; @Stebalien is this enough?

galargh: BTW, similarly, we could attach extra tag creation and cargo publishing to the releaser.yml.

Stebalien: This doesn't look like it handles multiple packages (we have multiple packages in the workspace). Honestly, I'm not actually sure how to dry-run with multiple packages.

See this thread: https://github.com/rust-lang/cargo/issues/1169. I was hoping that you might have found some magical solution...

(see https://github.com/rust-lang/cargo/issues/1169#issuecomment-1753461359)

galargh: OK, I think I'll get it now. I'm going to resolve this conversation and explain the changes I introduced in a comment below. Let me know what you think about them.

galargh:
Now, the introduced workflows will release packages defined by any of these `Cargo.toml`s:
```
fvm/Cargo.toml
testing/integration/Cargo.toml
ipld/amt/Cargo.toml
ipld/bitfield/Cargo.toml
ipld/blockstore/Cargo.toml
ipld/car/Cargo.toml
ipld/encoding/Cargo.toml
ipld/hamt/Cargo.toml
ipld/kamt/Cargo.toml
sdk/Cargo.toml
shared/Cargo.toml
```

The workflows will also be triggered on changes to the root `Cargo.toml` since some of the packages use the workspace version.

This is how we'll work out a version for a specific package:
1. If there is `workspace.package.version` or `package.version` or `version` defined in the config file, use it.
2. Otherwise, check `Cargo.toml` in the parent directory unless we were already checking the root directory.

As for the name of the package to use, we'll infer it as follows:
1. If there is `package.name` or `name` defined in the config file, use it.
2. Otherwise, use the relative path from root as the package name.

Now, let's say, we're trying to release package defined by `fvm/Cargo.toml`.
1. We modify the version root `Cargo.toml`
2. We create a PR
3. `release-check.yml` is triggered
4. The workflow finds the name of the package in `fvm/Cargo.toml` and the version to use in `Cargo.toml`
5. It creates a draft PR
6. It comments on the PR with a link to the draft
7. It recognizes that a draft for `fvm/Cargo.toml` was created so it proceeds to execute `cargo publish --dry-run` in the `fvm` directory.
8. We merge the PR
9. `release.yml` is triggered
10. The workflow, similarly, finds the name of the package in `fvm/Cargo.toml` and the version in `Cargo.toml`
11. It finds the draft release that was previously created
12. It publishes the draft release

I also added a separator input to the release workflows and specified it here as `@`. This means that in the example above, the package that would be released would be `fvm@vX.Y.Z`.

Let me know if this meets your expectations now.

Stebalien:
I'm going to wait for us to resolve all the conversations on https://github.com/filecoin-project/actors-utils/pull/235 first to avoid duplicating any work/discussions.

galargh:
I limited duplication here in https://github.com/filecoin-project/ref-fvm/pull/2027/commits/50273e3bec695e315ffb43f03adf21424fd9547f, similarly to changes requested in `actors-utils`. Let me know if there's anything else beyond this that you'd like to see in this PR.

CONTRIBUTING.md
```diff
@@ -78,7 +78,7 @@ Once the release is prepared, it'll go through a review:
 
 Finally, an [FVM "owner"](https://github.com/orgs/filecoin-project/teams/fvm-crate-owners/members) will:
 
-1. Merge the release PR to master.
+1. Merge the release PR to master (this will trigger GitHub release/tag creation if the workspace version in the root `Cargo.toml` was modified).
```
BigLep, line 81: Is this complete?  Are there steps from https://github.com/filecoin-project/ref-fvm/pull/2027#issuecomment-2254402698 we need to include?  I don't see discussion about opening a PR which creates a draft Release and that when we merge the PR we get a published Release.

And are the comments about an FVM owner still needing to run `cargo publish` accurate?  (or are we only doing a dryrun as part of CI?)

BigLep:
@Stebalien : I believe @galargh believes this is ready for review again.

I am personally confused about whether crate publishing has been automated or not but that might be my own current mental state.  I'm hopeful we get the readme updated to the new reality.

Stebalien:
https://github.com/filecoin-project/ref-fvm/pull/2031 should check the root crate. I think we need to match `/Cargo.toml` in addition to `**/Cargo.toml`.

.github/workflows/release-check.yml
```diff
@@ -0,0 +1,52 @@
...
+      fail-fast: false
+      matrix:
+        source: ${{ fromJSON(needs.release-check.outputs.json).*.source }}
+    steps:
+      - uses: actions/checkout@v4
+      - env:
+          source: ${{ matrix.source }}
+        run: |
+          pushd "$(dirname $source)"
+          cargo publish --dry-run
```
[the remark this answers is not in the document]
Stebalien: I.e., if we cut a release of the sdk + shared at the same time (which we always do), we'd need to push a release of shared to some fake registry for the release of the sdk to pass a `cargo publish` dry run (I think).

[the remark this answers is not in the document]
galargh: Ah, I see what you mean now! Yes, that would definitely show up as a failure in the proposed process. Would it make sense to split the release in that case, as in, you first release `shared` and only follow up with `sdk` release after `shared` is fully out?

galargh:
> #2031 should check the root crate. I think we need to match `/Cargo.toml` in addition to `**/Cargo.toml`.

`**/Cargo.toml` captures root `Cargo.toml` as well. It hasn't been triggered in your PR because we use the `pull_request_target` trigger which means that the workflow has to be defined in the PR target and we use that version of it.

.github/workflows/release-check.yml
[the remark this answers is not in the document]
Stebalien: That's going to make releases take even longer, which defeats the point of having automation.

I think the answer here is manual dry-run and it's not going to be pretty.

1. Figure out which crates need to be published.
2. Either determine the correct order, or hard-code it in a config.
3. Run `cargo vendor --versioned-dirs`, creating a `.cargo/config.toml` file to respect the vendor directory.
4. In publish order (reverse dependency order):
  1. Run `cargo package`.
  2. Copy the package (the extracted one) out of `target/packages` into `vendor`.
  3. Generate the `.cargo-checksum.json` file for that manually vendored package. This is the most annoying step, but shouldn't be hard with `find` + `jq`...

CONTRIBUTING.md
```diff
@@ -112,3 +108,45 @@ If that works, proceed with releasing these crates.
...
 [discuss]: https://github.com/filecoin-project/fvm-specs/discussions
 [issue]: https://github.com/filecoin-project/ref-fvm/issues
+
+### Release Process
+
+This section describes the automated parts of the release process and the manual steps FVM owners need to take.
+
+#### Current State
+
+1. On a release pull request creation, a [Release Checker](.github/workflows/release-check.yml) workflow will run. It will perform the following actions:
```
BigLep, line 118:
```suggestion
1. Create a pull request which updates the [`version` in one or more `Cargo.toml` files](https://github.com/search?q=repo%3Afilecoin-project%2Fref-fvm+path%3ACargo.toml+%2F%5Eversion+%3D%2F&type=code).
   - Title the PR `chore: release vX.Y.Z`
2. On such a release PR's creation, a [Release Checker](.github/workflows/release-check.yml) workflow will run. It will perform the following actions:
```

CONTRIBUTING.md
```diff
@@ -112,3 +108,45 @@ If that works, proceed with releasing these crates.
...
+    1. Extract the version from the modified `Cargo.toml` files. Process each crate in the workspace **independently**.
```
BigLep, line 119:
```suggestion
    1. Extract the version from the modified `Cargo.toml` files, and process each crate in the workspace **independently**.
```

CONTRIBUTING.md
```diff
@@ -112,3 +108,45 @@ If that works, proceed with releasing these crates.
...
+    2. Check if a git tag for the version, using the `crate_name@version` as the pattern, already exists. Continue only if it does not.
+    3. Create a draft GitHub release with the version as the tag.
+    4. Comment on the pull request with a link to the draft release.
+    5. Run `cargo publish --dry-run` for the crate for which the release is proposed.
```
BigLep, line 123: Is it intentional that it's automated here but manual in actors-utils? https://github.com/filecoin-project/actors-utils/pull/235/files#diff-2b1b69303b927a484e02c7fad9fc87d0d3ff0dc22ae1da0ecd0dc935d922a23cR12

CONTRIBUTING.md
```diff
@@ -112,3 +108,45 @@ If that works, proceed with releasing these crates.
...
+2. On pull request merge, a [Releaser](.github/workflows/release.yml) workflow will run. It will perform the following actions:
+    1. Extract the version from the modified `Cargo.toml` files. Process each crate in the workspace **independently**.
```
BigLep, line 125:
```suggestion
    1. Extract the version from the modified `Cargo.toml` files, and process each crate in the workspace **independently**.
```

CONTRIBUTING.md
```diff
@@ -112,3 +108,45 @@ If that works, proceed with releasing these crates.
...
+    2. Check if a git tag for the version, using the `crate_name@version` as the pattern, already exists. Continue only if it does not.
+    3. Check if a draft GitHub release with the version as the tag exists.
+    4. If the draft release exists, publish it. Otherwise, create a new release with the version as the tag.
+3. **[MANUAL]** Run `cargo publish` for each crate that has been released in the [reverse dependency order](#crate-dependency-graph).
```
BigLep, line 129:
```suggestion
3. **[MANUAL]** Run `cargo publish` for each crate that has been released in the [reverse dependency order](#crate-dependency-graph).
   - You will need to be part of [fvm-create-owners](https://github.com/orgs/filecoin-project/teams/fvm-crate-owners) to do this per https://crates.io/crates/fvm.
```

@Stebalien : this is my guess/understanding

BigLep (commented):
Thanks for adding the docs.  I'll leave for others to approve and make decisions here, but this still seems to me like it's moving the ball forward.

CONTRIBUTING.md
BigLep: PR adding who is part of fvm-crate-owners: https://github.com/filecoin-project/github-mgmt/pull/58

CONTRIBUTING.md
galargh: It was explicitly requested here - that's why it was added. But it's not clear to me whether we want it at all without being able to incorporate dependency graph knowledge into the dry run publish process.

BigLep:
Per verbal with @Stebalien, without crate publishing we don't think this workflow is good enough to move forward.  The reason is that in the manual workflow that a releaser like @Stebalien follows involves:
1. Merge the version changes into `main/master`
2. Attempt to publish crates.  There is some trial and error on publishing the crates at this point and sometimes discover that a dependency hasn't been published yet, which after fixing means bumping the version again.  
3. After publishing the crates successfully the release engineer then does the git tags.

The workflow proposed in this PR has git tagging happening before any publishing has occurred and there are reasons why a dryrun publish is not truly representative (I'm not up on these details).  As a result, this workflow could cause more work than before because you could be untagging in some cases.

It's a bummer not to get this across the line, but I know IPDX's contract hours with FilOz are up at this point.  As a result, I'm going to close this PR.  For next network upgrade, nv24, if the situation hasn't improved, we'll have someone like @rjan90 or @BigLep watch @Stebalien and capture the manual steps that are covered so we at least document the current manual release process better.  We'll also make sure that others have access to publish with https://github.com/filecoin-project/github-mgmt/pull/58 landed.

galargh:
Sad to hear we couldn't get it over the line in this iteration, but I think these are totally fair points. I hope we'll be able to work on these together in the future and get the full automation in place for both this repo and `actors-utils`. Thank you all for all the reviews and information sharing. Even though we didn't land this, I think that the back-and-forth that we did here is going to be very useful whenever one attempts to implement full release process automation.

Stebalien:
Yeah, the lack of a a viable `cargo publish --workspace --dry-run` makes all this really hard.

PR #2034, opened 2024-08-27
docs: remove any mentions of randomness lookback limits

Stebalien:
They don't exist.

### DIFF
--- a/sdk/src/sys/rand.rs
+++ b/sdk/src/sys/rand.rs
@@ -11,7 +11,8 @@ use crate::sys::ErrorNumber::*;
 super::fvm_syscalls! {
     module = "rand";
 
-    /// Gets 32 bytes of randomness from the ticket chain.
+    /// Gets 32 bytes of randomness from the ticket chain,
+    /// there is no lookback limit.
     ///
     /// # Arguments
     ///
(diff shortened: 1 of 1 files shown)


sdk/src/sys/rand.rs
```diff
@@ -11,34 +11,34 @@ use crate::sys::ErrorNumber::*;
...
     ///
     /// # Errors
     ///
-    /// | Error               | Reason                  |
-    /// |---------------------|-------------------------|
-    /// | [`LimitExceeded`]   | lookback exceeds limit. |
-    /// | [`IllegalArgument`] | invalid buffer, etc.    |
+    /// | Error               | Reason                 |
+    /// |---------------------|------------------------|
+    /// | [`IllegalArgument`] | epoch is in the future |
```
anorth, line 25: "epoch is not a valid prior epoch number"? e.g. negatives, too

anorth (approved):

sdk/src/sys/rand.rs
Stebalien: It'll just return randomness from epoch 0. Historically it hashed it with the epoch number (giving each epoch unique randomness), now we expect that to happen in the actors themselves.

I'll document that. Also, I'll document null tipset behavior (or try to... it changes network version to network version).

jennijuju (approved):

Stebalien:
Ideally it would include more extensive documentation around different  epochs, but we might as well merge it as-is as it's strictly an improvement.

PR #2035, opened 2024-09-10
chore: update wasmtime to v24

Stebalien:


Stebalien:
Ok, this appears to work: https://github.com/filecoin-project/lotus/pull/12449

fvm/src/engine/mod.rs
```diff
@@ -164,6 +166,7 @@ fn wasmtime_config(ec: &EngineConfig) -> anyhow::Result<wasmtime::Config> {
 
     // wasmtime default: true
     // We don't want threads, there is no way to ensure determinism
+    #[cfg(feature = "wasmtime/threads")]
     c.wasm_threads(false);
```
rvagg, line 170: should we just remove the feature then? or are you wanting to keep this just to be explicit about it for if it eventually shows up not behind a feature flag?

rvagg: ah, it's on by default anyway so always tehre

rvagg:
line 180 of fvm/src/engine/mod.rs, `wasm_tail_call` is default is now `true`, just needs a doc update

rvagg (approved):

PR #2036, opened 2024-09-11
fix: use unix signal handlers instead of Mach ports

Stebalien:
This will apparently interfere with some debugging tools, but we ran into an issue where Mach ports weren't always catching illegal instruction signals form wasmtime (affecting macos only).


rvagg (approved):

fvm/src/engine/mod.rs
```diff
@@ -247,6 +247,8 @@ fn wasmtime_config(ec: &EngineConfig) -> anyhow::Result<wasmtime::Config> {
     // todo(M2): make sure this is guaranteed to run in linear time.
     c.cranelift_opt_level(Speed);
```
rvagg, line 249: document

rvagg (approved):

PR #2037, opened 2024-09-12
[v3] fix: use unix signal handlers instead of Mach ports

Stebalien:
This will apparently interfere with some debugging tools, but we ran into an issue where Mach ports weren't always catching illegal instruction signals form wasmtime (affecting macos only).


Stebalien:
Need to update wasmtime first.

rvagg (approved):

Stebalien:
Requires #2039.

PR #2041, opened 2024-09-12
[v2] update wasmtime from v12 to v24

Stebalien:
* chore: use workspace dependencies for wasmtime
* chore: update wasmtime from v12 to v24


Stebalien:
Merging without review because this is just a backport to a very old FVM version.

PR #2049, opened 2024-09-17
Set epoch manually

ukstv:
What if I want to manually jump through the epochs?

### DIFF
--- a/fvm/src/machine/mod.rs
+++ b/fvm/src/machine/mod.rs
@@ -249,6 +249,12 @@ impl MachineContext {
         self
     }
 
+    /// Sets [`MachineContext::epoch`].
+    pub fn set_epoch(&mut self, epoch: ChainEpoch) -> &mut Self {
+        self.epoch = epoch;
+        self
+    }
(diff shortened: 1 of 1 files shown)


ukstv:
Based on some comments, I can deduce recreating the whole tree of dependencies that includes bunch of built-in actors and whatnot is damn costly to do every time. And, to be honest, this parameter change is much less code than recreation of the whole dependency tree.

Stebalien:
Re-creating the `Engine` is expensive, but you only need one engine per `NetworkConfig`.

- `NetworkConfig` specifies the network-wide options. Requires a new engine because different network-wide options, e.g., different gas costs, may require re-compiling the actors (very expensive).
- `MachineConfig` is created from the `NetworkConfig` with `for_epoch`. It specifies the "machine" state on which a specific epoch should be executed. It's very cheap. Actually re-loading the state-tree will take some time, but you're usually going to want to do that every time you switch epochs anyways.

Stebalien (approved):

PR #2053, opened 2024-10-04
chore: release v4.3.3

rjan90:
### Context: 

We do not want to pull in the new wasmtime versions for the Lotus v1.30.0 release, just to be conservative because Lotus v1.30.0 will be a mandatory network version.

I have therefore:
- Created a `release/4.3` branch which was branched off the [fvm@v4.3.2](https://github.com/filecoin-project/ref-fvm/tree/fvm%40v4.3.2) tag.
- **In this PR:** Cherry-picked https://github.com/filecoin-project/ref-fvm/pull/2051, and bumped the version to v4.3.3, added changelog entries and updated the cargo.lock.

I looked through other things that have landed in master since v4.3.2 was tagged, but I could not see anything else that was directly needed for the nv24 upgrade. But it would be good if someone else also double confirmed

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -725,7 +725,7 @@ dependencies = [
  "cid 0.10.1",
  "fvm_ipld_bitfield 0.6.0",
  "fvm_ipld_encoding 0.4.0",
- "fvm_shared 4.3.2",
+ "fvm_shared 4.3.3",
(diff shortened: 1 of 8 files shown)


Stebalien (approved):

Stebalien:
Tagged and published.

PR #2055, opened 2024-10-09
Release 4.4.2

Stebalien:
- Update wasmtime
- Update other deps

I've gone ahead and updated the rest of the deps as well (no major releases, just semver compatible versions) because there was a slew of other wasmtime and related crates that didn't appear to be automatically pulled in otherwise.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4,11 +4,11 @@ version = 3
 
 [[package]]
 name = "addr2line"
-version = "0.24.1"
+version = "0.24.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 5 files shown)


Cargo.lock
```diff
@@ -5590,7 +5590,7 @@ version = "0.1.9"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "cf221c93e13a30d793f7645a0e7762c55d169dbb0a49671918a2319d289b10bb"
 dependencies = [
- "windows-sys 0.59.0",
+ "windows-sys 0.52.0",
```
Stebalien, line 5593: :man_shrugging:

rjan90 (approved):

PR #2059, opened 2024-10-18
feat: wasmtime v25

### LINKED ISSUE #2058: [Investigation] Reported performance regression when updating wasmtime from 19 to 24.
We've seen a reported performance regression that appears to be associated with the wasmtime update that happened in FVM 4.4 (wasmtime 19 -> 24). This was an interesting performance regression because:

1. It shows up even when updating FVM 2.7 to 2.9 even when wasmtime 2.9 isn't being _used_, just included. This indicates that the issue is in some common wasmtime dependency.
2. None of the gas benchmarks in FVM v4 are picking up the regression, but those benchmarks mostly focus on non-wasm things. Importantly, I'm reasonably sure that:
    1. It's not related to hashing.
    2. It's not related to signature verification.
    3. It's not related to calling into wasm contracts (see #2057).
    4. It's not related to trapping and running out of gas (we have an explicit benchmark for this because it ocasionally rears its head and turns into an $N^2$ problem).
    5. It shouldn't be cranelift. That tends to be version-locked with a specific wasmtime version.

Things I haven't checked:

1. The cost of wasm instructions in general. We priced wasm instructions long ago and it was a bit of a manual process.
2. The cost of jumping in and out of wasm. Wasmtime has made some recent changes here _however_, I'd expect any issues there to be specific to the wasmtime version in-use (observation (1) above appears to conflict with this).

All I can think of now is... rustix? Or something like that?

Stebalien:
Importantly, this reverts a previous wasmtime change that enabled trace-logging in regalloc, massively slowing down compilation across all FVM versions.

fixes #2058

### DIFF
--- a/fvm/src/engine/mod.rs
+++ b/fvm/src/engine/mod.rs
@@ -259,6 +259,16 @@ fn wasmtime_config(ec: &EngineConfig) -> anyhow::Result<wasmtime::Config> {
     // debugging tools. But we'll just have to live with that.
     c.macos_use_mach_ports(false);
 
+    // wasmtime default: true
+    // Disable extended const support. We'll probably enable this in the future but that requires a
+    // FIP.
+    c.wasm_extended_const(false);
+
+    // wasmtime default: false
+    // Disable the component module.
+    #[cfg(feature = "wasmtime/component-model")]
+    c.wasm_component_model(false);
+
     Ok(c)
 }
 
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -887,18 +887,18 @@ dependencies = [
 
 [[package]]
 name = "cranelift-bforest"
-version = "0.111.1"
+version = "0.112.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "32d69b774780246008783a75edfb943eccc2487b6a43808503a07cd563f2ffde"
+checksum = "7b765ed4349e66bedd9b88c7691da42e24c7f62067a6be17ddffa949367b6e17"
 dependencies = [
  "cranelift-entity",
 ]
 
 [[package]]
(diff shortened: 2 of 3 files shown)


fvm/src/engine/mod.rs
```diff
@@ -259,6 +259,16 @@ fn wasmtime_config(ec: &EngineConfig) -> anyhow::Result<wasmtime::Config> {
     // debugging tools. But we'll just have to live with that.
     c.macos_use_mach_ports(false);
 
+    // wasmtime default: true
+    // Disable extended const support. We'll probably enable this in the future but that requires a
+    // FIP.
+    c.wasm_extended_const(false);
```
Stebalien, line 265: This is new, but all I'm doing here is disabling support for wasm modules that use this feature.

fvm/src/engine/mod.rs
```diff
@@ -259,6 +259,16 @@ fn wasmtime_config(ec: &EngineConfig) -> anyhow::Result<wasmtime::Config> {
...
+
+    // wasmtime default: false
+    // Disable the component module.
+    #[cfg(feature = "wasmtime/component-model")]
+    c.wasm_component_model(false);
```
Stebalien, line 270: This isn't new in this version, I'm just explicitly disabling it (it should already be disabled).

Stebalien:
(going to write a benchmark now to confirm that this is fixed)

Stebalien:
Benchmark in https://github.com/filecoin-project/ref-fvm/pull/2060, but I'm not seeing any compilation time changes between this version, v4.4, and v4.3.

LesnyRumcajs (approved):
Thank you for addressing that so quickly!

PR #2060, opened 2024-10-18
Add an actor compilation benchmark

Stebalien:
Specifically, it benchmarks compiling wasm -> native via cranelift.

### DIFF
--- a/testing/integration/benches/compile.rs
+++ b/testing/integration/benches/compile.rs
@@ -1,0 +1,46 @@
+// Copyright 2021-2023 Protocol Labs
+// SPDX-License-Identifier: Apache-2.0, MIT
+use std::hint::black_box;
+use std::time::Duration;
+
+use cid::Cid;
+use criterion::{criterion_group, criterion_main, Criterion};
+use fvm::engine::EnginePool;
+use fvm::machine::{Manifest, NetworkConfig};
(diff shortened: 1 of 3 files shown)


testing/integration/benches/compile.rs
```diff
@@ -0,0 +1,46 @@
...
+use criterion::{criterion_group, criterion_main, Criterion};
+use fvm::engine::EnginePool;
+use fvm::machine::{Manifest, NetworkConfig};
+
+use fvm_integration_tests::bundle;
+use fvm_ipld_blockstore::MemoryBlockstore;
+use fvm_ipld_encoding::CborStore;
+use fvm_shared::version::NetworkVersion;
+
+fn bench_compile(c: &mut Criterion) {
```
Stebalien, line 16: So, I think this benchmark is correct but... I'm not able to reproduce #2058.

LesnyRumcajs: Can it be that the slowdown occurs under specific actor version(s)?

LesnyRumcajs: I also didn't manage to reproduce it with this benchmark on a machine that showed a significant slowdown (used this benchmark on this branch and tag @fvm@v4.3.3

PR #2061, opened 2024-10-18
test: use the correct wasm profile

Stebalien:
And update the integer overflow tests to match. It also gives us better test coverage of panics/traps.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -87,20 +87,13 @@ fvm_integration_tests = { path = "testing/integration", version = "~4.4.1" }
 fvm_gas_calibration_shared = { path = "testing/calibration/shared" }
 fvm_test_actors = { path = "testing/test_actors" }
 
-[profile.actor]
-inherits = "release"
-panic = "abort"
-overflow-checks = true
-lto = true
-opt-level = "z"
-#strip = true
-
+# Same as in the built-in actors repo
 [profile.wasm]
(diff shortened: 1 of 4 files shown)


Cargo.toml
```diff
@@ -87,20 +87,13 @@ fvm_integration_tests = { path = "testing/integration", version = "~4.4.1" }
...
-#strip = true
-
+# Same as in the built-in actors repo
 [profile.wasm]
 inherits = "release"
-panic = "abort"
-overflow-checks = false
-lto = true
-opt-level = "z"
+panic = "unwind"
```
Stebalien, line 93: We allow panicing in wasm not because we catch panics (rust doesn't support that) but because we register a panic hook that exits with a nice exit code and error message.

testing/integration/tests/fil_integer_overflow.rs
```diff
@@ -60,94 +60,150 @@ fn integer_overflow() {
     // Instantiate machine
     tester.instantiate_machine(DummyExterns).unwrap();
```
Stebalien, line 61: This test is mostly the same except that I split the overflow case into two:

1. Case 1: rely on overflow checks and expect the contract to panic/revert.
2. Case 2: explicitly specify wrapping semantics and make sure it the math works out.

masih (approved):

PR #2067, opened 2024-10-21
Script license header check and update in rust

masih:
The previous license header update, implemented in bash, would duplicate the headers if run. The rules defined in that script also seem to have been inconsistently applied. Having said that it is not clear what the desired license header rule for this repo is.

The changes here implement the license check and update in rust. It runs far faster, is more maintainable, does not duplicate the license header if already present and would add it if it is missing.

The rules programmed in the rust implementation are specifically picked to reduce the churn in code, and hopefully by the grace of reviewers, learn what the intended rules for this repo should be.


scripts/check-license/Cargo.toml
```diff
@@ -0,0 +1,12 @@
+[package]
+name = "check-license"
+version.workspace = true
```
Stebalien, line 3: I'd use a separate version here and specify "publish = false".

Stebalien:
> The previous license header update, implemented in bash, would duplicate the headers if run. The rules defined in that script also seem to have been inconsistently applied. Having said that it is not clear what the desired license header rule for this repo is.

I think the idea is:

1. In general, make sure we have an SPDX line for all checked-in rust files.
2. If we touch it, add a "copyright PL" line.

The rules are mostly consistent for new files, although they can end up duplicating lines on occasion.

> The changes here implement the license check and update in rust. It runs far faster, is more maintainable, does not duplicate the license header if already present and would add it if it is missing.

I'm concerned that in this case, especially because this is _rust_, the cost of compiling likely exceeds the cost of running them (unless already cached). Even with the cache, It's only _slightly_ faster to use rust and wouldn't even compare with a little bit of bash optimization.

scripts/check-license/src/main.rs
```diff
@@ -0,0 +1,105 @@
...
+
+    if !missing_licenses.is_empty() {
+        println!("Adding license header to: {}", path.display());
+
+        let file_content = fs::read_to_string(path)?;
+        let mut file = OpenOptions::new().write(true).truncate(true).open(path)?;
+
+        for license in missing_licenses {
+            file.write_all(license.as_bytes())?;
+            file.write_all(b"\n")?;
```
Stebalien, line 98: I'd usually write this as:

```rust
writeln!(&file, "{license}")?;
```

String formatting uses compiler magic to do most of the work at compile time, so it'll be very fast.

scripts/check-license/src/main.rs
```diff
@@ -0,0 +1,105 @@
...
+        if !found_match && *is_required {
+            missing_licenses.push(text_to_add);
+        }
+    }
+
+    if !missing_licenses.is_empty() {
+        println!("Adding license header to: {}", path.display());
+
+        let file_content = fs::read_to_string(path)?;
+        let mut file = OpenOptions::new().write(true).truncate(true).open(path)?;
```
Stebalien, line 94: note: this doesn't _actually_ need to be mutable because `Write` is implemented for `&File` (a trick to allow multiple concurrent writers without locking).

scripts/check-license/src/main.rs
```diff
@@ -0,0 +1,105 @@
...
+    let file = fs::File::open(path)?;
+    let reader = BufReader::new(file);
+    let lines: Vec<String> = reader
+        .lines()
+        .take(HEADER_LINES_TO_CHECK)
+        .filter_map(Result::ok)
+        .collect();
+
+    let mut missing_licenses: Vec<&str> = Vec::new();
+    for (regex, is_required, text_to_add) in LICENSE_CHECKS.iter() {
```
Stebalien, line 76:
```suggestion
    for (regex, is_required, text_to_add) in &LICENSE_CHECKS {
```

(does the same thing)

scripts/check-license/src/main.rs
```diff
@@ -0,0 +1,105 @@
...
+        let mut found_match = false;
+        for line in &lines {
+            if regex.is_match(line) {
+                found_match = true;
+                break;
+            }
+        }
+
+        if !found_match && *is_required {
```
Stebalien, line 85: why not skip early if not required?

scripts/check-license/src/main.rs
```diff
@@ -0,0 +1,105 @@
...
+
+    let mut missing_licenses: Vec<&str> = Vec::new();
+    for (regex, is_required, text_to_add) in LICENSE_CHECKS.iter() {
+        let mut found_match = false;
+        for line in &lines {
+            if regex.is_match(line) {
+                found_match = true;
+                break;
+            }
+        }
```
Stebalien, line 83:
```rust
let found_match = lines.iter().find(|&l|regex.is_match(l)).is_some()
```

scripts/check-license/src/main.rs
```diff
@@ -0,0 +1,105 @@
...
+}
+
+fn check_and_add_license_to_file(path: &Path) -> io::Result<()> {
+    let file = fs::File::open(path)?;
+    let reader = BufReader::new(file);
+    let lines: Vec<String> = reader
+        .lines()
+        .take(HEADER_LINES_TO_CHECK)
+        .filter_map(Result::ok)
+        .collect();
```
Stebalien, line 73: Note: usually, when an iterator returns a result, the first failure means we've failed to read the file (and should abort).

In this case, I'd write:

```suggestion
    let lines: Vec<String> = reader
        .lines()
        .take(HEADER_LINES_TO_CHECK)
        .try_collect()?;
```

Except that it's a nightly-only API right now so I'd import https://crates.io/crates/itertools.

scripts/check-license/src/main.rs
```diff
@@ -0,0 +1,105 @@
...
+
+    let mut missing_licenses: Vec<&str> = Vec::new();
```
Stebalien, line 75: I'd personally write this as a filter map.

scripts/check-license/src/main.rs
```diff
@@ -0,0 +1,105 @@
...
+
+        if !found_match && *is_required {
+            missing_licenses.push(text_to_add);
+        }
+    }
+
+    if !missing_licenses.is_empty() {
+        println!("Adding license header to: {}", path.display());
+
+        let file_content = fs::read_to_string(path)?;
```
Stebalien, line 93: nit: if you're just going to write it back to a file, you might as well read it as bytes.

scripts/check-license/src/main.rs
```diff
@@ -0,0 +1,105 @@
...
+        let mut file = OpenOptions::new().write(true).truncate(true).open(path)?;
+
+        for license in missing_licenses {
+            file.write_all(license.as_bytes())?;
+            file.write_all(b"\n")?;
+        }
+
+        file.write_all(file_content.as_bytes())?;
```
Stebalien, line 101: General note: there's no need to mimic `cat` here. You _can_ just:

1. Open the file.
2. Write at the beginning.
3. Close it.

scripts/check-license/src/main.rs
```diff
@@ -0,0 +1,105 @@
...
+        let path = entry.path();
+
+        if IGNORED_DIRECTORIES
+            .iter()
+            .any(|ignored_dir| path.starts_with(ignored_dir))
+        {
+            continue;
+        }
+
+        if path.is_file() && path.extension().and_then(|s| s.to_str()) == Some("rs") {
```
Stebalien, line 59:
```suggestion
        if path.is_file() && path.extension().map(|e| e == "rs").unwrap_or(false) {
```

(slightly shorter and `&OsStr == &str` is valid, IIRC)

scripts/check-license/src/main.rs
```diff
@@ -0,0 +1,105 @@
...
+        set.insert(PathBuf::from("./target/"));
+        set
+    };
+}
+
+/// This program expects the target directory as the first command-line argument. It recursively
+/// checks all `.rs` files within the specified directory, skipping any directories that are listed
+/// in the `IGNORED_DIRECTORIES` set. If a required license is missing, it is automatically added
+/// to the beginning of the file.
+fn main() {
```
Stebalien, line 39: Note: main can return errors.

scripts/check-license/src/main.rs
```diff
@@ -0,0 +1,105 @@
...
+    static ref LICENSE_CHECKS: Vec<(&'static Regex, bool, &'static str)> = vec![
+        (&PL_LICENSE, false, "// Copyright 2021-2024 Protocol Labs"),
+        (&CS_LICENSE, false, "// Copyright 2019-2024 ChainSafe Systems"),
+        (&SPDX_LICENSE, true, "// SPDX-License-Identifier: Apache-2.0, MIT"),
+    ];
+
+    /// IGNORED_DIRECTORIES is a static set containing directories to ignore during the license check.
+    static ref IGNORED_DIRECTORIES: HashSet<PathBuf> = {
+        let mut set = HashSet::new();
+        set.insert(PathBuf::from("./target/"));
```
Stebalien, line 30:
```suggestion
        set.insert("./target/".into());
```

Rust is all about the magic (when it works).

scripts/check-license/src/main.rs
Stebalien: Although, if you want performance a bufwriter might be in order.

scripts/check-license/src/main.rs
Stebalien: Also, it looks like this is being used as a vec.

Stebalien (commented):
Random rust comments.

masih:
Thanks Steven, I have got more than I wanted out of this PR. I'm specifically grateful for random rust comments 🍻

PR #2068, opened 2024-10-21
chore: optimize the license script

Stebalien:
This script is still a little awkweard, but it's now 30x faster.


Stebalien:
Ok, this version takes a few liberties...

Stebalien:
Ah... that's better. Now it's cryptic enough that nobody can even tell if it's wrong!

Stebalien:
not worth it

PR #2080, opened 2024-11-20
feat(ipld): undeprecate `for_each` methods for hamt/kamt

Stebalien:
These are useful because they handle some error cases and the ranged version is especially useful as using iter + skip etc. is actually kind of annoying. It would still be better to replace these with iterators in many cases, but replacing them _everywhere_ is annoying and not worth the effort.


Stebalien:
The _downside_ of keeping these functions is that we won't be able to remove the anyhow errors.

PR #2082, opened 2024-11-20
chore: release fvm_ipld_amt@v0.7.2

Stebalien:


Stebalien:
The last patch mistakenly released the HAMT, not the AMT. Having three structures ending in AMT is... annoying.

PR #2083, opened 2024-11-20
fix(amt): saturate the index at MAX_INDEX+1

Stebalien:
When we hit the end of the AMT we'll update the "key" to where the next tree should start. Unfortunately, if we've filled in the high-end of the AMT, this key could overflow the max u64 by exactly one (i.e., it'll wrap to 0). So, instead, we saturate to the MAX_INDEX (which is `u64-1`).

### DIFF
--- a/ipld/amt/src/iter.rs
+++ b/ipld/amt/src/iter.rs
@@ -209,7 +209,7 @@ where
             let stack = self.stack.last_mut()?;
             match stack.node {
                 Node::Leaf { vals } => {
-                    while stack.idx < vals.len() {
+                    while stack.idx < vals.len() && self.key <= MAX_INDEX {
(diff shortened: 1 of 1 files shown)


ipld/amt/src/iter.rs
```diff
@@ -281,6 +281,41 @@ mod tests {
     use fvm_ipld_encoding::BytesDe;
     use quickcheck_macros::quickcheck;
 
+    // Check for regressions around checked additions when iterating over high-keys.
+    #[test]
+    fn check_iter_high_keys() {
```
Stebalien, line 286: Note: I've checked and the two sections below test both cases above.

rvagg (approved):

PR #2090, opened 2024-11-22
chore: update cargo lockfile

Stebalien:
This only really matters for tests as upstream crates will check-in their own lockfiles, but it's nice to see what changed (and, e.g., verify that old cid/multihash versions have been cleared from the dependency tree).


rvagg (approved):
🤞

PR #2092, opened 2024-12-04
feat: kamt: Add public fn clear to reset and clear all entries in KAMT

snissn:
This PR introduces a new `clear` method to the `Kamt` data structure, allowing users to reset the root to an empty node and clear all entries. 

#### Changes:
- Added `pub fn clear` to `Kamt` to reset and clear all entries.
- Updated the test suite with `test_clear` to verify:
  - Entries exist before clearing.
  - Entries are removed after calling `clear`.
  - The `Kamt` remains functional post-clear.

This addition enhances usability by providing an efficient way to clear the `Kamt`. 

This functionality is introduced to support the implementation of Transient Storage (EIP-1153) in `builtin-actors`, enabling efficient clearing of the KAMT while adhering to Rust's RAII expectations and best practices. The previous approach attempts of using `take`/`delete` followed by `new` was unsuitable in Rust due to its ownership model and memory safety guarantees. The new `clear` method resolves this limitation, providing a safe and efficient way to reset the KAMT without requiring a full reallocation.

### DIFF
--- a/ipld/kamt/src/kamt.rs
+++ b/ipld/kamt/src/kamt.rs
@@ -571,13 +571,13 @@ mod tests {
         assert_eq!(kamt.get(&2).unwrap(), Some(&"b".to_string()));
 
         // Verify the KAMT is not empty
-        assert_eq!(kamt.is_empty(), false);
+        assert!(!kamt.is_empty());
 
         // Clear the KAMT
         kamt.clear();
 
         // Verify the KAMT is empty
-        assert_eq!(kamt.is_empty(), true);
+        assert!(kamt.is_empty());
 
         // Verify previous entries are gone
         assert_eq!(kamt.get(&1).unwrap(), None);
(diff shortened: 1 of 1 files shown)


ipld/kamt/src/kamt.rs
```diff
@@ -130,6 +130,12 @@ where
     pub fn is_empty(&self) -> bool {
         self.root.is_empty()
     }
+
+    /// Clears all entries in the KAMT and resets the root to an empty node.
+    pub fn clear(&mut self) {
+        self.root = Node::default(); // Reset the root to an empty node
+        self.flushed_cid = None; // Invalidate the flushed CID
```
Stebalien, line 137: Nit: can we check `is_empty` first to avoid writing the root if we're already empty?

Stebalien (approved):
One nit but otherwise this seems reasonable.

Also, I'd love to have the same method on the HAMT/AMT.

ipld/kamt/src/kamt.rs
snissn: updated 04fee90f

Stebalien (approved):

PR #2095, opened 2024-12-04
feat: hamt: Add public fn clear to reset and clear all entries in HAMT

snissn:
following up on changes to the KAMT in #2092 this PR changes HAMT to support a clear() method

This PR introduces a new `clear` method to the `Hamt` data structure, allowing users to reset the root to an empty node and clear all entries. 

#### Changes:
- Added `pub fn clear` to `Hamt` to reset and clear all entries.
- Updated the test suite with `test_clear` to verify:
  - Entries exist before clearing.
  - Entries are removed after calling `clear`.
  - The `Hamt` remains functional post-clear.

This addition enhances usability by providing an efficient way to clear the `Hamt`.

### DIFF
--- a/ipld/hamt/src/hamt.rs
+++ b/ipld/hamt/src/hamt.rs
@@ -336,6 +336,17 @@ where
         self.root.is_empty()
     }
 
+    /// Clears all entries in the HAMT and resets the root to an empty node.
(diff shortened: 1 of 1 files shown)


Stebalien (approved):
Thanks!

PR #2098, opened 2024-12-10
move the actor logs into the trace

### LINKED ISSUE #2097: Move the actor logs into the trace
At the moment, when actor debugging is enabled logs are written to [stdout](https://github.com/filecoin-project/ref-fvm/blob/ed6d34abae49257f15f226699e36dd51e00f2e4d/fvm/src/kernel/default.rs#L955). Instead, we should record them in the execution trace inside of the call manager. I.e., each log becomes an `ExecutionEvent`:

https://github.com/filecoin-project/ref-fvm/blob/ed6d34abae49257f15f226699e36dd51e00f2e4d/fvm/src/trace/mod.rs#L21-L40

If we do this, it should become _much_ easier to debug actors from clients. Instead of having to:

1. Restart the client with some env var enabled.
2. Read stdout/stderr.

We can invoke the FVM with debug tracing enabled and return the serialized logs over RPC.

qingyunha:
close #2097

### DIFF
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -513,6 +513,10 @@ where
 
         Ok(())
     }
+
+    fn log(&mut self, msg: String) {
+        self.trace(ExecutionEvent::Log(msg))
+    }
 }
 
 impl<M> DefaultCallManager<M>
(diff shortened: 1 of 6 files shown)


Stebalien (approved):

Stebalien:
Thanks!

PR #2099, opened 2025-01-07
chore: remove the hash-utils dependency from the KAMT

Stebalien:
We still need to do this for the HAMT, but this is progress.

### DIFF
--- a/ipld/kamt/src/id.rs
+++ b/ipld/kamt/src/id.rs
@@ -2,8 +2,6 @@
 // SPDX-License-Identifier: Apache-2.0, MIT
 use std::borrow::Cow;
 
-use forest_hash_utils::Hash;
-
 use crate::{AsHashedKey, HashedKey};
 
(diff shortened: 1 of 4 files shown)


ipld/kamt/src/id.rs
```diff
@@ -30,7 +28,10 @@ macro_rules! identity_hash {
         $(
             impl AsHashedKey<$t, 32> for Identity {
                 fn as_hashed_key(key: &$t) -> Cow<HashedKey<32>> {
-                    Cow::Owned(IdentityHasher::hash(key))
+                    const BYTES: usize = <$t>::BITS as usize / 8;
+                    let mut output = [0u8; 32];
+                    output[..BYTES].copy_from_slice(&<$t>::to_ne_bytes(*key));
```
Stebalien, line 33: This is the same encoding used by the hash-utils library.

masih (approved):

PR #2100, opened 2025-01-10
Add reserved exit codes and remove commented-out ones

elmattic:
This PR exposes reserved exit codes and removes others that are likely unnecessary.

The rationale is that for certain endpoints in the Ethereum RPC-API (e.g., `Filecoin.EthTraceBlock`), aligning Forest's JSON output with Lotus's output is important.

Having these exit codes available will help in the `forest-tool api compare` tool.

See [go-state-types exit codes](https://github.com/filecoin-project/go-state-types/blob/master/exitcode/names.go#L16-L21) and [traceErrMsg](https://github.com/filecoin-project/lotus/blob/v1.31.0/node/impl/full/eth_trace.go#L123) that we want to match in Forest.

### DIFF
--- a/shared/src/error/mod.rs
+++ b/shared/src/error/mod.rs
@@ -61,7 +61,7 @@ impl ExitCode {
     /// - The sender's nonce nonce didn't match the message nonce.
     /// - The sender didn't have the funds to cover the message gas.
     pub const SYS_SENDER_STATE_INVALID: ExitCode = ExitCode::new(2);
-    //pub const SYS_RESERVED_3 ExitCode = ExitCode::new(3);
+    pub const SYS_RESERVED_1: ExitCode = ExitCode::new(3);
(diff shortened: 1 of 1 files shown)


Stebalien:
Really, these shouldn't have been exported anywhere. Are you actually seeing these error codes?

Stebalien:
If you are running into them in practice, we need to rename those to something else like (deprecated X).

elmattic:
> Are you actually seeing these error codes?

I believe I've seen some of them while testing `Filecoin.EthTraceBlock` with Lotus on Calibnet.

I'm good if we don't defined them. Let's close this PR then.

Stebalien:
SGTM. If you do notice one, please report a bug and we'll rename it to something sane (i.e., "deprecated X" instead of "reserved X").

Issue #1994, opened 2024-02-20
Expose intermediate state


Stebalien:
At the moment, the FVM _discards_ intermediate state and only flushes the final state-tree. However, for debugging (tracing, etc.), intermediate state can be useful.

We should provide a way to flush intermediate state. Something like:

1. A flag that causes intermediate state to be flushed. If we do it this way, we'll have to be careful to only use this flag when we've replaced the blockstore with a temporary blockstore.
2. A _second_ blockstore for intermediate state (along with, possibly, a flag to actually use it).

rjan90:
This is no longer a pre-requiste for https://github.com/filecoin-project/lotus/issues/11635. And it is not clear if there is a big request for exposing intermediate state from our users.

rvagg:
https://github.com/filecoin-project/ref-fvm/pull/2101

PR #2101, opened 2025-01-11
feat: add flush_all_blocks option to also write intermediate blocks

rvagg:


rvagg:
https://github.com/filecoin-project/ref-fvm/issues/1994

Stebalien:
This seems like a reasonable approach but I'm not super happy adding this to the Executor/Machine traits. What about adding a "flush intermediate blocks to" option to either:

1. The `DefaultMachine::new` constructor. That is, instead of taking a single blockstore, it would take a blockstore for the "real" blocks and an `Option<Blockstore>` for everything else (although we'd likely want to erase the types to avoid adding a new generic?).
2. Ditto, but just make it a flag? I.e., if some "write intermediates" flag is set, `flush` writes all blocks?
3. Maybe specify it as an option in the machine context? I.e., a flag next to the `tracing` flag?

This will also impact how it's integrated is writing these intermediate blocks to the real blockstore going to be an issue? IMO, probably not for debugging but definitely if we want to use it when tracing.

rvagg:
I'd really like to specify a separate blockstore for this, there's no good reason to pollute the main blockstore with a ton of garbage.

I thought maybe I could put this on the `MachineContext`: `pub flush_intermediates_to: Option<Arc<dyn Blockstore>>,` but that's pretty janky and other stuff has to break just to make the compiler happy with it being there. I could make another argument and then if it's `Some` take that to mean that they should always be flushed. But then I run into problems with the FFI interface:

https://github.com/filecoin-project/filecoin-ffi/blob/037aa5ce11958eb2b47799b4cf16922874f6e2e1/cgo/interface.go#L30-L38

Currently the blockstore is the whole `Externs` object as well, so we `exHandle := cgo.Register(context.TODO(), opts.Externs)` and then use `exHandle` for both the `Externs` and the `Blockstore`. So another blockstore needs another handle, and then I have to figure out how `nil`->`None` works for cgo. Maybe I can hang it off the existing `Externs`, do a `GetIntermediatBlockstore`?

rvagg:
i.e. it was significantly easier to have an entirely separate function and code path that doesn't touch FVM initialisation thanks to the complexity of the FFI layer

rvagg:
Changed to just be a `flush_all_blocks` option on the context. Still depends on https://github.com/filecoin-project/ref-fvm/pull/2109

fvm/src/kernel/default.rs
```diff
@@ -485,6 +485,9 @@ where
             .or_fatal()?;
         self.blocks.mark_reachable(&k);
 
+        // TODO: remove this, it's just for debugging.
+        self.call_manager.log(format!("block_link({})", k));
```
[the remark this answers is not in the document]
Stebalien: I'm happy to add a new event type to the execution log. But that'll be a bit of a larger change (we'll probably need to generalize `CallManager::log`.

I've also considered changing how our gas events work a bit: ideally we'd log syscalls/actor calls then associate gas charges with those syscalls. I'd leave this out of this change and chat about it sync next week.

fvm/src/blockstore/buffered.rs
```diff
@@ -34,6 +34,19 @@ where
...
+    /// regardless of whether they're reachable from any state root.
+    /// Unlike the standard flush() operation which only writes blocks connected
+    /// to the final state tree, this writes every block created during execution.
+    pub fn flush_all(&self) -> Result<()> {
+        log::info!(
+            "Flushing all ({}) cache blocks to blockstore",
+            self.write.borrow().len()
+        );
+        self.base
+            .put_many_keyed(self.write.borrow().iter().map(|(k, v)| (*k, v.as_slice())))
```
Stebalien, line 48: nit: `flush` actually _removes_ them from this blockstore to avoid writing twice (just in case). I'd do that here as well. I.e., `put_many_keyed(self.write.borrow_mut().drain()...`.

fvm/src/blockstore/buffered.rs
```diff
@@ -34,6 +34,19 @@ where
     pub fn into_inner(self) -> BS {
         self.base
     }
+
+    /// Flushes all blocks from the write cache to the provided blockstore,
+    /// regardless of whether they're reachable from any state root.
+    /// Unlike the standard flush() operation which only writes blocks connected
+    /// to the final state tree, this writes every block created during execution.
+    pub fn flush_all(&self) -> Result<()> {
+        log::info!(
```
Stebalien, line 43: IMO, this should be debug at most.

Stebalien (approved):
This seems like a reasonable approach.

fvm/src/blockstore/buffered.rs
rvagg: ahh good point, I've made it do this and added some tests to assert the behaviour of this vs flush()

rvagg:
Converting to draft because it's stacked on #2109 plus a master rebase so this branch is a bit messy and shouldn't be merged as is

rvagg:
So actually, I've just removed the dependency on #2109, that's not really necessary. This is now updated and minimal against master. Would you mind taking another quick look @Stebalien ?

Stebalien (approved):

PR #2103, opened 2025-02-12
Teep pricing changes

Stebalien:


rvagg (approved):
we should probably hold off on merging this for now eh?

Stebalien:
> we should probably hold off on merging this for now eh?

Yeah, probably.

Stebalien:
Ok, I made two changes there:

1. I charged less for batches, assuming some amount of parallelism.
2. I changed some functions around, but that had no real impact on functionality.

BigLep:
> > we should probably hold off on merging this for now eh?
> 
> Yeah, probably.

When do we plan to merge this?

rvagg:
If FIP-0100 gets accepted this gets merged. If it doesn't get accepted we're going to need to have another talk about this.

BigLep:
@rvagg : I assume we're good to merge this now that FIP-0100 is [effectively accepted](https://github.com/filecoin-project/FIPs/discussions/1105#discussioncomment-12491243) (and will be accepted by EI=OD 2025-03-14).

I assume there is communication we need to do with consumers like Venus and Forest to make sure they update?

rvagg:
We have #2107 too, but when we cut a release they'll need to update.

PR #2104, opened 2025-02-13
chore: update ipld-core and serde_ipld_dagcbor

rvagg:
Adds support for default fields by being less strict up-front when decoding maps and lists and allowing the visitors to determine what/how to consume the tokens.

I had a bit of trouble using `cargo` commands to make this happen, particularly to get ipld/encoding/Cargo.toml updated, so I ended up just bumping the numbers manually and doing `cargo build` to get Cargo.lock to update. So I'd like to know what the proper way would be to do this.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2718,7 +2718,7 @@ dependencies = [
  "fvm_ipld_blockstore 0.3.1",
  "multihash-codetable",
  "serde",
- "serde_ipld_dagcbor 0.6.1",
+ "serde_ipld_dagcbor 0.6.2",
  "serde_json",
  "serde_repr",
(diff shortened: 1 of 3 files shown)


Stebalien (approved):

Stebalien:
I usually edit the files manually.

`cargo update ipld-core` won't work because that _just_ updates the lock file, not the requirement specified in the cargo.toml file.

There's a `cargo upgrade` command in the `cargo-edit` package for managing dependencies in `Cargo.toml` files but... that's tricky to use (it tends to upgrade everything) and I honestly don't like it unless I need to batch-upgrade a bunch of packages.

PR #2109, opened 2025-03-14
chore: update to rust 1.81.0

Stebalien:
Matches builtin actors and fixes cargo-deny.


BigLep (approved):
Approving per 2025-03-14 verbal with @rjan90 , @ZenGround0 , and @Stebalien

rjan90 (approved):

Stebalien:
Ok, this is going to take some more work.

Stebalien:
I'm going to need to make a few fixes here. So I'll do this in the next release and we'll just have to live with cargo-deny failing (that's fine, tbh).

rvagg:
@Stebalien I hope you don't mind but I pushed a commit here to make it happier; I'd like to get back to tinkering with https://github.com/filecoin-project/ref-fvm/pull/2101 but I get errors all over my IDE due to ABI mismatches because of the old Rust version and would like to not have to deal with any of that so I'll work on top of this branch instead.

rvagg:
ok, clippy is really annoying to run in this environment .. 🤞

Stebalien:
I've filed https://github.com/bytecodealliance/wasmtime/issues/10454. Our use of features here wasn't effective anyways.

rvagg:
@Stebalien should we just merge this as is now? I think we agreed last week we should get this one done and follow-up later if we get the wasmtime change figured out

PR #2111, opened 2025-03-14
chore: switch to cargo-deny and fix some lints

Stebalien:
- Update url dependency. We don't use it directly, but wasmtime does.
- Update the builtin actors (dev) dependency to remove some crufty dependencies.
- Add licenses everywhere.


deny.toml
```diff
@@ -0,0 +1,25 @@
+[advisories]
+ignore = [
+  "RUSTSEC-2024-0436", # Paste is unmaintained, whatever.
+]
+
+[bans]
+multiple-versions = "allow"
```
Stebalien, line 7: https://github.com/filecoin-project/ref-fvm/issues/2112

deny.toml
```diff
@@ -0,0 +1,25 @@
...
+  "Apache-2.0 WITH LLVM-exception",
+  "BSD-2-Clause",
+  "BSD-3-Clause",
+  "CC0-1.0",
+  "MIT",
+  "NCSA",
+  "Unlicense",
+  "MPL-2.0",
+  "Unicode-3.0",
+  "Zlib",
```
Stebalien, line 24: These are all "permissive" licenses. But we should try to trim down our dependency tree a bit.

rjan90 (approved):

Issue #1724, opened 2023-03-21
FVM versioning and technical debt


lemmih:
The `fvm` crate can be difficult to use natively from Rust for a few reasons, including:

- The incompatibility of `fvm-2` and `fvm-3`,
- the lack of semantic versioning,
- primitive types are dated and that apparently shouldn't be used[1],
- many dependencies (like `cid`) are frozen and cannot easily be upgraded (this is both an annoyance and a security issue).

I'd like to work towards a future where we have a single crate that captures all of the complexity of the Filecoin network, offers a stable interface that other players in the ecosystem can depend on, and actively try to reduce the amount of code that has to be maintained (by de-duplicating code that is shared by `fvm-2` and `fvm-3`).

The Forest team has a lot of Rust experience and we can devote several engineers to work on this full-time. What are you thoughts on this and what are your plans for future versions of the FVM?

[1]: For example, the `fvm_shared::crypto::signature::Signature` type doesn't support delegated signatures which are required to connect to a filecoin network. The Forest team was told that clients aren't supposed to use the types exported by the FVM.

@jennijuju @arajasek @maciejwitowski

Stebalien:
I'd love to see work towards this now that the launch chaos has died down. The current situation is... nasty.

I'd be very interested in separate `fvm1`, `fvm2`, etc. crates instead of using major versions (#358 is in scope right now). I'll just call these FVM epochs for clarity.

We also definitely need to split up and reduce the scope of `fvm_shared`:

1. Crate 1: Basic types that cross between the FVM and actors via syscalls. These types need to be versioned along with FVM epochs (i.e., `fvm_shared1`, `fvm_shared2`). Types that don't change between epochs shouldn't be included.
    - Until we start allowing wasm actor deployment. At that point, we'll have to freeze the these types. But we're not ready for that yet.
2. Crate 2: Types that cross the boundary between the FVM and the client, like `Address`. These types need to be shared by all FVM epochs. However:
    - We need to be _very_ careful with these types to avoid consensus breaking changes. See https://github.com/filecoin-project/go-state-types/blob/393e0b6ffa8839295e6c909b23e57b4a2a099ca5/abi/address.go#L13-L28.
    - This should not include types like `Signature` (that's really a client-only type).

I'd love to see some prototypes and/or design proposals here.

lemmih:
[...]

PR #2113, opened 2025-03-31
feat: log(LogEntry) to handle different log types

rvagg:
Just toying around here to see if I can come up with an answer to https://github.com/filecoin-project/ref-fvm/pull/2101#discussion_r2018903823; @Stebalien I'm not sure what you mean by generalising the execution log but this is my take on it.

### DIFF
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -24,7 +24,7 @@ use crate::eam_actor::EAM_ACTOR_ID;
 use crate::engine::Engine;
 use crate::gas::{Gas, GasTracker};
 use crate::kernel::{
-    Block, BlockRegistry, ClassifyResult, ExecutionError, Kernel, LogEntry, Result, SyscallError,
+    Block, BlockRegistry, ClassifyResult, ExecutionError, Kernel, Result, SyscallError,
 };
 use crate::machine::limiter::MemoryLimiter;
 use crate::machine::Machine;
@@ -514,11 +514,12 @@ where
         Ok(())
     }
 
-    fn log(&mut self, entry: LogEntry) {
-        self.trace(ExecutionEvent::Log(match entry {
-            LogEntry::Syscall(msg) => format!("SyscallLog({})", msg),
-            LogEntry::BlockLink(cid) => format!("BlockLink({})", cid),
-        }));
+    fn log(&mut self, msg: String) {
+        self.trace(ExecutionEvent::Log(msg))
+    }
+
+    fn trace_ipld(&mut self, op: crate::trace::IpldOperation, cid: Cid, size: usize) {
+        self.trace(ExecutionEvent::Ipld { op, cid, size })
     }
 }
 
--- a/fvm/src/call_manager/default.rs
+++ b/fvm/src/call_manager/default.rs
@@ -24,7 +24,7 @@ use crate::eam_actor::EAM_ACTOR_ID;
 use crate::engine::Engine;
 use crate::gas::{Gas, GasTracker};
 use crate::kernel::{
-    Block, BlockRegistry, ClassifyResult, ExecutionError, Kernel, Result, SyscallError,
+    Block, BlockRegistry, ClassifyResult, ExecutionError, Kernel, LogEntry, Result, SyscallError,
 };
 use crate::machine::limiter::MemoryLimiter;
 use crate::machine::Machine;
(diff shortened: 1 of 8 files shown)


fvm/src/call_manager/default.rs
```diff
@@ -514,8 +514,11 @@ where
         Ok(())
     }
 
-    fn log(&mut self, msg: String) {
-        self.trace(ExecutionEvent::Log(msg))
+    fn log(&mut self, entry: LogEntry) {
+        self.trace(ExecutionEvent::Log(match entry {
```
Stebalien, line 518: I was thinking we could add an entirely new event type (not just string logs). New event types aren't breaking (personally, I'd add "put" and "get" events).

The more _breaking_ version would be to change the structure of the event log. Right now the log preserves the tree structure of the calls but not the tree structure of the syscalls, but we could _also_ preserve the structure of the syscalls. But I don't think that's critical right now.

Stebalien: note: I'd also just add these as separate methods? Maybe?

Or maybe we can avoid adding more methods to the kernel by recording put events at the blockstore level? But then we'd need to wrap the blockstore somehow which is... probably even more annoying.

rvagg: > I was thinking we could add an entirely new event type (not just string logs)

OK, so this would involve changing the serialisation as it travels through FFI and would break existing clients. What's the general rule we've followed here for versioning? If we did that we would have to bump some number, is a semver-minor appropriate for that? What justifies a semver-major bump in here?

Stebalien: Well:

1. Adding new event types isn't a breaking change, users will just ignore the ones they don't understand.
2. All FFI serialization is internal to the FFI anyways.
3. We allow API breaking changes in the FVM in _minor_ releases, patch releases are for releases with no breaking changes. This violates semver, but we use major versions for network upgrades. See https://github.com/filecoin-project/ref-fvm/issues/1724 and https://github.com/filecoin-project/ref-fvm/issues/358.

rvagg: OK, so maybe just an internally tagged enum serialization, `["syscall","some debug logging from sys::debug::log"]` and `["block_write",{"/":"bafy..."}]` for what I want, but extended as far as we want with additional stuff.

Thoughts on what other log items might be useful in execution traces?

Stebalien: I wouldn't think hard about the serialization format. Well, that'll be an issue in the _Lotus_ API but the execution trace has a completely different structure there.

I was thinking:

```patch
diff --git a/fvm/src/trace/mod.rs b/fvm/src/trace/mod.rs
index 370828b1..32fbd045 100644
--- a/fvm/src/trace/mod.rs
+++ b/fvm/src/trace/mod.rs
@@ -1,3 +1,4 @@
+use cid::Cid;
 // Copyright 2021-2023 Protocol Labs
 use fvm_ipld_encoding::ipld_block::IpldBlock;
@@ -39,4 +40,15 @@ pub enum ExecutionEvent {
         state: ActorState,
     },
     Log(String),
+    Ipld {
+        op: IpldOperation,
+        cid: Cid,
+        size: u64,
+    },
+}
+
+#[derive(Clone, Debug)]
+pub enum IpldOperation {
+    Get, // open
+    Put, // link
 }
```

rvagg:
OK, introduced a `trace_ipld()` on the `CallManager` that adds a new `ExecutionEvent::Ipld`, for both reads and writes. I wrestled for far too long to try and do it at the `Blockstore` level by introducing a wrapper that could record on reads and writes but I just couldn't reconcile the lifetimes and borrowing and all that guff so I gave up. There's a `preload()` call in here that I also track and it uses a `Blockstore` that gets passed in, but I'm going with the fairly-safe assumption that we're only going to try and load a single block, if any. We don't know whether we had a cache hit or not but I record it anyway. I'm not sure whether this block load will end up being useful or not but they're raw blocks anyway so shouldn't be too hard to filter out if we cared.

Stebalien (approved):

Stebalien:
> OK, introduced a trace_ipld() on the CallManager that adds a new ExecutionEvent::Ipld, for both reads and writes. I wrestled for far too long to try and do it at the Blockstore level by introducing a wrapper that could record on reads and writes but I just couldn't reconcile the lifetimes and borrowing and all that guff so I gave up.

Yeah, rust really expects exactly one owner. I'm not surprised.

> There's a preload() call in here that I also track and it uses a Blockstore that gets passed in, but I'm going with the fairly-safe assumption that we're only going to try and load a single block, if any.

We don't even use that right now. We added that for user-deployable actors but it isn't hooked up anywhere.

Stebalien:
In terms of the tests, it's dying with exit code 143 which apparently means "SIGTERM".

Stebalien:
The hidden error only visible in the raw logs is:

> 2025-04-02T05:27:02.6034191Z ##[error]The runner has received a shutdown signal. This can happen when the runner service is stopped, or a manually started runner is canceled.

Of course, this is also visible to copilot so it was able to figure out the issue immediately....

PR #2115, opened 2025-04-01
chore: remove minicov

rvagg:
Follow-up from https://github.com/filecoin-project/ref-fvm/pull/2114, I don't think any of this is used anywhere so we may as well just remove it?


Stebalien (approved):
The goal was to get code coverage of the SDK, but it would probably require some work to actually wire everything up.

PR #2119, opened 2025-04-04
fix(gas): properly charge NI-PoRep gas according to FIP-0092

rvagg:
PR against the new `release/4.6` branch which is @ tag `fvm@v4.6.0` / 11c4e06f2fbdb4842f53fe2aabe1a1e9ec50a99e

### DIFF
--- a/fvm/src/gas/price_list.rs
+++ b/fvm/src/gas/price_list.rs
@@ -191,9 +191,7 @@ lazy_static! {
                     ]
                 )
             )
-        ].iter()
-        .cloned()
-        .collect(),
+        ].iter().cloned().collect(),
 
         verify_consensus_fault: Gas::new(516422),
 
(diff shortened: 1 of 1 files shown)


Kubuxu (approved):
the fix seems good, not sure about code sequencing/branches/releases

fvm/src/gas/price_list.rs
```diff
@@ -316,6 +314,84 @@ lazy_static! {
...
+                        Step{start: 205, cost: Gas::new(304253590)},
+                        Step{start: 410, cost: Gas::new(509880640)},
+                    ]
+                )
+            ),
+            (
+                RegisteredSealProof::StackedDRG32GiBV1P2_Feat_NiPoRep,
+                StepCost (
+                    vec![
+                        Step{start: 1, cost: Gas::new(112356810)}, // 1
```
irenegia, line 370: for 1 sector, the value should be 118025550 

Indeed, the formula from the FIP test says
44990*126 + 112356810 = 118025550

Stebalien (approved):
Release branch looks correct.

BigLep:
@rvagg @Stebalien : should we also merge this change into master so it goes out with the next batch of ref-fvm changes as well?

PR #2121, opened 2025-04-04
Update to Rust 1.86.0

Stebalien:
This fixes a crash we were seeing in CI in v1.81.0.


Stebalien:
We're going to need to be careful about updating rust in the builtin actors, but we'll need to do it eventually so we might as well do this now.

rvagg:
> need to be careful about updating rust in the builtin actors

Can you describe how you would go about being careful other than pushing through and using existing integration tests to see what breaks? I guess digging through changelogs would be a start, but what else would you look for/at?

Stebalien:
https://github.com/filecoin-project/builtin-actors/issues/1650#issuecomment-2727855449

rvagg (approved):

PR #2130, opened 2025-04-08
chore: update wasmtime to v31.0.0

Stebalien:


rvagg (approved):

Stebalien:
Welp, there's a reason we have the `test_gasfuzz` test. We'll have to figure out why it hung tomorrow.

Stebalien:
Unless you want to try some debugging yourself today.

rvagg:
It's `memory_reservation` somehow. On this branch it spins and spins, not even giving output beyond `test test_gasfuzz has been running for over 60 seconds`. Comment out `c.memory_reservation(instance_memory_maximum_size);` and it runs great. Slightly faster than master too, according to my unscientific testing.

So merge this in and it'll pass: https://github.com/filecoin-project/ref-fvm/pull/2131

fwiw `instance_memory_maximum_size` is still 512MiB in the integration test. I have no clue why it might be a problem, maybe we should understand that!

rvagg:
It's actually `memory_may_move`, this PR changes it from `true` (`static_memory_forced`) to `false, and that's why it's failing. I suspect it would fail also if we were to just change that boolean and not upgrade wasmtime.

I can't figure out why though. Enabling debug logging it seems to do just as much cranelift work before it gets to execution and then there's no output, it freezes up.

There is a test upstream that sort of replicates the situation, but only looking at a basic `memory.grow` and asserting an error from it with our setup, but we don't see an error: https://github.com/bytecodealliance/wasmtime/blob/94ec88eacfc93203f168fc6fd688445e1d49188f/tests/all/memory.rs#L429-L440

Even making `memory_reservation_for_growth` bigger to take up the slack doesn't fix it. Smells like a wasmtime bug to me :shrug:.

I think we should just make it `true` and rely on `max_memory_size` to handle what we care about.

Stebalien:
Yeah, we set it that way because performance. But yeah, we'll just change it and file an issue.

Stebalien:
Any size less than 4GiB fails with `memory_may_move` is false. I looked at the debug logs and everything just stops, so something is clearly getting stuck in wasm a few recursive calls in (we're creating an EVM smart contract and having it call itself recursively).

There seems to be some kind of lowering issue where lowering code with bounds checking spins forever (or is just much, much slower?) and lowering code without bounds checking doesn't.

Stebalien:
In the working case, we recurse 6 times before running out of gas. In the broken case, we recurse 4 times and hang. I tried with an optimized build (`cargo test --release`) in case that was it and the test was much faster, but it hung at the same place.

Stebalien:
Nevermind, I wasn't logging correctly. It literally just hits the EVM and stops. The 4 calls I saw were related to constructing the EVM contract. On failure, we see:

```
calling 100 -> 10 (account calling the evm)
calling 10 -> 100 (eam checking what the caller is)
calling 10 -> 1 (eam calling the init actor)
calling 1 -> 101 (constructor)
```

Disabling the test.... basically everything hangs, so there's definitely a bug here.

Stebalien:
I'm going to merge this because the issue appears to be fixed and getting to the bottom of it isn't going to be quick.

Issue #2128, opened 2025-04-08
Consider reverting `wasmtime::Config::memory_reservation` to the default 4GiB


Stebalien:
Reverting this to 4GiB can unlock performance as wasmtime can elide many bounds checks. We still limit _actual_ instance memory to 512MiB, this 4GiB is just the amount of _virtual_ memory allocated to the instance.

The main concern will be RPC providers and anyone who sets a high FVM max parallelism. Doing so will increase the number of wasmtime instances in the pool which will use more virtual memory (address-space). We should consider dividing the max instances by 8 to match the current max virtual memory used.

Stebalien:
Actually, before we were:

1. Using the default guard-page size of 2GiB either side.
2. 512MiB reservation for the instance itself.

Now, the default guard-page size is 32MiB, so we can increase the instance reservation size to 4GiB while using less virtual address-space than we were before.

PR #2131, opened 2025-04-08
chore: use default wasmtime memory_reservation

rvagg:
Closes: https://github.com/filecoin-project/ref-fvm/issues/2128

Default is 4GiB, the initial size of the virtual address space. The full 32-bit address space means runtime memory bounds checking should be elided.

I think this is all that's needed, but it might be interesting to try and do some basic before & after benchmarking. I can't find an obvious way to do that though, @Stebalien suggestions?

### DIFF
--- a/fvm/src/engine/mod.rs
+++ b/fvm/src/engine/mod.rs
@@ -168,11 +168,7 @@ fn wasmtime_config(ec: &EngineConfig) -> anyhow::Result<wasmtime::Config> {
     // compiled wasm modules.
     c.memory_init_cow(false);
 
-    // wasmtime default: 4GB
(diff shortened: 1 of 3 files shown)


Stebalien (approved):
This passes the test and it ensures that memory won't move, so I'm going with this one. We haven't priced-in the cost of moving allocated wasm memory, so I'd much prefer to leave `may_move_memory` as false.

Stebalien:
I think part of our testing process here will be to ask glif to upgrade their calibnet node to see if anything breaks. We allocate all instances up-front (virtual-memory wise) so it should be pretty obvious.

PR #2133, opened 2025-04-08
test(fvm): improve oom tests

Stebalien:
This adds a memory test for just less than the max and changes one of the oom tests to allocate just a bit more than that. The "just a bit" is less than a wasm page (64KiB) so we should be pretty accurately testing the available memory. Unfortunately, it is a it fuzzy because I don't feel like writing wasm directly (even though we could...).

### DIFF
--- a/testing/integration/tests/main.rs
+++ b/testing/integration/tests/main.rs
@@ -810,7 +810,58 @@ fn backtraces() {
 }
 
 #[test]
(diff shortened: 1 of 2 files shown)


Stebalien:
Merged with https://github.com/filecoin-project/ref-fvm/pull/2131.

PR #2142, opened 2025-04-09
chore: update wasmtime to v31.0.0

Stebalien:


rvagg (approved):

Stebalien:
We're stalling while compiling the wasm modules. Reducing the optimization level or changing the register allocation algorithm to single-pass fixes it. My guess is that pre-running wasm-opt on the wasm modules would also fix it.

More investigation required as to why this happens _here_ (it specifically seems to affect the evm and datacap actors).

rvagg:
I debugged my way to the same assumption. It starts failing with wasmtime 28 and only using the _new_ register allocation algorithm works. The patch that introduced the new allocation algorithm didn't touch the old one, but there was a large refactor that did do some heavy touching, I wonder if they introduced a regression there: https://github.com/bytecodealliance/regalloc2/pull/196

Stebalien:
I've narrowed it down to regalloc v0.11.3 (v0.11.2 works). That's going to make this much easier.

Stebalien:
The regalloc bug appears to also be responsible for the issues we ran into in the gassfuzz test. It only showed up multiple calls deep because we _don't_ pre-compile actors modules when testing (to avoid pre-compiling actors that never get used).

We _do_ pre-compile actors in the conformance tests because we're better able to share those pre-compiled actors between test runs.

rvagg (approved):

Issue #2144, opened 2025-04-09
Remove `async-std` dependency


Stebalien:
It has been deprecated and we only use it in the bundler and the conformance tests, and only there because we inherited it from the car library we pulled from forest.

We could consider switching to tokio, but we should probably just switch to blocking and thread-pools in these cases (we're not doing any networking anyways).

rvagg:
Done in #2180 and #2181; should be gone from Cargo.lock when we update the next builtin-actors bundle (when we have a release to work with, we're still using 16.0.1), https://github.com/filecoin-project/builtin-actors-bundler is the remaining transitive dependency that uses async-std and v8.0.0 is already in builtin-actors without async-std waiting for a release.

PR #2153, opened 2025-04-10
Remove async from `fvm_ipld_car`

Stebalien:
This isn't highly parallel networking code so making it async isn't worth the added complexity. The logic in this library isn't _that_ complex, but the async logic in the conformance test engine definitely is and I want to remove all async code from that library next.

TL;DR: slowly remove an unmaintained dependency (`async-std`) and make everything a little more maintainable at the same time.

Part one of #2144

### DIFF
--- a/ipld/car/src/block.rs
+++ b/ipld/car/src/block.rs
@@ -1,0 +1,50 @@
+// Copyright 2021-2023 Protocol Labs
+// SPDX-License-Identifier: Apache-2.0, MIT
+use cid::Cid;
+use multihash_codetable::{Code, MultihashDigest};
+
+use super::Error;
+
+/// IPLD Block
+#[derive(Clone, Debug)]
+pub struct Block {
+    pub cid: Cid,
+    pub data: Vec<u8>,
+}
+
+impl From<Block> for (Cid, Vec<u8>) {
+    fn from(block: Block) -> Self {
+        (block.cid, block.data)
+    }
+}
+
+impl From<(Cid, Vec<u8>)> for Block {
+    fn from((cid, data): (Cid, Vec<u8>)) -> Self {
+        Block { cid, data }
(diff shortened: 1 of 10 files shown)


ipld/car/src/util.rs
```diff
@@ -2,18 +2,17 @@
 // Copyright 2019-2022 ChainSafe Systems
 // SPDX-License-Identifier: Apache-2.0, MIT
 
+use std::io::{self, Read};
```
rvagg, line 5:
```suggestion
use std::io::{self, Cursor, Read};
```
?

rvagg: or down in the `mod tests` block?

rvagg: nah, actually, nevermine

ipld/car/src/block.rs
```diff
@@ -0,0 +1,50 @@
...
+                if self.cid.hash().digest() != self.data {
+                    return Err(Error::InvalidFile(
+                        "CAR has an identity CID that doesn't match the corresponding data".into(),
+                    ));
+                }
+            }
+            code => {
+                let code = Code::try_from(code)?;
+                let actual = Cid::new_v1(self.cid.codec(), code.digest(&self.data));
+                if actual != self.cid {
```
rvagg, line 40: does this mean we don't support CARs with v0 CIDs?

rvagg: or is there an equality overload somewhere in `Cid` that's able to do a proper comparison?

rvagg (approved):

ipld/car/src/block.rs
Stebalien: We don't really care about v0, but that's still a bug. I'll fix it in a followup PR.

PR #2156, opened 2025-04-15
fix(fvm): accept malleable secp256k1 signatures (per EVM, etc.)

Stebalien:
And add a test case.


rvagg (approved):
yeah, ok, worth recording here that the context is mostly in https://github.com/RustCrypto/elliptic-curves/issues/988, foundry-rs had a similar issue and it was resolved in revm which looks a lot like what we're implementing here: https://github.com/bluealloy/revm/pull/870

PR #2162, opened 2025-04-15
Switch to rust 2024 edition

Stebalien:
The first commit makes the switch, the second commit is just formatting.


ipld/bitfield/tests/bitfield_tests.rs
```diff
@@ -10,7 +10,7 @@ use rand_xorshift::XorShiftRng;
 
 fn random_indices(range: u64, seed: u64) -> Vec<u64> {
     let mut rng = XorShiftRng::seed_from_u64(seed);
-    (0..range).filter(|_| rng.gen::<bool>()).collect()
+    (0..range).filter(|_| rng.r#gen::<bool>()).collect()
```
Stebalien, line 13: This is why there's a new rand release. `gen` is now a reserved keyword so it can only be used by treating it as a "raw" identifier hence the `r#` prefix.

Stebalien:
Note: we don't need releases here. The main issue is that this change is _noisy_ so I want to get it over with while things are mostly quiet.

fvm/src/engine/mod.rs
```diff
@@ -491,7 +491,7 @@ impl Engine {
         let module = match cache.get(k) {
             Some(m) => m.module.clone(),
             None => {
-                let module = Module::deserialize(&self.inner.engine, compiled)?;
+                let module = unsafe { Module::deserialize(&self.inner.engine, compiled)? };
```
rvagg, line 494: what's going on here? Did wasmtime mark this as unsafe internally and now 2024 is forcing us to acknowledge it?

fvm/src/kernel/mod.rs
```diff
@@ -278,7 +278,7 @@ pub trait RandomnessOps {
     /// beacon from a given epoch.
     /// This randomness is not tied to any fork of the chain, and is unbiasable.
     fn get_randomness_from_beacon(&self, rand_epoch: ChainEpoch)
-        -> Result<[u8; RANDOMNESS_LENGTH]>;
+    -> Result<[u8; RANDOMNESS_LENGTH]>;
```
rvagg, line 281: yuk, this seems like the opposite of the introduction of indenting rules in other places!

rvagg (approved):
nice improvements on the formatting rules, except for that one case where we lose indenting!

fvm/src/engine/mod.rs
Stebalien: The method we're calling this from is marked as `unsafe`. Previously, functions marked as unsafe could call unsafe functions without declaring an unsafe block. Now, marking the function as unsafe doesn't let you call unsafe functions without declaring a separate unsafe block.

fvm/src/kernel/mod.rs
Stebalien: Yeah, I agree...

PR #2168, opened 2025-04-16
fix: correctly deserialize test vectors

Stebalien:
I renamed the field when switching to rust 2024 but this broke deserializing test vectors (which we only use in v3/v2).


Stebalien:
I've created a task for extracting test vectors with fvm v4: https://github.com/filecoin-project/ref-fvm/issues/2169.

PR #2175, opened 2025-04-28
feat: add nv27-skeleton

TippyFlitsUK:
Ref: #2174

### DIFF
--- a/fvm/Cargo.toml
+++ b/fvm/Cargo.toml
@@ -60,6 +60,7 @@ gas_calibration = []
 # The current implementation keeps it by default for backward compatibility reason.
 # See <https://github.com/filecoin-project/ref-fvm/issues/2001>
 verify-signature = []
+nv27-dev = []
 
 # Allow coverage attribute.
 [lints.rust]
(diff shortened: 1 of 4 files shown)


rjan90:
Thanks! Looks relatively good to me. One nit I think we can fix in the PR as well, is adding the correct network version name for V25 (Teep) here: https://github.com/filecoin-project/ref-fvm/blob/a3f209bec55f19122a1f953f89ff01cdf66e7ad5/shared/src/version/mod.rs#L63-L67

And additionally add V26 (Tock) and V27 (TBD)

TippyFlitsUK:
The mod.rs changes had already been made, so there was nothing to commit for this. I will tweak the comments slightly, though, as you suggested. New commit incoming. 🚀

rjan90:
> The mod.rs changes had already been made, so there was nothing to commit for this. 

Yeah, looks like the additional changes was merged here: https://github.com/filecoin-project/ref-fvm/pull/2107/files

rjan90 (approved):

Issue #1996, opened 2024-02-20
Replace `for_each_mut` (internal iteration) with `iter_mut` (external iteration) in the AMT


Stebalien:
The iterator will have to yield a smart pointer that:

1. Can be dereferenced immutably with no affect.
2. Can be dereferenced mutably, marking the value as changed.
3. Can be deleted (e.g., by calling some form of `ptr.delete()` method). This method should _consume_ the smart pointer (possibly returning the inner value).

This will get rid of the last "internal" iteration feature.

PhantomOz:
can i work on this?

Stebalien:
Absolutely! Take a look at how we currently use it [here](https://github.com/filecoin-project/builtin-actors/blob/a0e34d22665ac8c84f02fea8a099216f29ffaeeb/actors/miner/src/bitfield_queue.rs#L63-L85).

rjan90:
Hey @PhantomOz! Are you still interested in working on this, or should I unassign you?

rjan90:
Unassigning since I have not hear back yet.

PhantomOz:
I apologize didn't notice that i was assigned till now, Please reassign so that I can work on it @rjan90

rjan90:
No worries! Reassigned to you now 👍

rjan90:
Hey @PhantomOz! Do you need some additional guidance on this issue ticket? How is the progress on going on this?

PhantomOz:
@rjan90 Yes I do

Sahilgill24:
Hey **@rjan90** , If the issue is still open to be worked upon , Can i give it a try ?

rjan90:
Hey @Sahilgill24! Yes, it is still open to be worked on. Go ahead and give it a try, and feel free to ask any questions you may have

Sahilgill24:
Hey @rjan90 
My understanding of the issue is that currently  `for_each_mut()` is based on internal iteration and works like this  ```amt.for_each_mut(f)```  where the function `for_each_mut` calls the `for_each_while_mut()` inside of it  , If The function `iter_mut()` is based on  external iteration so it may look like this  
```
for i in amt.iter_mut() { 
       *(i.derefmut()).for_each_while_mut(f)
 }
``` 
 where `i` is the custom smart pointer and `f` is the closure. 
Is this the correct approach or am i doing something wrong ?

Stebalien:
[...]

PR #2178, opened 2025-05-22
Refactor AMT to use iter_mut Instead of for_each_mut 

Sahilgill24:
This PR refactors the AMT mplementation by replacing internal iteration via `for_each_mut` with external iteration using `iter_mut` and adds a test case showing the use case of the `iter_mut` function . 

**Motivation:**

Solves #1996 

**Progress:**

- [x] The iter_mut function yields a Smart pointer `Amtptr`.
- [x] `Amtptr` can be dereferenced immutably.
- [x] Can be dereferenced mutably.
- [ ] Can be deleted to yield the inner value. 

> [!NOTE]
>  Instead of using the `deref_mut` ,  the `get_mut` function can be used instead as written in the test case and as shown 
> below 

```rust

    for ptr in new_amt.iter_mut() {
        let current_idx = ptr.0;
        let mut val = ptr.1;
        let mut val_ref = val.get_mut();
        f(current_idx, &mut ValueMut::new(&mut *val_ref)).unwrap();
    }
```

### DIFF
--- a/ipld/amt/src/amt.rs
+++ b/ipld/amt/src/amt.rs
@@ -111,7 +111,7 @@ use fvm_ipld_encoding::ser::Serialize;
 use fvm_ipld_encoding::serde::Deserialize;
 use itertools::sorted;
 use multihash_codetable::Code;
-use std::cell::{Ref, RefCell, RefMut};
+use std::cell::{RefMut, RefCell};
 use std::ops::{Deref, DerefMut};
 use std::rc::Rc;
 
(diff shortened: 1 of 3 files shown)


Sahilgill24:
hey @rvagg , so while trying out the issue , I found that there was no direct way to make a iterator from the `for_each_while_mut` function , so my thought was to make a cloned copy of the AMT where all the changes would happen then going to each node to do the changes in the original AMT , but I feel we could also add a function to the node.rs that gives a mutable referance to the val and then make an external iterator directly rather than cloning the tree. Could you help me out with this ?

PR #2181, opened 2025-07-07
chore: migrate conformance test runner to tokio

rvagg:
Forward-port of #2173


testing/conformance/tests/runner.rs
```diff
@@ -74,76 +75,64 @@ async fn conformance_test_runner() -> anyhow::Result<()> {
...
-                let res = run_vector(path.clone(), stats, tracer)
-                    .await
-                    .with_context(|| format!("failed to run vector: {}", path.display()))?;
-                anyhow::Ok((path, res))
-            })
-            .map(futures::future::Either::Left),
-        )
+    // Collect test vector files
+    let vector_paths = if path.is_file() {
+        vec![path]
```
LesnyRumcajs, line 80: Double checking - this will not go into some weird behaviour if `path` is a symlink, correct? I mean, `path.is_file` should follow the link correctly, but I'm not sure about the further usage.

testing/conformance/tests/runner.rs
```diff
@@ -74,76 +75,64 @@ async fn conformance_test_runner() -> anyhow::Result<()> {
...
     let mut failed = 0;
     let mut skipped = 0;
 
-    while let Some((path, res)) = results.next().await.transpose()? {
-        match res {
-            VariantResult::Ok { id } => {
-                report!("OK".on_green(), path.display(), id);
-                succeeded += 1;
+    // Process results
+    for result in results.drain(..) {
```
LesnyRumcajs, line 112: Any reason for a `drain` here and not an `into_iter` or something like this? Do we re-use the `results` anywhere later?

testing/conformance/tests/runner.rs
rvagg: shouldn't do, the usage is fairly straightforward and the symlink should be transparent from this point onward (we use `File::open(&path)` which should also be fine with symlinks)

testing/conformance/tests/runner.rs
```diff
@@ -74,76 +75,64 @@ async fn conformance_test_runner() -> anyhow::Result<()> {
...
+            let stats = stats.clone();
+            let tracer = tracer.clone();
+
+            async move {
+                // Run the vector processing in a blocking task
+                tokio::task::spawn_blocking(move || run_vector(path, stats, tracer)).await?
+            }
+        })
+        .buffer_unordered(*TEST_VECTOR_PARALLELISM)
+        .collect::<Vec<_>>()
```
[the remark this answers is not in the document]
rvagg: will keep for consistency for now but something to consider adopting more generally as a batch

testing/conformance/tests/runner.rs
rvagg: nope, good point! changed

Issue #2186, opened 2025-07-22
update `wasmtime` to at least 33.0.2


LesnyRumcajs:
The current `wasmtime` version (31.0.0) major version doesn't have a patch for https://rustsec.org/advisories/RUSTSEC-2025-0046.html. I'm not sure if it's a big deal in the FVM case (the official severity score is low), but it'd be good to do regardless.

Note that `wasmtime` updates can significantly affect performance, so it'd be great to do some benchmarks with this update. See https://github.com/filecoin-project/ref-fvm/issues/2058 for historical reference.

LesnyRumcajs:
updated via https://github.com/filecoin-project/ref-fvm/pull/2226 to 36

PR #2187, opened 2025-07-22
chore: suppress RUSTSEC-2025-0046

LesnyRumcajs:
This suppresses https://rustsec.org/advisories/RUSTSEC-2025-0046.html. See https://github.com/filecoin-project/ref-fvm/issues/2186 for full context. Without this, unrelated PRs will fail.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -1,7 +1,7 @@
 [advisories]
 ignore = [
   "RUSTSEC-2024-0436", # Paste is unmaintained, whatever.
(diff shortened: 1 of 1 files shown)


BigLep:
@LesnyRumcajs : it looks like followup permissions work was done in https://github.com/filecoin-project/github-mgmt/pull/156#event-18934889200 .  Hopefully this is self-service in the future.  Let me know if it isn't.

PR #2189, opened 2025-07-31
feat: cacheless amt iteration

LesnyRumcajs:
This exposes an alternative iteration method without caching. Without it, the memory usage for iterating over deal proposals is too high (my 64 GiB machine gets knocked out under a minute). For reference, this works fine in Go's AMT implementation.

With `for_each_cacheless`, I can iterate the proposals in ~90s (no-op iteration).

The logic is mostly taken from the `for_each_while_mut`.

### DIFF
--- a/ipld/amt/src/lib.rs
+++ b/ipld/amt/src/lib.rs
@@ -5,7 +5,7 @@
 //! AMT crate for use as rust IPLD data structure
 //!
 //! Data structure reference:
-//! <https://github.com/ipld/specs/blob/51fab05b4fe4930d3d851d50cc1e5f1a02092deb/data-structures/vector.md>
+//! https://github.com/ipld/specs/blob/51fab05b4fe4930d3d851d50cc1e5f1a02092deb/data-structures/vector.md
 
 mod amt;
 mod diff;
--- a/ipld/amt/src/lib.rs
+++ b/ipld/amt/src/lib.rs
@@ -5,7 +5,7 @@
 //! AMT crate for use as rust IPLD data structure
 //!
 //! Data structure reference:
-//! https://github.com/ipld/specs/blob/51fab05b4fe4930d3d851d50cc1e5f1a02092deb/data-structures/vector.md
+//! <https://github.com/ipld/specs/blob/51fab05b4fe4930d3d851d50cc1e5f1a02092deb/data-structures/vector.md>
 
 mod amt;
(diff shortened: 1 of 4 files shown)


LesnyRumcajs:
CI blocked by https://github.com/filecoin-project/ref-fvm/pull/2187

rvagg:
The only concern I have here is that this reportedly reads twice as many blocks as the cached version which seems off to me. I'd like to figure out what's going on there first.

rvagg (commented):
needs docs, but lgtm
can be rebased on master because the other PR is merged now

ipld/amt/src/lib.rs
```diff
@@ -5,7 +5,7 @@
 //! AMT crate for use as rust IPLD data structure
 //!
 //! Data structure reference:
-//! https://github.com/ipld/specs/blob/51fab05b4fe4930d3d851d50cc1e5f1a02092deb/data-structures/vector.md
+//! <https://github.com/ipld/specs/blob/51fab05b4fe4930d3d851d50cc1e5f1a02092deb/data-structures/vector.md>
```
LesnyRumcajs, line 8:
```
warning: this URL is not a hyperlink
 --> ipld/amt/src/lib.rs:8:5
  |
8 | //! https://github.com/ipld/specs/blob/51fab05b4fe4930d3d851d50cc1e5f1a02092deb/data-structures/vector.md
  |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  |
  = note: bare URLs are not automatically turned into clickable links
  = note: `#[warn(rustdoc::bare_urls)]` on by default
help: use an automatic link instead
  |
8 | //! <https://github.com/ipld/specs/blob/51fab05b4fe4930d3d851d50cc1e5f1a02092deb/data-structures/vector.md>
  |     +                                                                                                     +

```

LesnyRumcajs: There are other doc links issues in the repo; I'll have a look later and see if there can be a friendly CI check to ensure this doesn't happen in the future.

rvagg (approved):
nice
I bet the HAMT could get exactlty the same treatment

PR #2204, opened 2025-08-25
chore: release v2.11.2

LesnyRumcajs:
Release with backported `filecoin-proofs-api` version bump, without which there's a dependency issue in Forest (using v19 of the `filecoin-proofs-api` and then providing it to the v2.11.1 causes a compilation fiasco, as expected).


rvagg (approved):
we need v3 too don't we?

PR #2210, opened 2025-08-27
chore: release v3.13.2

LesnyRumcajs:
Release with backported:
- https://github.com/filecoin-project/ref-fvm/pull/2206
- https://github.com/filecoin-project/ref-fvm/pull/2207

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -8,7 +8,7 @@ members = [
 ]
 
 [workspace.package]
-version = "3.13.1"
+version = "3.13.2"
 license = "MIT OR Apache-2.0"
 edition = "2024"
 repository = "https://github.com/filecoin-project/ref-fvm"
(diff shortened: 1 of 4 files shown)


rvagg (approved):

Cargo.toml
```diff
@@ -23,9 +23,9 @@ ipld-core = { version = "0.4.2", features = ["serde"] }
 wasmtime = {version = "31.0.0", default-features = false, features = ["cranelift", "pooling-allocator", "parallel-compilation", "runtime"] }
 wasmtime-environ = "31.0.0"
 
-fvm = { path = "fvm", version = "~3.13.0", default-features = false }
-fvm_shared = { path = "shared", version = "~3.13.0", default-features = false }
-fvm_sdk = { path = "sdk", version = "~3.13.0" }
```
rvagg, line 28: did we get these (and v2) wrong in the .1 releases?

LesnyRumcajs: To my understanding, the minor version is completely ignored with the [tilde requirement](https://doc.rust-lang.org/cargo/reference/specifying-dependencies.html#tilde-requirements). That is, `~3.13.0` is equivalent to `~3.13.42`; perhaps we should just use `~3.13` so that the `cargo set-version` doesn't bother with it. What do you think?

rvagg: yeah, I guess it depends on what ends up in the lock, but it does seem like we should just keep them in step

PR #2266, opened 2026-03-18
chore: revive and improve release automation

galargh:
This PR revives the release automation effort from #2027 with several key improvements:

- **Workspace-Aware Dry-Run**: Updated `release-check.yml` to use `cargo publish --workspace --dry-run`. This leverages Cargo 1.90+'s native multi-package publishing to correctly verify the entire dependency graph, solving the previous failure with interdependent crates.
- **Automated Publishing**: Updated `releaser.yml` to include a `cargo publish --workspace` step. This automates the final push to crates.io upon merge, ensuring crates are published in the correct order.
- **Updated Documentation**: Refreshed `RELEASE.md` to reflect the new automated process and cleaned up `CONTRIBUTING.md` to match the current project structure.
- **Merge Master**: Incorporated the latest changes from `master`, including the transition of release documentation to its own file.

### Test Plan

To verify these changes, we will perform a live test by simulating a multi-crate release:

1. **Create a Test PR**: Open a new PR targeting the `galargh/release` branch.
2. **Version Bumps**: In the test PR, bump the versions of `fvm_shared`, `fvm`, and `fvm_sdk` simultaneously in `Cargo.toml`.
3. **Verify Release Checker**:
   - Ensure the `release-check` workflow triggers correctly.
   - Verify that the `Dry-run publish` step (`cargo publish --workspace --dry-run`) passes.
   - **Crucial Check**: This should succeed even though the new versions of the dependencies are not yet on crates.io, confirming that Cargo's native workspace-aware publishing is correctly resolving internal dependencies.
4. **Validation**: Check that the draft GitHub Releases are correctly created for each crate.

Checklist:

- [x] ✅ https://github.com/filecoin-project/ref-fvm/pull/2267

PR #2284, opened 2026-04-17
Revert "Revert "chore: revive and improve release automation (#2266)""

galargh:
Reverts filecoin-project/ref-fvm#2275

See https://github.com/filecoin-project/ref-fvm/pull/2266#issuecomment-4270415546

### DIFF
--- a/.github/workflows/release-check.yml
+++ b/.github/workflows/release-check.yml
@@ -1,0 +1,58 @@
+name: Release Checker
+
+on:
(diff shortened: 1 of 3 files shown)


rvagg (approved):
ok, I'm playing catch up here; this seems good to me but  I see a note in #2275 from @rjan90 about not proceeding with this for now so I don't know if there remains a valid objection with this figured out?

rjan90:
No objections from my side! Fine to proceed with this now

PR #2285, opened 2026-04-20
fix: backport multihash-codetable bump to `release/v3`

rjan90:
## What changed

This backports the `multihash-codetable` bump from #2281 onto `release/v3`.

It updates the workspace dependency to `multihash-codetable = 0.2`, refreshes the lockfile so the resolved graph moves to `cid 0.11.2`, `multihash 0.19.4`, and `multihash-codetable 0.2.1`, and adds the same `RUSTSEC-2026-0097` ignore in `deny.toml` that `master` needed after the dependency refresh.

## Why

The `release/v3` branch was still resolving older `cid`/`multihash` packages that pulled in `core2`. Backporting the bump here removes `core2` from the branch and unblocks downstream consumers that still depend on the v3 line.

## Impact

This is a dependency-only backport. There are no source changes beyond the dependency and advisory configuration updates.

## Validation

- `cargo check --workspace`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -634,36 +634,36 @@ dependencies = [
 
 [[package]]
 name = "cranelift-assembler-x64"
-version = "0.123.6"
+version = "0.123.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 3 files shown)


hanabi1224 (approved):
LGTM, I'd run `cargo update` to fix the `cargo deny` CI failure

PR #2313, opened 2026-07-29
ci: port release automation to release/v2

galargh:
## Summary
- add the releaser and release-check workflows to release/v2
- limit release automation sources to the v2 publishable crates: fvm, sdk, and shared

## Validation
- Ruby YAML parsed both workflow files
- confirmed listed Cargo.toml paths exist
- cargo metadata --no-deps --format-version 1 passed

Note: actionlint is not installed locally, so it was not run.

### DIFF
--- a/.github/workflows/releaser.yml
+++ b/.github/workflows/releaser.yml
@@ -1,0 +1,50 @@
+name: Releaser
+
+on:
+  push:
+    paths:
+      - "**/Cargo.toml"
+  workflow_dispatch:
+
+permissions:
+  contents: write
+  id-token: write
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.sha }}
+  cancel-in-progress: true
+
+jobs:
+  releaser:
+    uses: ipdxco/unified-github-workflows/.github/workflows/releaser.yml@v1.0
+    with:
+      sources: |
+        [
+          "fvm/Cargo.toml",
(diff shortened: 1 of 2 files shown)


.github/workflows/releaser.yml
```diff
@@ -0,0 +1,51 @@
...
+          max_attempts: 3
+          shell: bash
+          command: |
+            sudo apt-get update
+            sudo apt-get install --no-install-recommends --yes libhwloc-dev ocl-icd-opencl-dev
+      - name: Publish to crates.io
+        run: |
+          # Use cargo publish --workspace to publish all changed crates in the workspace.
+          # This handles interdependent crates correctly and only publishes those with version bumps.
+          cargo publish --workspace
```
rvagg, line 51:
```suggestion
          cargo publish --workspace --exclude fvm_conformance_tests
```

it looks like it's in v2 as well so we should exclude this as well I think?

rvagg (approved):
sans that conformance test comment, pls check before merging

.github/workflows/releaser.yml
galargh: Actually, I think we can safely remove that exclude from v3 after all. I misread the configs. It shouldn't be needed, but let me double-check before merging either of those.

galargh: Double-checked, it's fine to skip

PR #2341, opened 2026-09-14
fix: configure trusted cargo publishing

galargh:
This adds a token exchange step to the releaser workflow, which should fix the trusted publishing setup here. Once it's merged, we should cherry-pick this to the v2 and v3 release branches as well. I'll also do a review of all the other repos where we have trusted publishing set up to see if we're not missing it there either.


rvagg (approved):
looks close enough https://crates.io/docs/trusted-publishing

PR #2342, opened 2026-09-15
fix: configure trusted cargo publishing

galargh:
This PR is cherry-picking https://github.com/filecoin-project/ref-fvm/pull/2341 onto the release branch.


BigLep (approved):
Approving given @rvagg approved https://github.com/filecoin-project/ref-fvm/pull/2341 and no difference here.

PR #2346, opened 2026-09-16
ci: use cargo-release for publishing

galargh:
Context: https://filecoinproject.slack.com/archives/C029MT4PQB1/p1789406428177479

This PR introduces cargo release - we want it here because `cargo release publish` skips publishing already published crates while raw `cargo publish` fails on that scenario. We need this functionality because we might be release a subset of publishable crates from this repo.

I tested this setup in https://github.com/galargh/crates


.github/workflows/release-check.yml
```diff
@@ -51,8 +54,10 @@ jobs:
           command: |
             sudo apt-get update
             sudo apt-get install --no-install-recommends --yes libhwloc-dev ocl-icd-opencl-dev
+      - name: Install cargo-release
+        run: cargo install cargo-release --version 1.1.5 --locked
```
LesnyRumcajs: As you noticed, building this tool is slow. We have decent tooling in the Rust ecosystem at our disposal, though.

https://github.com/taiki-e/install-action

LesnyRumcajs (approved):
Looks correct, we'll see if all works well in practice at the next release I guess. :)

